PR #1, opened 2025-04-03
Introducing Subprocess

iCharlesHu:
Imported from:
https://github.com/iCharlesHu/Subprocess/commit/ec56c66c6c289caee566b9e69d58f1c71681d71e

I will address the following issues separately (before we tag `0.0.1`) once this PR lands since it's big enough as it is:

- [ ] https://github.com/iCharlesHu/Subprocess/issues/46
- [ ] https://github.com/iCharlesHu/Subprocess/issues/45
- [ ] https://github.com/iCharlesHu/Subprocess/issues/44
- [ ] https://github.com/iCharlesHu/Subprocess/issues/55


README.md
```diff
@@ -1 +1,305 @@
-# swift-subprocess
\ No newline at end of file
+<a id="readme-top"></a>
+
+# Subprocess
+
+Subprocess is a cross-platform package for spawning processes in Swift.
+
+It's like [Foundation.Process](https://developer.apple.com/documentation/foundation/process), but written for Swift and build on top of structural concurrency.
```
parkera: Let's omit this part of the description. I think the first sentence is sufficient, no need to reference `Process` here.

README.md
```diff
@@ -1 +1,305 @@
...
+Use it by setting `.bytes` or `.bytes(limit:)` for `input` or `error`.
+
+#### `SequenceOutput`:
+
+This option redirects the child output to the `.standardOutput` or `.standardError` property of `Execution`. It’s only for the `run()` family that takes a custom closure.
+
+
+### Cross-platform support
+
+`Subprocess` works on all major platforms supported by Swift, including macOS, Linux, and Windows, with feature parity on all platforms as well as platform-specific options for each.
```
parkera: "all" contradicts the table below, where it is unsupported on iOS (for example). I think I would just say here something like "the following platforms are currently supported for Subprocess".

README.md
```diff
@@ -1 +1,305 @@
...
+
+## Contributing to Subprocess
+
+Subprocess is part of the Foundation project. We have a dedicated [Foundation Forum][forum] where people can ask and answer questions on how to use or work on this package. It's also a great place to discuss its evolution.
+
+[forum]: https://forums.swift.org/c/related-projects/foundation/
+
+If you find something that looks like a bug, please open a [Bug Report][bugreport]! Fill out as many details as you can.
+
+[bugreport]: https://github.com/iCharlesHu/Subprocess/issues/new?assignees=&labels=bug&template=bug_report.md
```
parkera: This URL is outdated.

README.md
```diff
@@ -1 +1,305 @@
...
+## Code of Conduct
+
+Like all Swift.org projects, we would like the Subprocess project to foster a diverse and friendly community. We expect contributors to adhere to the [Swift.org Code of Conduct](https://swift.org/code-of-conduct/).
+
+
+<p align="right">(<a href="#readme-top">back to top</a>)</p>
+
+## Contact information
+
+The current code owner of this package is Charles Hu ([@iCharlesHu](https://github.com/iCharlesHu)). You can contact him [on the Swift forums](https://forums.swift.org/u/icharleshu/summary).
```
parkera: Instead of you personally, let's list the Foundation Workgroup as the maintainers of the package. I think that more accurately represents the community-driven process we went through to accept the creation of the package into swiftlang.

Sources/Subprocess/API.swift
```diff
@@ -0,0 +1,423 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
```
parkera: Can you do a find/replace for 2024 -> `2024 - 2025`?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -0,0 +1,77 @@
...
+        return Iterator(fileDescriptor: self.fileDescriptor)
+    }
+}
+
+extension RangeReplaceableCollection {
+    /// Creates a new instance of a collection containing the elements of an asynchronous sequence.
+    ///
+    /// - Parameter source: The asynchronous sequence of elements for the new collection.
+    @inlinable
+    internal init<Source: AsyncSequence>(_ source: Source) async rethrows where Source.Element == Element {
```
parkera: I could see this being confusing for maintenance of the package in the future. Can we give the label an argument so it's clearer that we're doing here? Like:

```
let array = await Array(awaitingContentsOf: other)
```

Sources/Subprocess/Buffer.swift
```diff
@@ -0,0 +1,152 @@
...
+    ///   closure’s SequenceOutput.
+    /// - Returns: The return value, if any, of the body closure parameter.
+    public func withUnsafeBytes<ResultType>(
+        _ body: (UnsafeRawBufferPointer) throws -> ResultType
+    ) rethrows -> ResultType {
+        #if os(Windows)
+        return try self.data.withUnsafeBytes(body)
+        #else
+        // Although DispatchData was designed to be uncontiguous, in practice
+        // we found that almost all DispatchData are contiguous.
```
parkera: I think the important detail missing here is: because they are almost all contiguous, we can access this body in O(1) most of the time.

Sources/Subprocess/Buffer.swift
```diff
@@ -0,0 +1,152 @@
...
+#endif
+extension SequenceOutput.Buffer {
+    /// Access the raw bytes stored in this buffer
+    /// - Parameter body: A closure with an `UnsafeRawBufferPointer` parameter that
+    ///   points to the contiguous storage for the type. If no such storage exists,
+    ///   the method creates it. If body has a return value, this method also returns
+    ///   that value. The argument is valid only for the duration of the
+    ///   closure’s SequenceOutput.
+    /// - Returns: The return value, if any, of the body closure parameter.
+    public func withUnsafeBytes<ResultType>(
```
parkera, line 65: I'm tempted to say that this API should be *unavailable* when `SubprocessSpan` is on. There should be no reason to use the unsafe API over the safe one - I think you can still get an unsafe pointer from `RawSpan`.

Sources/Subprocess/Configuration.swift
```diff
@@ -0,0 +1,965 @@
...
+        }
+        if let errorError = errorError {
+            throw errorError
+        }
+    }
+}
+
+// MARK: - Executable
+
+/// `Executable` defines how should the executable
```
parkera:
```suggestion
/// `Executable` defines how the executable should
```

Sources/Subprocess/Configuration.swift
```diff
@@ -0,0 +1,965 @@
...
+        }
+    }
+
+    var stringValue: String {
+        return self ?? "nil"
+    }
+}
+
+// MARK: - Stubs for the one from Foundation
+public enum QualityOfService: Int, Sendable {
```
parkera: If this is supposed to be public, shouldn't it be nested inside the platform struct - and probably limited to Darwin only?

Sources/Subprocess/Configuration.swift
```diff
@@ -0,0 +1,965 @@
...
+    onCleanup handler: @Sendable @escaping () async -> Void,
+    isolation: isolated (any Actor)? = #isolation
+) async rethrows -> Result {
+    return try await withThrowingTaskGroup(
+        of: Void.self,
+        returning: Result.self
+    ) { group in
+        group.addTask {
+            // wait until cancelled
+            do { while true { try await Task.sleep(nanoseconds: 1_000_000_000) } } catch {}
```
parkera, line 944: Where does this timeout value come from?

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension Execution where Output == SequenceOutput {
+    /// The standard output of the subprocess.
+    /// Accessing this property will **fatalError** if
+    /// - `.output` wasn't set to `.redirectToSequence` when the subprocess was spawned;
+    /// - This property was accessed multiple times. Subprocess communicates with
+    ///   parent process via pipe under the hood and each pipe can only be consumed ones.
```
parkera:
```suggestion
    ///   parent process via pipe under the hood and each pipe can only be consumed once.
```

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+    #endif  // os(Windows)
+}
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension Execution where Output == SequenceOutput {
+    /// The standard output of the subprocess.
+    /// Accessing this property will **fatalError** if
+    /// - `.output` wasn't set to `.redirectToSequence` when the subprocess was spawned;
```
parkera: Is this line still valid? I think we prevented this via the type system now.

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension Execution where Error == SequenceOutput {
+    /// The standard error of the subprocess.
+    /// Accessing this property will **fatalError** if
+    /// - `.error` wasn't set to `.redirectToSequence` when the subprocess was spawned;
```
parkera: Same question here about if this line of docs is still valid.

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+    /// - This property was accessed multiple times. Subprocess communicates with
+    ///   parent process via pipe under the hood and each pipe can only be consumed ones.
```
parkera:
```suggestion
    ///   parent process via pipe under the hood and each pipe can only be consumed once.
```

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+    Error: OutputProtocol
+>: Sendable {
+    /// The process identifier of the current execution
+    public let processIdentifier: ProcessIdentifier
+
+    internal let output: Output
+    internal let error: Error
+    internal var outputPipe: CreatedPipe
+    internal var errorPipe: CreatedPipe
+    internal let outputConsumptionState: LockedState<OutputConsumptionState>
```
parkera: We should be able to remove all uses of LockedState and replace it with the Mutex from Synchronization. In fact, doing that may expose a threading bug - if we have a `Mutex` then we need the `Execution` type to be a class, or `~Copyable`.

Sources/Subprocess/Execution.swift
```diff
@@ -0,0 +1,200 @@
...
+#endif
+extension Execution where Output == SequenceOutput {
+    /// The standard output of the subprocess.
+    /// Accessing this property will **fatalError** if
+    /// - `.output` wasn't set to `.redirectToSequence` when the subprocess was spawned;
+    /// - This property was accessed multiple times. Subprocess communicates with
+    ///   parent process via pipe under the hood and each pipe can only be consumed ones.
+    public var standardOutput: some AsyncSequence<SequenceOutput.Buffer, any Swift.Error> {
+        let consumptionState = self.outputConsumptionState.withLock { stateStore in
+            let newState = stateStore.rawValue ^ OutputConsumptionState.standardOutputConsumed.rawValue
```
parkera: Can we just use `insert` here instead of bit-twiddling?

Package.swift
```diff
@@ -0,0 +1,95 @@
...
+// Enable SubprocessFoundation by default
+var defaultTraits: Set<String> = ["SubprocessFoundation"]
+#if compiler(>=6.2)
+// Enable SubprocessSpan when Span is available
+defaultTraits.insert("SubprocessSpan")
+#endif
+
+let package = Package(
+    name: "Subprocess",
+    platforms: [.macOS("15.0"), .iOS("18.0"), .tvOS("18.0"), .watchOS("11.0")],
```
parkera: I think this will need to be `15.4`, et. al. in order for `Span` to be available on Darwin.

Sources/Subprocess/IO/Input.swift
```diff
@@ -0,0 +1,289 @@
...
+    /// Signal all writes are finished
+    public func finish() async throws {
+        try self.fileDescriptor.safelyClose()
+    }
+}
+
+extension StringProtocol {
+    package func byteArray<Encoding: Unicode.Encoding>(using encoding: Encoding.Type) -> [UInt8]? {
+        if Encoding.self == Unicode.ASCII.self {
+            let isASCII = self.utf8.allSatisfy {
```
parkera, line 280: Does it make sense to do this check yourself here instead of just asking `String` to convert itself? It has fast paths for ASCII. Here we're going to have to go through the bytes twice at least.

Sources/Subprocess/IO/Input.swift
```diff
@@ -0,0 +1,289 @@
...
+        }
+        if Encoding.self == UTF16.self {
+            return Array(self.utf16).flatMap { input in
+                var uint16: UInt16 = input
+                return withUnsafeBytes(of: &uint16) { ptr in
+                    Array(ptr)
+                }
+            }
+        }
+        // TODO: More encodings
```
parkera: This might be a place where it makes sense to defer to Foundation, which has much richer support for this (when Foundation is imported).

Sources/Subprocess/IO/LockedState.swift
```diff
@@ -0,0 +1,168 @@
...
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
```
parkera: Discussed elsewhere, but we need to delete this file and stop the viral spread of copy/pasting it into every new package.

Sources/Subprocess/IO/Output.swift
```diff
@@ -0,0 +1,401 @@
...
+    /// The max amount of data to collect for this output.
+    var maxSize: Int { get }
+}
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension OutputProtocol {
+    /// The max amount of data to collect for this output.
+    public var maxSize: Int { 128 * 1024 }
```
parkera, line 49: Doesn't this max size only apply to String outputs? There is no max size for discarded, or file descriptor, right?

Sources/Subprocess/IO/Output.swift
```diff
@@ -0,0 +1,401 @@
...
+}
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension OutputProtocol where Self == StringOutput<UTF8> {
+    /// Create a `Subprocess` output that collects output as
+    /// UTF8 String with 128kb limit.
+    public static var string: Self {
+        .init(limit: 128 * 1024, encoding: UTF8.self)
```
parkera, line 244: Is this supposed to use the maxBytes property? Same for the byte output below.

Sources/Subprocess/IO/Output.swift
```diff
@@ -0,0 +1,401 @@
...
+            guard let fileDescriptor = fileDescriptor else {
+                // Show not happen due to type system constraints
+                fatalError("Trying to capture output without file descriptor")
+            }
+
+            fileDescriptor.wrapped.readUntilEOF(upToLength: self.maxSize) { result in
+                do {
+                    switch result {
+                    case .success(let data):
+                        //FIXME: remove workaround for rdar://143992296
```
parkera: Would be more useful to relate to a GitHub issue here than a radar.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -0,0 +1,365 @@
...
+        hasher.combine(groupID)
+        hasher.combine(supplementaryGroups)
+        hasher.combine(processGroupID)
+        hasher.combine(createSession)
+        // Since we can't really hash closures,
+        // use a random number such that as long as
+        // `preSpawnProcessConfigurator` is set, it will
+        // never equal to other PlatformOptions
+        if self.preSpawnProcessConfigurator != nil {
+            hasher.combine(Int.random(in: 0 ..< .max))
```
parkera: This doesn't seem right to me. Either the random number is stored in the type (although I can not think of any place where we used a random number for a closure hash), or the closure is not part of the `==` check, or we don't make the type hashable/equatable at all. With this implementation the hash is never going to be the same value, and it breaks the contract.

README.md
itingliu: From your proposal and inspiration from foundation's other packages

```suggestion
It improves the ergonomics of `Foundation.Process` and the experience of using Swift for scripting and other areas such as server-side development, with first-class integration with `async/await` and other language features.
```

README.md
```diff
@@ -1 +1,305 @@
...
+# Subprocess
+
+Subprocess is a cross-platform package for spawning processes in Swift.
+
+It's like [Foundation.Process](https://developer.apple.com/documentation/foundation/process), but written for Swift and build on top of structural concurrency.
+
+
+## Getting Started
+
+Subprocess uses [SwiftPM](https://swift.org/package-manager/) as its build tool, so we recommend using that as well. If you want to depend on Subprocess in your own project, it's as simple as adding a `dependencies` clause to your `Package.swift`:
```
itingliu: Can you build it *without* using swiftPM? Can we just omit it like the other packages if that's the default?

```suggestion
To use `Subprocess` in a [SwiftPM](https://swift.org/package-manager/) project, add it as a package dependency to your `Package.swift`:
```

README.md
```diff
@@ -1 +1,305 @@
...
+
+```swift
+dependencies: [
+    .package(url: "https://github.com/iCharlesHu/Subprocess.git", branch: "main")
+]
+```
+and then adding the `Subprocess` module to your target dependencies.
```
itingliu:
```suggestion
Then, add the `Subprocess` module to your target dependencies:
```

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -0,0 +1,415 @@
...
+        hasher.combine(self.groupID)
+        hasher.combine(self.supplementaryGroups)
+        hasher.combine(self.processGroupID)
+        hasher.combine(self.createSession)
+        // Since we can't really hash closures,
+        // use an random number such that as long as
+        // `preSpawnProcessConfigurator` is set, it will
+        // never equal to other PlatformOptions
+        if self.preSpawnProcessConfigurator != nil {
+            hasher.combine(Int.random(in: 0 ..< .max))
```
parkera: Commented on the other file for Linux, but using a random number here does not seem correct.

README.md
```diff
@@ -1 +1,305 @@
...
+    environment: .inherit.updating(["NewKey": "NewValue"]),
+    workingDirectory: "/Users/",
+)
+```
+
+### Platform Specific Options and “Escape Hatches”
+
+`Subprocess` provides **platform-specific** configuration options, like setting `uid` and `gid` on Unix and adjusting window style on Windows, through the `PlatformOptions` struct. Check out the `PlatformOptions` documentation for a complete list of configurable parameters across different platforms.
+
+Besides these platform-specific settings, `PlatformOptions` also includes an “escape hatch” via a closure. This closure allows you to have access to low level platform specific spawning constructs to perform customizations if `Subprocess` doesn’t have higher-level APIs.
```
itingliu:
```suggestion
`PlatformOptions` also allows access to platform-specific spawning constructs and customizations via a closure.
```

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,545 @@
...
+        uidPtr: UnsafeMutablePointer<uid_t>?,
+        gidPtr: UnsafeMutablePointer<gid_t>?,
+        supplementaryGroups: [gid_t]?
+    ) {
+        // Prepare environment
+        let env = self.environment.createEnv()
+
+        var uidPtr: UnsafeMutablePointer<uid_t>? = nil
+        if let userID = self.platformOptions.userID {
+            uidPtr = .allocate(capacity: 1)
```
parkera, line 341: Isn't this a memory leak?

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+internal typealias PlatformFileDescriptor = FileDescriptor
+
+// MARK: - Read Buffer Size
+@inline(__always)
+internal var readBufferSize: Int {
+    #if canImport(Darwin)
+    return 16384
```
parkera: I'm not actually sure this is consistent across all Darwin platforms, either.

README.md
```diff
@@ -1 +1,305 @@
...
+
+let result = try await run(.path("/bin/exe"), platformOptions: platformOptions)
+```
+
+
+### Flexible Input and Output Configurations
+
+By default, `Subprocess`:
+- Doesn’t send any input to the child process’s standard input
+- Captures the child process’s standard output as a `String`, up to 128kb
```
itingliu:
```suggestion
- Captures the child process’s standard output as a `String`, up to 128kB
```

README.md
```diff
@@ -1 +1,305 @@
...
+let content = "Hello Subprocess"
+
+// Send "Hello Subprocess" to the standard input of `cat`
+let result = try await run(.name("cat"), input: .string(content, using: UTF8.self))
+
+// Collect both standard error and standard output as Data
+let result = try await run(.name("cat"), output: .data, error: .data)
+```
+
+`Subprocess` ships with these input options:
```
itingliu:
```suggestion
`Subprocess` supports these input options:
```

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -0,0 +1,1287 @@
...
+        hasher.combine(userCredentials)
+        hasher.combine(consoleBehavior)
+        hasher.combine(windowStyle)
+        hasher.combine(createProcessGroup)
+        // Since we can't really hash closures,
+        // use a random number such that as long as
+        // `preSpawnProcessConfigurator` is set, it will
+        // never equal to other PlatformOptions
+        if self.preSpawnProcessConfigurator != nil {
+            hasher.combine(Int.random(in: 0 ..< .max))
```
parkera: Same comment on using random in hash here. Didn't want to forget about this one.

README.md
```diff
@@ -1 +1,305 @@
...
+
+#### `NoInput`
+
+This option means no input is sent to the subprocess.
+
+Use it by setting `.none` for `input`.
+
+#### `FileDescriptorInput`
+
+This option reads input from a specified `FileDescriptor` you provide. If `closeAfterSpawningProcess` is set to `true`, the subprocess will close the file descriptor after spawning. If `false`, you need to close it, even if the subprocess fails to spawn.
```
itingliu:
```suggestion
This option reads input from a specified `FileDescriptor`. If `closeAfterSpawningProcess` is set to `true`, the subprocess will close the file descriptor after spawning. If `false`, you are responsible for closing it, even if the subprocess fails to spawn.
```

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -0,0 +1,1287 @@
...
+                }
+            }
+        }
+    }
+
+    internal func readUntilEOF(
+        upToLength maxLength: Int,
+        resultHandler: @Sendable @escaping (Swift.Result<[UInt8], any (Error & Sendable)>) -> Void
+    ) {
+        DispatchQueue.global(qos: .userInitiated).async {
```
parkera, line 1039: What does a QoS of user initiated on a dispatch queue even mean on Windows?

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -0,0 +1,1287 @@
...
+                completion(Int(writtenBytes), error)
+            } else {
+                completion(Int(writtenBytes), nil)
+            }
+        }
+    }
+}
+
+extension Optional where Wrapped == String {
+    fileprivate func withOptionalCString<Result, Encoding>(
```
parkera, line 1164: I think this is just a reimplementation of `map` on Optional?

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -0,0 +1,1287 @@
...
+        switch self {
+        case .none:
+            return try body(nil)
+        case .some(let value):
+            return try value.withNTPathRepresentation(body)
+        }
+    }
+}
+
+// MARK: - Remove these when merging back to SwiftFoundation
```
parkera, line 1188: Not sure if this comment is still relevant.

Sources/Subprocess/Span+Subprocess.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+// swift-format-ignore-file
+
+#if SubprocessSpan
+
+@_unsafeNonescapableResult
```
parkera, line 16: Should this be public? Why do we need it?

Sources/Subprocess/SubprocessFoundation/Input+Foundation.swift
```diff
@@ -0,0 +1,177 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if SubprocessFoundation
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#elseif canImport(Foundation)
```
parkera: I'm not sure about the way this is written. It's more like - on Darwin we want to import the system Foundation always, and elsewhere we always want FoundationEssentials and not swift-corelibs-foundation (Foundation).

Sources/Subprocess/SubprocessFoundation/Output+Foundation.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
+extension OutputProtocol where Self == DataOutput {
+    /// Create a `Subprocess` output that collects output as `Data`
+    /// up to 128kb.
+    public static var data: Self {
+        return .data(limit: 128 * 1024)
```
parkera, line 54: Should this be using `maxSize`?

Sources/Subprocess/SubprocessFoundation/Span+SubprocessFoundation.swift
```diff
@@ -0,0 +1,74 @@
...
+
+internal import Dispatch
+
+@available(SubprocessSpan, *)
+extension Data {
+    init(_ s: borrowing RawSpan) {
+        self = s.withUnsafeBytes { Data($0) }
+    }
+
+    public var bytes: RawSpan {
```
parkera, line 30: This should be available in Foundation itself soon - but then, only on 6.2 and later... I'm not sure this package is the right place to add a public API to Foundation, though. Won't it end up being ambiguous?

Sources/_SubprocessCShims/include/process_shims.h
```diff
@@ -0,0 +1,83 @@
...
+    const char * _Nonnull format,
+    char * _Nonnull str1,
+    char * _Nonnull str2
+);
+#endif
+
+#endif // !TARGET_OS_WINDOWS
+
+#if TARGET_OS_WINDOWS
+#include <windows.h>
```
parkera: Do we actually need to import all of the windows headers here? Is it just for `DWORD` or does it flow into the rest of the code? Can we import the Windows modules there instead?

README.md
```diff
@@ -1 +1,305 @@
...
+
+#### `DataAsyncSequenceInput` (available with `SubprocessFoundation` trait)
+
+This option reads input from an async sequence of `Data`.
+
+Use it by setting `.asyncSequence` for `input`.
+
+---
+
+`Subprocess` also ships these output options:
```
itingliu:
```suggestion
`Subprocess` also supports these output options:
```

README.md
```diff
@@ -1 +1,305 @@
...
+
+This option redirects the child output to the `.standardOutput` or `.standardError` property of `Execution`. It’s only for the `run()` family that takes a custom closure.
+
+
+### Cross-platform support
+
+`Subprocess` works on all major platforms supported by Swift, including macOS, Linux, and Windows, with feature parity on all platforms as well as platform-specific options for each.
+
+The table below describes the current level of support that Subprocess has
+for various platforms:
```
itingliu: Inconsistent line wrapping here

README.md
```diff
@@ -1 +1,305 @@
...
+```
+swift package --disable-sandbox preview-documentation --target Subprocess
+```
+
+<p align="right">(<a href="#readme-top">back to top</a>)</p>
+
+
+## Contributing to Subprocess
+
+Subprocess is part of the Foundation project. We have a dedicated [Foundation Forum][forum] where people can ask and answer questions on how to use or work on this package. It's also a great place to discuss its evolution.
```
itingliu:
```suggestion
Subprocess is part of the Foundation project. Discussion and evolution take place on [Swift Foundation Forum](https://forums.swift.org/c/related-projects/foundation/99).
```

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -0,0 +1,675 @@
...
+        // Child process
+        close(pipefd[0]);  // Close unused read end
+
+        // Perform setups
+        if (number_of_sgroups > 0 && sgroups != NULL) {
+            if (setgroups(number_of_sgroups, sgroups) != 0) {
+                int error =  errno;
+                write(pipefd[1], &error, sizeof(error));
+                close(pipefd[1]);
+                _exit(EXIT_FAILURE);
```
parkera, line 166: Is exiting the process really the right response to an error here? I'm honestly not sure. (I assume this is in the child process)

Tests/SubprocessTests/SubprocessTests+Linux.swift
```diff
@@ -0,0 +1,88 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Glibc) || canImport(Bionic) || canImport(Musl)
+
+import Foundation
```
parkera: Although it's fine to depend on the swift-corelibs-foundation Foundation here, I am curious what was missing from FoundationEssentials that you needed in this file on non-Darwin.

Tests/TestResources/Resources/windows-tester.ps1
```diff
@@ -0,0 +1,39 @@
+param (
```
parkera, line 12: Missing copyright header here

Tests/TestResources/Resources/getgroups.swift
```diff
@@ -0,0 +1,22 @@
+#if canImport(Darwin)
```
parkera, line 12: Missing copyright header here

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: Actually I just realized it's not used anywhere so I removed it.

Sources/Subprocess/Buffer.swift
iCharlesHu: I agree

Sources/Subprocess/Configuration.swift
iCharlesHu: Ahh good catch. Thanks!

Sources/Subprocess/Configuration.swift
iCharlesHu: Added a comment.

Sources/Subprocess/Execution.swift
iCharlesHu: You're absolutely right. For now I think I'll just stick with `class` but I would love to figure out how to mark `Execution` `~Copyable` in the long run.

Package.swift
iCharlesHu: Discussed offline and removed the platform requirements here.

Sources/Subprocess/IO/Input.swift
iCharlesHu: The idea here is to return `nil` if self is not all ASCII and ASCII is requested. I don't think `String` offers any public API to check if it's all ASCII

Sources/Subprocess/IO/Input.swift
iCharlesHu: Ahh good idea thanks!

Sources/Subprocess/IO/Output.swift
iCharlesHu: And also `BytesOutput`. But correct, no max size for `discarded` nor `file descriptor`.

Sources/Subprocess/IO/Output.swift
iCharlesHu: No since this is a static var. This is the "short cut" for the default collection size and encoding case. `maxSize` is set if you call `.string(limit: 42)` which specifies a new limit.

Sources/Subprocess/IO/Output.swift
iCharlesHu: Sounds good I'll create a mirror issue

Sources/Subprocess/Execution.swift
iCharlesHu: I switched to using `Atomic + .bitwiseXor`. I want to avoid TOCTOU issue here by making setting the state and checking one atomic operation `bitwiseXor`. This way I only need to check whether the state is set once.

Sources/Subprocess/IO/Output.swift
iCharlesHu: https://github.com/swiftlang/swift-subprocess/issues/3

Sources/Subprocess/Execution.swift
parkera: You're inside a lock - no one better be mutating it at the same time anyway.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: Initially I thought `Hashable`  conformance for `Configuration` would be useful. But I do agree with you this solution doesn't seem right either so now I'm leaning towards removing `Hashable` conformance from `Configuration` and `PlatformOptions` since we can't truly hash `PlatformOptions` with closures.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
iCharlesHu: We allocate here and in the call site of `preSpawn` we deallocate them:
```swift
let (
    env,
    uidPtr,
    gidPtr,
    supplementaryGroups
) = try self.preSpawn()
defer {
    for ptr in env { ptr?.deallocate() }
    uidPtr?.deallocate()
    gidPtr?.deallocate()
}
```

iCharlesHu: But I think maybe I'll change this to a closure based approach.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
iCharlesHu: It's mostly here to make the call site across different platforms look more consistent.

Sources/Subprocess/Span+Subprocess.swift
iCharlesHu: This entire file will go away as soon as Guillaume's PR to add these properties (`.bytes`) lands on Array and friends. Until then this is code from the standard library essentially.

Sources/Subprocess/SubprocessFoundation/Input+Foundation.swift
iCharlesHu: Ahh okay. I think in that case it should be
```
#if canImport(Darwin)
// On Darwin always prefer system Foundation
import Foundation
#else
// On other platforms prefer FoundationEssentials
#if canImport(FoundationEssentials)
import FoundationEssentials
#elseif canImport(Foundation)
import Foundation
#endif

#endif // canImport(Darwin)
```

Sources/Subprocess/SubprocessFoundation/Output+Foundation.swift
iCharlesHu: No. This is setting what `maxSize` is here.

Sources/Subprocess/SubprocessFoundation/Span+SubprocessFoundation.swift
iCharlesHu: No all these are workarounds until they land in Foundation. My plan is to remove these `Span+` workarounds as soon as it lands in Foundation and the standard library.

Sources/_SubprocessCShims/include/process_shims.h
iCharlesHu: We actually do need `<windows.h>` for other functions like `GetWindowThreadProcessId`. We use them in `process_shims.c`.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Yes. This is the standard `fork/exec` pattern that you terminate the child process asap if `exec` fails to avoid leaving child hanging around in a zombie state.

Tests/SubprocessTests/SubprocessTests+Linux.swift
iCharlesHu: I believe it's for `Data` in `testSuspendResumeProcess`.

Sources/_SubprocessCShims/include/process_shims.h
parkera: We can import it in the .c file if we need it there, though - that'll limit the import to just one place, which could be useful for preventing leakage of that usage elsewhere.

Tests/SubprocessTests/SubprocessTests+Linux.swift
parkera: `Data` is in `FoundationEssentials` though.

Sources/_SubprocessCShims/include/process_shims.h
iCharlesHu: Ahh I see your point. Will do.

Tests/SubprocessTests/SubprocessTests+Linux.swift
iCharlesHu: Ahh that's what you meant. I only needed `Data` I just forgot that `FoundationEssentials` is now in the toolchain. Updated to `import FoundationEssentials`

README.md
iCharlesHu: Tony actually suggested to remove this line altogether 😅

README.md
iCharlesHu: > Can you build it without using swiftPM?

Not easily since there's also the c bit so they'll have to set up module maps manually. 

With that said, I do love you version better. Thanks!

PR #4, opened 2025-04-11
Add LockedState as fallback when Synchronization is not available

iCharlesHu:


iCharlesHu:
Resolves #7

parkera (approved):

gwynne:
At the moment, this doesn't actually resolve the problem; `#if canImport(Synchronization)` still succeeds when the macOS 15 SDK is in use, regardless of the `.macOS` value in `Package.swift`.

PR #6, opened 2025-04-16
[CI] Add support for GitHub Actions

shahmishal:


Sources/_SubprocessCShims/include/target_conditionals.h
```diff
@@ -6,7 +6,6 @@
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
-// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
```
iCharlesHu, line 9: Is this change intentional? Should we remove it from all headers?

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  tests:
+    name: Test
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@macos-support
+    with:
+      linux_swift_versions: '["nightly-main"]'
+      windows_swift_versions: '["nightly-main"]'
+      enable_macos_checks: true
+      macos_xcode_versions: '["16.2"]'
```
iCharlesHu: Ideally we would want a `macOS 13` VM to test backwards compatibility (since this package needs to support macOS 13, which does not have `Span`) and also a `nightly` to test Span integration. But having nightly for now is enough.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@macos-support
+    with:
+      linux_swift_versions: '["nightly-main"]'
```
iCharlesHu: Do we have to specify here that we are using `swift test` and `swift build`?

Sources/_SubprocessCShims/include/target_conditionals.h
shahmishal: Yes

.github/workflows/pull_request.yml
shahmishal: We currently don't support backwards deployment.

.github/workflows/pull_request.yml
shahmishal: we are using `swift test`, however we can overwrite it to use `swift build`

Issue #7, opened 2025-04-18
Bump up the macOS platform to .macOS(.v15) in Package.swift & Package@swift-6.0.swift


IgorMuzyka:
**Bug**
Currently both [`Package.swfit`](https://github.com/swiftlang/swift-subprocess/blob/main/Package.swift) and [`Package@swift-6.0.swift`](https://github.com/swiftlang/swift-subprocess/blob/main/Package%40swift-6.0.swift) state **platforms** as `.macOS(.v13)`.
The following lines in [`Execution.swift`](https://github.com/swiftlang/swift-subprocess/blob/main/Sources/Subprocess/Execution.swift) raise errors: 49, 82, 96, 97, 99, 120, 121, 123.
The errors are:
> 'X' is only available in macOS 15.0 or newer

**To Reproduce**
Just add this package as dependency to another package and it's target and see build fail.

**Expected behavior**
Correct platform version.

**Environment (please complete the following information):**
 - macOS `15.4.1`
 - Xcode `16.3`
 - Swift version `swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0`

fpseverino:
If this library could also work on macOS 13 that would be ideal, would that be possible?

There are a lot of projects, CLI tools and various libraries, that need to support macOS 13 and would benefit greatly from using the new Subprocess

iCharlesHu:
@fpseverino yes that's the goal. See #4

PR #8, opened 2025-04-18
Bumped macOS platfrom from .v13 to v15

IgorMuzyka:
Fixes issue #7

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,7 @@ defaultTraits.insert("SubprocessSpan")
 
 let package = Package(
     name: "Subprocess",
-    platforms: [.macOS(.v13)],
+    platforms: [.macOS(.v15)],
     products: [
         .library(
             name: "Subprocess",
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -9,7 +9,7 @@ let availabilityMacro: SwiftSetting = .enableExperimentalFeature(
 
 let package = Package(
     name: "Subprocess",
-    platforms: [.macOS(.v13)],
+    platforms: [.macOS(.v15)],
     products: [
         .library(
             name: "Subprocess",


iCharlesHu:
Hi @IgorMuzyka ! I actually have a in flight PR to fix this issue:
https://github.com/swiftlang/swift-subprocess/pull/4

Please give it a try to see if that fixes it for you.

iCharlesHu:
Unfortunately this package needs to maintain backwards compatibility to macOS 13 (hence the traits and different `Package.swift` file) so we can't simply bump the target platform to macOS 15.

IgorMuzyka:
@iCharlesHu just swapped Subprocess with your fork like this:
`.package(url: "https://github.com/iCharlesHu/Subprocess", branch: "charles/mutex-availability", traits: ["SubprocessFoundation"]),`

Here's the result:
screenshot 2025-04-18 at 21 42 08

Same error.

**I acknowledge** that bumping version is rather crude and simplistic solution not a viable if you have to maintain backward compatibility with macOS 13. 
But i guess it'll have to do for me for now(cause i quite literally already swapped [`SwiftShell`](https://github.com/kareman/SwiftShell) for [`Subprocess`](https://github.com/swiftlang/swift-subprocess) in my current project today lol).

The issue with macOS 15 requirement can be further separated into two parts:
First is `AtomicBox` requiring `macOS 15` [here](https://github.com/iCharlesHu/Subprocess/blob/361ce946e1c30eff23473d588079365bc0417600/Sources/Subprocess/Atomic.swift#L56).
```swift
#if canImport(Synchronization)
@available(macOS 15, *)
extension Atomic: AtomicBoxProtocol where Value == UInt8
```
Second is `AsyncSequence` having Failure generic only since `macOS 15`:
```swift
@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
public protocol AsyncSequence<Element, Failure> { ....
```

Feel free to close this issue and reject my pull request.

Also disregarding current issue.
I already had a suite of tests to check if my code works as expected with `SwiftShell`, and the first and most non obvious thing i ran into is not having `userID` and `groupID` set in `PlatformOptions` which resulted in not being able to run executable with arguments in directory just created by the code few lines above.
I've come up with following:
```swift
#if canImport(Darwin)
import Darwin
#elseif canImport(Bionic)
import Bionic
#elseif canImport(Glibc)
import Glibc
#elseif canImport(Musl)
import Musl
#endif

public extension PlatformOptions {
    static var inherited: PlatformOptions {
        var options = PlatformOptions()
        #if canImport(Darwin) || canImport(Glibc) || canImport(Bionic) || canImport(Musl)
            options.userID = getuid()
            options.groupID = getgid()
        #endif
        return options
    }
}
```

IgorMuzyka:
@iCharlesHu come up with alternative workaround.

Issue #14, opened 2025-04-20
Does not compile on macOS 15 + Swift 6.1


sebsto:
**Describe the bug**
`git clone && swift build`  fails on macOS 15.4 + Swift 6.1

**To Reproduce**
1. `git clone https://github.com/swiftlang/swift-subprocess`
2. `cd swift-subprocess`
3. `swift build`

**Expected behavior**
Should compile out of the box

**Environment (please complete the following information):**
```
» sw_vers  
ProductName:		macOS
ProductVersion:		15.4.1
BuildVersion:		24E263

» uname -a 
Darwin bcd07458a91c 24.4.0 Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:47 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6000 arm64
```

 - Swift version (run `swift --version`) [e.g. swiftlang-6.2.0.1.23 clang-1700.3.1.3]
```
» swift --version 
Apple Swift version 6.1 (swift-6.1-RELEASE)
Target: arm64-apple-macosx15.0
```

**Additional context**
```
  » git clone https://github.com/swiftlang/swift-subprocess
Cloning into 'swift-subprocess'...
remote: Enumerating objects: 127, done.
remote: Counting objects: 100% (127/127), done.
remote: Compressing objects: 100% (89/89), done.
remote: Total 127 (delta 45), reused 105 (delta 34), pack-reused 0 (from 0)
Receiving objects: 100% (127/127), 573.34 KiB | 5.91 MiB/s, done.
Resolving deltas: 100% (45/45), done.
bcd07458a91c :: code/swift/bedrock » swift --version 
Apple Swift version 6.1 (swift-6.1-RELEASE)
Target: arm64-apple-macosx15.0

» cd swift-subprocess/                                                                                                   

» swift build         
Fetching https://github.com/apple/swift-system from cache
Fetching https://github.com/apple/swift-docc-plugin from cache
Fetched https://github.com/apple/swift-docc-plugin from cache (0.81s)
Fetched https://github.com/apple/swift-system from cache (1.05s)
Computing version for https://github.com/apple/swift-docc-plugin
Computed https://github.com/apple/swift-docc-plugin at 1.4.3 (2.34s)
Fetching https://github.com/swiftlang/swift-docc-symbolkit from cache
Fetched https://github.com/swiftlang/swift-docc-symbolkit from cache (0.85s)
[...]

PR #16, opened 2025-04-22
Fix build on macOS 13 while maintaining minimal compatibility

iCharlesHu:
This PR introduces `AtomicBox` as an internal abstraction on atomic values. It uses `Atomic` on platforms where `Synchronization` is available, and falls back to `LockedState` otherwise.

I also had to expose `AsyncBufferSequence` as a public type due to a change in `AsyncSequence`: `AsyncSequence`'s [`Failure` ](https://github.com/swiftlang/swift/blob/e9253b749ff7cd4365985cd5e6ae773ae39d461d/stdlib/public/Concurrency/AsyncSequence.swift#L83) property has newer availability than `AsyncSequence` itself. This means we can't vend properties such as `var standardOutput: some AsyncSequence<Buffer, any Swift.Error>` on platforms "sandwiched" between the two availabilities (i.e. macOS 13) because the compiler requires this property to be at least the same availability as `AsyncSequence.Failure`. Unfortunately, there are no known workarounds besides just exposing the concrete type directly.

Tested on macOS 13.5 + Swift 6.0 / Swift 6.1

### DIFF
--- a/Sources/Subprocess/AsyncBufferSequence.swift
+++ b/Sources/Subprocess/AsyncBufferSequence.swift
@@ -10,7 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 #if canImport(System)
-import System
+@preconcurrency import System
 #else
 @preconcurrency import SystemPackage
 #endif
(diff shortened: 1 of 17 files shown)


iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/14

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -10,22 +10,21 @@
...
 #endif
 
 #if SubprocessSpan
 @available(SubprocessSpan, *)
 #endif
-internal struct AsyncBufferSequence: AsyncSequence, Sendable {
-    internal typealias Failure = any Swift.Error
-
-    internal typealias Element = SequenceOutput.Buffer
+public struct AsyncBufferSequence: AsyncSequence, Sendable {
```
itingliu, line 21: I wonder how we can remind ourselves better why this is now an API since it's not covered in the proposal

itingliu (approved):

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: I will start the CHANGELOG file to track API changes.

PR #18, opened 2025-04-23
Remove dead code causing a failure on Windows

jakepetroules:
This appears to be unused, and causes the build to fail on Windows with Swift 6.1.


iCharlesHu (approved):

iCharlesHu:
Thanks @jakepetroules !

PR #24, opened 2025-04-24
Test fixes and intermittent Linux failures

cmcgee1024:
There is a test failing when using swiftly because the swift binary isn't in
"/usr/bin/swift". Change the test to use the simple name "swift" for the
executable, finding the swift that's on the PATH.

Tests are intermittently failing on PATH lookups for fully qualified path
executables specified in the test, such as `/usr/bin/id`. These should be
full `.path()` executables instead.

Add EACCES as another type of error that can come back through execve()
because of permissions or irregular file types. When this error is encountered
continue to the next possible path.


Sources/Subprocess/Configuration.swift
```diff
@@ -871,7 +871,11 @@ internal struct CreatedPipe {
     }
 
     internal init(closeWhenDone: Bool) throws {
+#if os(Windows)
```
jakepetroules: This is already fixed by #23

.swift-version
```diff
@@ -0,0 +1 @@
+main-snapshot-2025-04-12
```
jakepetroules: I really don't understand the purpose of .swift-version... it doesn't really make sense for things that aren't a _product_ (like libraries) and are expected to work across many different Swift versions.

And if we're fixing the compile error right now, what's the value in pinning to a snapshot anyways?

cmcgee1024: I'm just trying to build and generate the documentation for this project and I couldn't get it to work on Linux reliably with the usual 6.1 toolchain. It might have been a weird bug I was encountering though so I can change this back to 6.1.0 for other developers working with this project.

The swift version file helps with development environments, and CI if it's configured to get everyone on a known good toolchain for developing the project.

jakepetroules: With #23 this builds successfully for me on Linux with Swift 6.1, although `runLinter()` hangs and `testSubprocessPlatfomOptionsPreSpawnProcessConfigurator()` and `testSuspendResumeProcess()` fail. Is that what you're seeing as well?

cmcgee1024: Yes, also `testSubprocessPlatformOptionsUserID` is failing with an `EACCES` errno intermittently.

cmcgee1024: `testExecutableNamed` is also hanging periodically

.swift-version
```diff
@@ -0,0 +1 @@
+6.1.0
```
iCharlesHu, line 1: Could elaborate more on what this file does? I don't think `6.1.0` is an appropriate version because we need 6.2 for `Span` and the majority of the package depends on that trait (i.e. would require 6.2).

cmcgee1024: This file marks what the standard swift toolchain version for this project should be for development, and testing. So, if I were developing with this package, what toolchain should I use to `swift build` and `swift test` and expect it to work.

If a developer has swiftly installed then swiftly will use it automatically. If CI supports it then that will be the toolchain that it uses as well. This has no effect on downstream packages.

https://www.swift.org/swiftly/documentation/swiftly/use-toolchains#Sharing-recommended-toolchain-versions

cmcgee1024: If 6.2 of Swift is required to develop/test subprocess, then the swift version file could have a particular `main-snapshot-xxxx-yy-zz` that's known to be good at a point in time, possibly tested by CI.

cmcgee1024:
@iCharlesHu is there anything else I can do to help get this PR merged?

.swift-version
iCharlesHu: Ahh thanks. In that case let's leave it at `6.1.0` until we officially releases `6.2`.

iCharlesHu (approved):

iCharlesHu:
> @iCharlesHu is there anything else I can do to help get this PR merged?

LGTM. Thanks for the fix!

PR #30, opened 2025-04-24
Block the package from building on iOS-derived platforms

jakepetroules:
Process spawning is not available there. Also remove a couple overly-specific platform conditions in tests that would be incorrect on other Unix-like platforms such as Android and FreeBSD.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@
 import PackageDescription
 
 let availabilityMacro: SwiftSetting = .enableExperimentalFeature(
-    "AvailabilityMacro=SubprocessSpan: macOS 9999",
+    "AvailabilityMacro=SubprocessSpan: macOS 9999, iOS 9999, tvOS 9999, watchOS 9999, visionOS 9999",
 )
 
 var dep: [Package.Dependency] = [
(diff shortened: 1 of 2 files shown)


lynchsft (approved):

jakepetroules:
@iCharlesHu Can I get an approval, please?

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -166,6 +166,9 @@ extension Configuration {
         error: Error,
         errorPipe: CreatedPipe
     ) throws -> Execution<Output, Error> {
+        #if !os(macOS)
```
itingliu: The motivation of this change makes sense, but this runtime error inside this function is very non obvious to me.

itingliu (commented):
If we are to do this should we add an iOS CI (if possible at all)? Otherwise how would we know if a change causes the project to fail to build on iOS?

jakepetroules:
OK, I've made those adjustments.

lynchsft (approved):

iCharlesHu (approved):

PR #31, opened 2025-04-25
Fix currentWorkingDirectory

kojirou1994:
String.init(cString:) may change the content

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -888,7 +888,7 @@ extension FilePath {
     static var currentWorkingDirectory: Self {
         let path = getcwd(nil, 0)!
         defer { free(path) }
-        return .init(String(cString: path))
+        return .init(platformString: path)
     }
 }
 


jakepetroules (approved):

jakepetroules (changes requested):
Actually, this will break Windows because the platform string types won't match (UTF-16 != UTF-8). And Windows needs to use GetCurrentDirectoryW in order to support Unicode paths.

As an aside, according to POSIX, passing nil to getcwd is undefined behavior.

Essentially, we need to copy this: https://github.com/swiftlang/swift-foundation/blob/7002497ffceeeaf4916e06fd206724f1f801d2b5/Sources/FoundationEssentials/FileManager/FileManager%2BDirectories.swift#L494

jakepetroules:
@iCharlesHu Is there any reason not to use `FileManager.default.currentDirectoryPath` for this?

iCharlesHu:
> @iCharlesHu Is there any reason not to use `FileManager.default.currentDirectoryPath` for this?

This package can't hard depend on Foundation because it's optional via a trait.

jakepetroules:
In #74 we opted to make an API change which resulted in removing FilePath.currentWorkingDirectory entirely. So this fix is no longer necessary.

PR #33, opened 2025-04-25
Fix hang in Linux process termination handler

jakepetroules:
Calling waitid inside the _childProcessContinuations lock can cause deadlocks in setting up the termination handlers of one process while waiting for another to complete. Instead, only hold the lock when the continuations map is actually being accessed.


iCharlesHu (approved):
Thanks for the fix!

PR #35, opened 2025-04-28
Block signal masks pre-fork and restore post-fork.

iCharlesHu:
This PR introduces proper signal blocking and restoring for custom `fork/exec` on Linux.

Resolves: https://github.com/iCharlesHu/Subprocess/issues/45
Resolves: https://github.com/iCharlesHu/Subprocess/issues/46

—

During testing, one of the newly added tests revealed that we were not managing `DispatchIO` correctly during early terminations on Darwin and Linux. Currently, we simply close the file descriptors in the clean-up handler, which is incorrect because `DispatchIO` is used for read/write operations on these file descriptors. Modifying the file descriptor after it has been “transferred” to `DispatchIO` is considered a programmer error.

Rearchitect `TrackedFileDescriptor` (now renamed `DiskIO`) to either hold a `FileDescriptor` or a `DispatchIO`. We create `DispatchIO` as the final step of `spawn()` and use it for read/write operations and cleanup, instead of using a raw file descriptor. This allows `DispatchIO` to naturally close the file descriptor itself.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -101,7 +101,7 @@ public struct Configuration: Sendable {
             // Body runs in the same isolation
             let result = try await body(
                 execution,
-                .init(fileDescriptor: execution.inputPipe.writeFileDescriptor!)
+                .init(diskIO: execution.inputPipe.writeEnd!)
             )
             return ExecutionResult(
                 terminationStatus: try await waitingStatus,
@@ -165,7 +165,7 @@ public struct Configuration: Sendable {
                 standardError
             ) = try await execution.captureIOs()
             // Write input in the same scope
-            guard let writeFd = execution.inputPipe.writeFileDescriptor else {
+            guard let writeFd = execution.inputPipe.writeEnd else {
                 fatalError("Trying to write to an input that has been closed")
             }
             try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, any Swift.Error>) in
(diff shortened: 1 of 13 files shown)


iCharlesHu:
@swift-ci please test

Sources/Subprocess/Configuration.swift
```diff
@@ -822,27 +801,51 @@ internal enum StringOrRawBytes: Sendable, Hashable {
     }
 }
 
-/// A simple wrapper on `FileDescriptor` plus a flag indicating
-/// whether it should be closed automactially when done.
-internal struct TrackedFileDescriptor: Hashable {
+/// A wrapped `FileDescriptor` or `DispatchIO` and
+/// whether it should beeddsw closed automactially when done.
+internal struct DiskIO {
```
parkera: I'm not sure I like this direction. These really feel like two types smashed into a hidden enum. I think it will feel better if we create a new type for tracking the dispatch IO and use it where appropriate instead.

iCharlesHu:
Instead of a monolithic `DiskIO`, split `CreatedPipe` to `InputPipe` and `OutputPipe` so we don't have to rely on `guard` and `fatalError()` at runtime to guard against scenarios that shouldn't happen.

iCharlesHu:
@swift-ci please test

Sources/Subprocess/Configuration.swift
```diff
@@ -822,17 +801,17 @@ internal enum StringOrRawBytes: Sendable, Hashable {
     }
 }
 
-/// A simple wrapper on `FileDescriptor` plus a flag indicating
+/// A wrapped `FileDescriptor` or `DispatchIO` and
```
parkera: Is this still true?

parkera (approved):

Issue #34, opened 2025-04-25
runLinter() fails very frequently on Linux


jakepetroules:
**Describe the bug**
Running this test results in the swift-format process never exiting (it shows it's stuck in writing to stderr). Closing the parent file descriptors after the call to `captureIOs()` seems to work around this, but I don't think that's correct and I'm not sure what's going on here.

**To Reproduce**
`swift test --filter runLinter()`

**Expected behavior**
Test does not hang.

**Environment (please complete the following information):**
- Linux debian 6.1.0-31-arm64 #\1 SMP Debian 6.1.128-1 (2025-02-07) aarch64 GNU/Linux
- Debian GNU/Linux 12.9 (bookworm)
- Swift version 6.1 (swift-6.1-RELEASE)
- Target: aarch64-unknown-linux-gnu

**Additional context**

PR #36, opened 2025-05-01
Disable linter test until we decided on the rules

iCharlesHu:


jakepetroules (commented):
Should this really close #34? I expect that hang is a legitimate bug in the underlying implementation that will manifest in other ways.

iCharlesHu:
> Should this really close #34? I expect that hang is a legitimate bug in the underlying implementation that will manifest in other ways.

Good point. I'll leave that issue open to investigate.

Issue #37, opened 2025-05-01
`close` _ignores_ EBADF (`.badFileDescriptor`) which is super dangerous, needs to preconditionFailure this


weissi:
https://github.com/swiftlang/swift-subprocess/blob/c0bbfa57d0ad36c993e8aa7be909b46c919403f3/Sources/Subprocess/Configuration.swift#L829-L831

This is an extremely dangerous line of code. If you every hit `EBADF` on `close` then you know all bets are off and the only thing you should do is crash. Why so extreme? If some piece of code lost track of the file descriptors and closed 'the wrong one', security vulnerabilities will happen.

For example:

A confused piece of code accidentally double-closes
```swift
let fd = openSomething() // A
defer {
    try? close(fd) // B
}

[...]

write(fd, "some data", ...) // C

try? close(fd) // D
```

this looks harmless because it looks somewhat sensible that `close` would be able to detect a failure to close. But that is _not true_ on UNIX because POSIX guarantees to always use the lowest available fd number.

So the above piece of code will absolutely close an unrelated file descriptor if some other thread opens a file descriptor between point A and D which is very likely. Now it might sound kinda harmless to accidentally close an unrelated file descriptor but it's really not. This is how security vulnerabilities are made: If we closed an unrelated file descriptor a 'hole' opens up in the file descriptor table that will often promptly be filled with something else. This means that writes of sensitive data might hit unrelated network connections or files. I have personally seen a process that I worked on that has sent sensitive data means for one HTTP connection into another HTTP connection by accident. It was medical data...

weissi:
The only correct way to handle `EBADF` is to `preconditionFailure("hit EBADF -- something closed owned file descriptor \(fd)")`. See SwiftNIO code which handles that correctly.

Issue #2, opened 2025-04-07
Making `Execution` `~Copyable` instead of class


iCharlesHu:
`Execution` semantically should be `~Copyable` because it represents an instance of a process that has already executed. We ran into issues marking it `~Copyable` because we have to capture output and error concurrently, thus through different contexts. We should re-explore making `Execution` `~Copyable`

PR #38, opened 2025-05-07
Adopt ~Copyable in `Subprocess`

iCharlesHu:
This PR introduces `~Copyable` into `Subprocess`, including `CreatedPipe`, `TrackedFileDescriptor`, and `TrackedDispatchIO`.

Resolves https://github.com/swiftlang/swift-subprocess/issues/37
Resolves https://github.com/swiftlang/swift-subprocess/issues/2

Additionally, we opted to slightly modify the API by moving the `.standardOutput` and `.standardError` `AsyncSequence` from `Execution` to the closure parameter, alongside `Execution`. This adjustment eliminated the necessity for `Atomic` (and `AtomicBox`) within `Execution`.

Furthermore, this change rendered `SequenceOutput` internal, as we now rely on overloads to infer the output and error types.

---

This patch also drops support for `Swift 6.0` because it can't properly support the `~Copyable` work. Now `Swift 6.1` is the minimal required Swift version to build `Subprocess`


Sources/Subprocess/API.swift
```diff
@@ -91,16 +91,83 @@ public func run<
...
+    ).run(
+        input: try customInput.createPipe(),
+        output: try output.createPipe(),
+        error: try error.createPipe()
+    ) { execution, inputIO, outputIO, errorIO in
+        // Write input, capture output and error in parallel
+        var inputIOBox: TrackedPlatformDiskIO? = consume inputIO
+        var outputIOBox: TrackedPlatformDiskIO? = consume outputIO
+        var errorIOBox: TrackedPlatformDiskIO? = consume errorIO
+        return try await withThrowingTaskGroup(
```
jakepetroules: Instead of using a task group, how about using `async let`? It's shorter, simpler, and takes better advantage of the type system (no force unwraps or OutputCapturingState helper type needed):

```swift
...
    ) { execution, inputIO, outputIO, errorIO in
        // Write input, capture output and error in parallel
        var inputIOBox: TrackedPlatformDiskIO? = consume inputIO
        var outputIOBox: TrackedPlatformDiskIO? = consume outputIO
        var errorIOBox: TrackedPlatformDiskIO? = consume errorIO

        var outputIOContainer: TrackedPlatformDiskIO? = outputIOBox.take()
        var errorIOContainer: TrackedPlatformDiskIO? = errorIOBox.take()
        async let stdout = output.captureOutput(
            from: outputIOContainer.take()
        )
        async let stderror = error.captureOutput(
            from: errorIOContainer.take()
        )

        // Write span at the same isolation
        if let writeFd = inputIOBox.take() {
            let writer = StandardInputWriter(diskIO: writeFd)
            _ = try await writer.write(input._bytes)
            try await writer.finish()
        }

        return try await (
            processIdentifier: execution.processIdentifier,
            standardOutput: stdout,
            standardError: stderror
        )
    }
...
```

Sources/Subprocess/Configuration.swift
```diff
@@ -815,34 +594,83 @@ internal struct TrackedFileDescriptor {
...
-                throw errno
+            // Getting `.badFileDescriptor` suggests that the file descriptor
+            // might have been closed unexpectedly. This can pose security risks
+            // if another part of the code inadvertently reuses the same file descriptor
+            // number. This problem is especially concerning on Unix systems due to POSIX’s
+            // guarantee of using the lowest available file descriptor number, making reuse
+            // more probable. We use `preconditionFailure` upon receiving `.badFileDescriptor`
+            // to prevent accidentally closing a different file descriptor.
+            guard errno != .badFileDescriptor else {
+                preconditionFailure(
```
jakepetroules: Should this be `fatalError`? The string of `preconditionFailure` is not retained in release builds, which will make the cause of crashes here, harder to identify.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -192,67 +202,121 @@ extension Configuration {
                 defer {
                     posix_spawn_file_actions_destroy(&fileActions)
                 }
+
                 // Input
                 var result: Int32 = -1
-                if let inputRead = inputPipe.readFileDescriptor {
-                    result = posix_spawn_file_actions_adddup2(&fileActions, inputRead.platformDescriptor, 0)
+                if inputReadFileDescriptor != nil {
```
jakepetroules, line 208: Suggest using `if let inputReadFileDescriptor {` so you don't need the force unwrap below.

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -15,45 +15,57 @@
 @preconcurrency import SystemPackage
 #endif
 
+internal import Dispatch
```
parkera, line 19: Should this only import dispatch on `!os(Windows)`?

Sources/Subprocess/API.swift
parkera: The reason for the boxes and force-unwraps here was to work around the lack of a feature to tell the compiler that `withThrowingTaskGroup` may only be called once, and therefore the non copyable types are safe to pass in to the closure. If using `async let` avoids that, it would be better.

iCharlesHu: @parkera unfortunately like @jakepetroules showed above, we still need `inputIOBox` and friends because `async let` itself is considered an escaping closure.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
iCharlesHu: Unfortunately this pattern, along with `if let a = a` is not compatible with `~Copyable` types. Writing `if let inputReadFileDescriptor` would have consumed `inputReadFileDescriptor` so the line below would be double consume.

iCharlesHu:
Addressed review comments and removed `Swift 6.0` support.

parkera (approved):

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -397,67 +392,32 @@ extension FileDescriptor {
...
+extension TrackedFileDescriptor {
+    internal consuming func createPlatformDiskIO() -> TrackedPlatformDiskIO {
+        let dispatchIO: DispatchIO = DispatchIO(
+            type: .stream,
+            fileDescriptor: self.platformDescriptor(),
+            queue: .global(),
+            cleanupHandler: { error in
+                // Close the file descriptor
+                if self.closeWhenDone {
+                    try? self.safelyClose()
```
weissi, line 404: why ignore the error? That can't be quite right

Sources/Subprocess/Configuration.swift
```diff
@@ -815,34 +594,83 @@ internal struct TrackedFileDescriptor {
...
+            // number. This problem is especially concerning on Unix systems due to POSIX’s
+            // guarantee of using the lowest available file descriptor number, making reuse
+            // more probable. We use `fatalError` upon receiving `.badFileDescriptor`
+            // to prevent accidentally closing a different file descriptor.
+            guard errno != .badFileDescriptor else {
+                fatalError(
+                    "FileDescriptor \(fileDescriptor.rawValue) is already closed"
+                )
+            }
+            // Otherwise ignore the error
```
weissi, line 659: this is a bad idea. Closing in `deinit` for file descriptors is bad enough (`close` may block too depending on the fd) but ignoring the errors isn't a thing we can do.

Sources/Subprocess/Configuration.swift
```diff
@@ -815,34 +594,83 @@ internal struct TrackedFileDescriptor {
...
+                )
             }
+            // Throw other kinds of errors to allow user to catch them
+            throw error
         }
     }
 
-    internal var platformDescriptor: PlatformFileDescriptor {
+    deinit {
+        guard self.closeWhenDone else {
```
weissi, line 637: This looks like we're hiding some lifecycle bugs here, no? We should know if we closed a file descriptor or not, right?

Sources/Subprocess/Configuration.swift
FranzBusch: I agree with @weissi here FDs cannot be modeled with `~Copyable` types at this point both due the `async` close nature and the errors on close. My recommendation is to use `with-style` methods which allow to solve both issues.

weissi: Yup, agree with Franz. `with*` solves those issues correctly and also clearly for a reader of the code.

itingliu: @weissi can you open up issues for your comments? This PR is merged and discussions here might be lost.

Issue #22, opened 2025-04-24
Many tests fail on Windows with Swift 6.1


jakepetroules:
**Describe the bug**
The following tests hang and never complete:
- testInputSequence
- testInputAsyncSequence
- testInputSequenceCustomExecutionBody
- testInputAsyncSequenceCustomExecutionBody
- testRedirectedOutputRedirectToSequence

The following test fails with "tester.ps1 cannot be loaded because running scripts is disabled on this system. For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.", but it's nondeterministic and tends not to reproduce when the test is run by itself instead of as part of the whole suite.
- testPlatformOptionsCreateNewConsole

The following tests fail with exit code 1 and produce no output:
- testPlatformOptionsPreSpawnConfigurator
- testSuspendResumeProcess

**To Reproduce**
`swift test`

**Expected behavior**

**Environment (please complete the following information):**
 - Microsoft Windows [Version 10.0.26100.3624]
 - Swift version 6.1 (swift-6.1-RELEASE)
- Target: x86_64-unknown-windows-msvc

**Additional context**
I know this package is mostly Swift 6.2-aligned, and I haven't tested with Swift 6.2 in case that makes any difference.

jakepetroules:
Note that `package func write(_ ptr: UnsafeRawBufferPointer, completion: @escaping (Int, Swift.Error?) -> Void)` never calls its completion handler.

jakepetroules:
`internal func readChunk(upToLength maxLength: Int) async throws -> SequenceOutput.Buffer?` never returns `nil` when it receives the empty buffer indicating EOF, making the AsyncSequence infinite.

jakepetroules:
Even after fixing those two issues, closing the parent file descriptors in "cleanup" appears to cause the process to immediately exit with code 1.

PR #40, opened 2025-05-08
Allow callers to run a subprocess and provide low and high water marks when using SequenceOutput to emit standard output and standard error as soon as it arrives.

### LINKED ISSUE #39: Processes with small and infrequent output doesn't get emitted in a timely manner when using sequence to capture the output
**Describe the bug**
I have a process I want to monitor that is long running. It has small and infrequent output that I need to get in as soon as it is available. I'm running this process with swift-subprocess and monitoring its output  by using `SequenceOutput`. In the custom execution closure, I only get the output either when the process exits or when the output size reaches `readBufferSize` defined in `AsyncBufferSequence`.

**To Reproduce**

Here is some sample code that demonstrates the problem:

```
        let script = """
        echo "[NOTICE] Important log line"
        sleep 3600
        """

        let catResult = try await Subprocess.run(
            .path("/bin/bash"),
            arguments: ["-c", script],
            output: .sequence(lowWater: 0),
            error: .discarded,
            body: { (execution, _) in
                for try await chunk in execution.standardOutput {
                    let string = chunk.withUnsafeBytes { String(decoding: $0, as: UTF8.self) }

                    if string.hasPrefix("[NOTICE]") {
                        // Do Something interesting
                        print(string)
                    }
                }

                return ""
            }
        )
```

**Expected behavior**

There should be a mechanism for getting the output as soon as it is available.

 - macOS 15
 - swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)

**Additional context**

I'll follow up with a PR that has a proposed solution.

rdingman:
Also, fixed a bug where DispatchIO.read could end up resuming a Swift continuation multiple times

Resolves #39


Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -27,38 +29,100 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+        private var streamIterator: AsyncThrowingStream<StreamStatus, Swift.Error>.AsyncIterator
 
-        internal init(diskIO: TrackedPlatformDiskIO) {
+        internal init(diskIO: TrackedPlatformDiskIO, bufferSize: Int) {
             self.diskIO = diskIO
+            self.bufferSize = bufferSize
             self.buffer = []
             self.currentPosition = 0
             self.finished = false
+            self.streamIterator = Self.createDataStream(with: diskIO.dispatchIO, bufferSize: bufferSize).makeAsyncIterator()
```
iCharlesHu: `AsyncBufferSequence` is a shared type across all platforms therefore we can't unconditionally refer to platform specific type `dispatchIO` here. We may see Windows build failure as a result

Sources/Subprocess/IO/Output.swift
```diff
@@ -210,8 +210,15 @@ public struct BytesOutput: OutputProtocol {
...
-    internal init() {}
+    internal let lowWater: Int?
+    internal let highWater: Int?
+    internal let bufferSize: Int
+
+    internal init(lowWater: Int? = nil, highWater: Int? = nil, bufferSize: Int = readBufferSize) {
+        self.lowWater = lowWater
+        self.highWater = highWater
+        self.bufferSize = bufferSize
+    }
```
iCharlesHu: I don’t think it’s appropriate to include these parameters here for a couple of reasons:

- (This isn’t directly related to your change) right now, we’re in the middle of some major architectural updates: https://github.com/swiftlang/swift-subprocess/pull/38. This PR makes `SequenceOutput` internal, so you can’t use `.sequence` or `.sequence(lowWater: …)` anymore.
- More importantly, this looks like a platform-specific feature. Setting this parameter won’t have any impact on Windows, and (also unrelated to your change) we’re planning to move away from `DispatchIO` on Linux soon, so it won’t work there either.

Considering all this, I suggest we move these parameters to Darwin’s specific `PlatformOptions`, maybe under a nested `struct` `PlatformOptions.StreamOptions`.

iCharlesHu (commented):
Thanks so much for the issue and PR! I agree this is a great addition to the API surface.

As a overall comment, could you add a test to make sure the new behavior works as intended?

rdingman:
> Thanks so much for the issue and PR! I agree this is a great addition to the API surface.
> 
> As a overall comment, could you add a test to make sure the new behavior works as intended?

I did add a test named "testSlowDripRedirectedOutputRedirectToSequence". Does that not cover the new behavior like you intend?

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Thanks for the feedback. I'll look into implementing this in such a way that platforms without dispatchIO don't break.

Sources/Subprocess/IO/Output.swift
rdingman: Noted. I'll look into moving these parameters.

rdingman:
@iCharlesHu Any suggestions on how to build and test this on windows? I have things building on windows, but I'm having trouble debugging the tests.

rdingman:
@iCharlesHu I figured out how to get the debugger working in VSCode on Windows.

However, it appears that several of the tests crash with an exception because a file descriptor is being closed more than once (this is the case on `main`). Is this a known issue?

I tried out your PR https://github.com/swiftlang/swift-subprocess/pull/38 to see if it fixed those issues, but it has not.

For what its worth I'm running on:

* VSCode 1.100.0
* Swift Extension 2.2.0
* LLDB DAP 0.2.13
* Windows 11 Pro 24h2
* Swift version 6.1 (swift-6.1-RELEASE)
  Target: aarch64-unknown-windows-msvc

iCharlesHu:
> > Thanks so much for the issue and PR! I agree this is a great addition to the API surface.
> > As a overall comment, could you add a test to make sure the new behavior works as intended?
> 
> I did add a test named "testSlowDripRedirectedOutputRedirectToSequence". Does that not cover the new behavior like you intend?

Ahh yes! Sorry I totally missed `testSlowDripRedirectedOutputRedirectToSequence`. That should work thanks!

iCharlesHu:
> @iCharlesHu I figured out how to get the debugger working in VSCode on Windows.
> 
> However, it appears that several of the tests crash with an exception because a file descriptor is being closed more than once (this is the case on `main`). Is this a known issue?
> 
> I tried out your PR #38 to see if it fixed those issues, but it has not.
> 
> For what its worth I'm running on:
> 
> * VSCode 1.100.0
> * Swift Extension 2.2.0
> * LLDB DAP 0.2.13
> * Windows 11 Pro 24h2
> * Swift version 6.1 (swift-6.1-RELEASE)
>   Target: aarch64-unknown-windows-msvc



> @iCharlesHu I figured out how to get the debugger working in VSCode on Windows.
> 
> However, it appears that several of the tests crash with an exception because a file descriptor is being closed more than once (this is the case on `main`). Is this a known issue?
> 
> I tried out your PR #38 to see if it fixed those issues, but it has not.
> 
> For what its worth I'm running on:
> 
> * VSCode 1.100.0
> * Swift Extension 2.2.0
> * LLDB DAP 0.2.13
> * Windows 11 Pro 24h2
> * Swift version 6.1 (swift-6.1-RELEASE)
>   Target: aarch64-unknown-windows-msvc

Thanks so much for looking into the Windows build. Unfortunately we do have some known test failures on Windows currently (https://github.com/swiftlang/swift-subprocess/issues/22) and I'll address them separately. Right now we want to make sure all new changes at least build on Windows.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -451,6 +461,48 @@ extension CreatedPipe {
 
 // MARK: - TrackedDispatchIO extensions
 extension TrackedDispatchIO {
+    internal func readDataStream(upToLength maxLength: Int) -> AsyncThrowingStream<StreamStatus, Swift.Error> {
```
iCharlesHu: You'll need to add
```
#if SubprocessSpan
@available(SubprocessSpan, *)
#endif
```
here because `SequenceOutput.Buffer` is only available with `SubprocessSpan`. Otherwise this package will fail to compile on Swift 6.2.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -1043,6 +1043,70 @@ extension CreatedPipe {
 }
 
 extension TrackedFileDescriptor {
+    internal func readDataStream(upToLength maxLength: Int) -> AsyncThrowingStream<StreamStatus, Swift.Error> {
```
iCharlesHu: Ditto with
```
#if SubprocessSpan
@available(SubprocessSpan, *)
#endif
```

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -62,6 +74,14 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
     }
 }
 
+extension TrackedPlatformDiskIO {
+    internal enum StreamStatus {
```
iCharlesHu: Ditto
```
#if SubprocessSpan
@available(SubprocessSpan, *)
#endif
```

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -451,6 +461,48 @@ extension CreatedPipe {
...
+
+                if done {
+                    continuation.finish()
+                }
+            }
+        }
+    }
+
 #if SubprocessSpan
     @available(SubprocessSpan, *)
```
iCharlesHu: I believe `readChunk` below is no longer used. Could you remove it please?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -30,24 +32,34 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
-                // We finished reading. Close the file descriptor now
+        public mutating func next() async throws -> SequenceOutput.Buffer? {
+            if let status = try await streamIterator.next() {
+                switch status {
+                case .data(let data):
+                    return data
+
+                case .endOfStream(let data):
+                    streamIterator = diskIO.readDataStream(upToLength: readBufferSize).makeAsyncIterator()
+                    return data
```
iCharlesHu: This does not seem right. Why are we creating a new iterator when the first one ends? There will be nothing to read from this second iterator because all the data in the pipe would already been read.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -451,6 +461,48 @@ extension CreatedPipe {
...
+
+                switch (buffer, done) {
+                case (.some(let data), false):
+                    status = .data(SequenceOutput.Buffer(data: data))
+
+                case (.some(let data), true):
+                    status = .endOfStream(SequenceOutput.Buffer(data: data))
+
+                case (nil, false):
+                    return
```
iCharlesHu: This case shouldn't happen according to the document. IMO we should `fatalError()` here instead of just `return` in case it does happen and we will end up not resuming.

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -665,6 +665,48 @@ extension SubprocessUnixTests {
         #expect(catResult.terminationStatus.isSuccess)
         #expect(catResult.standardError == expected)
     }
+
+    @Test func testSlowDripRedirectedOutputRedirectToSequence() async throws {
+        let threshold: Double = 0.5
```
iCharlesHu, line 670: Unfortunately in tests you'll have to write
```
guard #available(SubprocessSpan , *) else {
    return
}
```
In the beginning to work around the same availability issue. See other tests for examples.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -197,6 +199,18 @@ public struct PlatformOptions: Sendable {
...
 
+extension PlatformOptions {
+    public struct StreamOptions: Sendable {
+        let lowWater: Int?
+        let highWater: Int?
+
+        init(lowWater: Int? = nil, highWater: Int? = nil) {
+            self.lowWater = lowWater
+            self.highWater = highWater
+        }
```
iCharlesHu: Sorry I know I initially suggested using a `StreamOptions` nested struct, but after revisiting this API, I think we should reconsider the `lowWater` and `highWater` properties. Here’s why: 1) Their names can be quite confusing outside of the DispatchIO context, and 2) we’d need to add runtime validation to ensure `lowWater < highWater`.

How about we try something like this instead?
```swift
struct PlatformOptions {
    …
    let preferredStreamBufferSizeRange: Range<Int>? = nil
}
```
This approach makes it clear that we’re requesting a range (with a lower and upper bound), and it eliminates the need for validation.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -176,6 +176,8 @@ public struct PlatformOptions: Sendable {
     // Creates a session and sets the process group ID
     // i.e. Detach from the terminal.
     public var createSession: Bool = false
+    public var outputOptions: StreamOptions = .init()
+    public var errorOptions: StreamOptions = .init()
```
iCharlesHu: IMO we only need one (see my comments below).

Sources/Subprocess/Platforms/Subprocess+Unix.swift
rdingman: Fixed.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
rdingman: Fixed.

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Fixed.

Tests/SubprocessTests/SubprocessTests+Unix.swift
rdingman: @iCharlesHu When I first started working on this, I was very confused as to why some of the tests weren't running my new code and it was because of this check. Wouldn't it be better to have them skipped and noted as such in the test output rather than falsely succeeding? I'm thinking something like this:

```
    @Test(
        .enabled(
            if: {
                if #available(SubprocessSpan , *) {
                    true
                } else {
                    false
                }
            }(),
            "This test requires SubprocessSpan"
        )
    )
    func testSlowDripRedirectedOutputRedirectToSequence() async throws {
    }
```

Of course, we can have a helper function to make this less verbose.

Thoughts?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
rdingman: @iCharlesHu The issue I see with this suggestion is that it presumes that you will set both the lower and upper bound, or neither. There is no mechanism for setting just one or the other. As I see it, we have a few options:

1. Adopt your suggestion and take the stance that you cannot set these independently.
2. Recognize that these are platform specific options and rename them to indicate that. On Linux, these options would be removed once the Linux implementation moves away from DispatchIO because they are DispatchIO specific. We'd add the appropriate runtime check here. (FWIW, DispatchIO handles when lowWater > highWater and makes them the same).
3. Attempt to use some sort of sentinel values to represent the "don't set this" or "use the default" case.  For `lowWater` mark, this could be something like -1 (the default is "unspecified"). For `highWater` this is tougher because the documentation says the default value is `SIZE_MAX` which isn't representable by `Int`. IMO, this option doesn't seem very intuitive, but I thought I'd include it anyways.
4. Change to use an enum which represents the four cases of set neither, set lower, set upper, set both. Something like:
```
enum BufferSizeOptions {
case none
case lowerBound(Int)
case upperBound(Int)
case range(Range<Int>)
}
```

Thoughts?

I think option 3 is too awkward and unintuitive, so I don't think we should consider it. If you feel strongly about option 1, we can go with that, but I wanted to bring up this issue.

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Thanks for the pushback on this. I was going to explain my thinking and then realize if I have to explain this then it probably should be written in a more straightforward manner. The first implementation was using one iterator per chunk and when we reached a chunk boundary we'd switch to a new iterator. While this did work, it was admittedly a little clunky. Now, we use one AsyncThrowingStream (and iterator) across all chunks (even if the chunk is broken up into sub-chunks fora single read as in my original motivation for this issue. Please check on the new implementation to see if it makes sense to you.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
rdingman: Removed.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
rdingman: Updated.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
rdingman: @iCharlesHu  I modified your proposal a bit to use a `RangeExpression` rather than just a concrete `Range` that way we can express things like `0...` to only set the lower bounds or `...4096` to only set the upper bounds. I didn't want to make `PlatformOptions` fully generic because this would be more cumbersome. Instead, I added some API on `PlatformOptions` to enforce the requirement that `RangeExpression.Bound` must be an `Int`.

Check it out and let me know what you think.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
rdingman: I replaced these with `preferredStreamBufferSizeRange` above.

Tests/SubprocessTests/SubprocessTests+Unix.swift
rdingman: @iCharlesHu I went ahead and conditionalized this one test this way as an example. Let me know if you don't like that and would like me to revert to a `guard`

iCharlesHu: @rdingman unfortunately this won't work because on Swift 6.2 and above the compiler will complain the code inside of `testSlowDripRedirectedOutputRedirectToSequence` needs `SubprocessSpan` and ask you to put `@available` around the function. That doesn't work either because the macro won't pick it up.

Unfortunately so far this is the only way I found that works. Very ugly... but it'll have to do for now.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: > The issue I see with this suggestion is that it presumes that you will set both the lower and upper bound, or neither. There is no mechanism for setting just one or the other.

Ahh I see your point. In this case let's go back to
```
struct StreamOptions {
    public let minimalBufferSize: Int? = nil
    public let maximumBufferSize: Int? = nil
}
```
I think that's the simplest.

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Fixed.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
rdingman: @iCharlesHu Ok, great. I've gone back to configuring this through `StreamOptions`

Tests/SubprocessTests/SubprocessTests+Unix.swift
rdingman: @iCharlesHu Hmm. This builds just fine for me with the latest Swift 6.2 toolchain, has the intended results, and works with the currently shipping Swift 6.1 toolchain. However, I'll go ahead and revert this and match the existing tests.

rdingman:
@iCharlesHu I merged in the latest changes from `main`. This removed the place where I was consuming StreamOptions in order to set the low and high watermark on DispatchIO. The latest implementation passes it into `AsyncBufferSequence.init` and does the work there because that caused the least internal API changes. However, it seems like it might be more appropriate to pass it to `TrackedDispatchIO.init` and do the work there, but the code that creates `TrackedDispatchIO` doesn't have access to `PlatformOptions` and I didn't want to go changing a bunch of your internal API without asking.

What are your thoughts? Do you prefer one or the other? Do you have a third option you'd like to pursue?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -35,44 +35,93 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
     public func makeAsyncIterator() -> Iterator {
-        return Iterator(diskIO: self.diskIO)
+        return Iterator(diskIO: self.diskIO, streamOptions: streamOptions)
+    }
+}
+
+extension AsyncBufferSequence {
+#if SubprocessSpan
+@available(SubprocessSpan, *)
+#endif
```
iCharlesHu: You'll have to move these three lines above `extension AsyncBufferSequence` because `AsyncBufferSequence` itself is only available with `SubprocessSpan`.

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -35,44 +35,93 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+    internal enum StreamStatus {
+        case data(AsyncBufferSequence.Buffer)
+        case endOfChunk(AsyncBufferSequence.Buffer)
+        case endOfFile
```
iCharlesHu: I don't think we need to use this `enum` to model the stream states (see my other comments below)

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -414,37 +415,42 @@ extension DispatchIO {
...
+
+            case (nil, true):
+                status = .endOfFile
+            }
+
+            continuation.yield(status)
+
+            if done {
+                continuation.finish()
             }
```
iCharlesHu: Instead of setting using a `StreamStatus` to model the state, I think just having `Iterator.Stream` be `AsyncThrowingStream<Buffer>` is good enough. Because when the throwing stream returns `nil`, it naturally implies "end of stream"

```swift
switch (buffer, done) {
case (.some(let data), false):
    continuation.yield(AsyncBufferSequence.Buffer(data: data))
case (.some(let data), true):
    continuation.yield(AsyncBufferSequence.Buffer(data: data))
    continuation.finish() // Finish the stream now
case (nil, false):
    fatalError("Unexpectedly received no data from DispatchIO with it indicating it is not done.")
case (nil, true):
    continuation.finish() // Finish the stream
}
```

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -35,44 +35,93 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+                    return data
+
+                case .endOfFile:
+                    #if os(Windows)
+                    try self.diskIO.close()
+                    #else
+                    self.diskIO.close()
+                    #endif
+                    return nil
+                }
```
iCharlesHu: If `streamIterator` returned `AsyncThrowingStream<Buffer>` here you can just return it directly.

iCharlesHu (commented):
Overall looks pretty good! Thanks so much for working on this! This should be my last bit of comment before we are good to merge! 

Just to write down some of the code I was suggesting:
1) Rename `.readChunk` to `.stream` (like you suggested earlier), and have it return a `AsyncThrowingStream<Buffer>` directly (no need for `StreamStatus`)
2) Have `AsyncBufferSequence.Iterator` call into `.stream` directly.

```
extension DispatchIO {
    #if SubprocessSpan
    @available(SubprocessSpan, *)
    #endif
    internal func stream(
        upToLength maxLength: Int
    ) -> AsyncThrowingStream<AsyncBufferSequence.Buffer, any Error> {
        return AsyncThrowingStream { continuation in
            self.read(
                offset: 0,
                length: maxLength,
                queue: .global()
            ) { done, data, error in
                if error != 0 {
                    continuation.finish(
                        throwing: SubprocessError(
                            code: .init(.failedToReadFromSubprocess),
                            underlyingError: .init(rawValue: error)
                        )
                    )
                    return
                }

                switch (done, data) {
                case (true, .none):
                    continuation.finish()
                case (true, .some(let dispatchData)):
                    continuation.yield(.init(data: dispatchData))
                    continuation.finish()
                case (false, .some(let dispatchData)):
                    continuation.yield(.init(data: dispatchData))
                case (false, .none):
                    fatalError()
                }
            }
        }
    }
}

// Inside `AsyncBufferSequence.Iterator`
public struct Iterator: AsyncIteratorProtocol {
    public typealias Element = Buffer

    private let diskIO: DiskIO

    private var base: AsyncThrowingStream<Buffer, any Error>.AsyncIterator

    internal init(diskIO: DiskIO) {
        self.diskIO = diskIO
        self.base = diskIO.stream(upToLength: .max).makeAsyncIterator()
    }

    public mutating func next() async throws -> Buffer? {
        let data = try await self.base.next()
        if data == nil {
            // We finished reading. Close the file descriptor now
            #if os(Windows)
            try self.diskIO.close()
            #else
            self.diskIO.close()
            #endif
            return nil
        }
        return data
    }
}
```

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Removed StreamStatus

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: Removed StreamStatus

Sources/Subprocess/Platforms/Subprocess+Unix.swift
rdingman: @iCharlesHu Simplified this.

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: @iCharlesHu Yes, simplified this.

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -35,44 +35,68 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+        private let continuation: Stream.Continuation
+        private var bytesRemaining: Int
 
-        internal init(diskIO: DiskIO) {
+        internal init(diskIO: DiskIO, streamOptions: PlatformOptions.StreamOptions) {
             self.diskIO = diskIO
             self.buffer = []
             self.currentPosition = 0
             self.finished = false
+            let (stream, continuation) = AsyncThrowingStream<Buffer, Swift.Error>.makeStream()
```
FranzBusch, line 53: I'm a bit concerned with the usage of the `AsyncStream` here for two reasons. The first is that  `AsyncStream` has no way to propagate the internal async sequence back pressure to the external system that is producing the elements. 

We had the same issue in `swift-nio` and created the [NIOAsyncSequenceProducer](https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift) that allowed us to bridge NIO's back pressure into the back pressure of an async sequence. The learned a lot from the APIs and implementation in NIO and have an open PR in `swift-async-algorithms` that generalizes this concept as a [MultiProducerSingleConsumerAsyncChannel](https://github.com/apple/swift-async-algorithms/pull/305). 

Now having said that it might be fine here back pressure wise since the `DispatchIO` is going to call the `ioHandler` multiple times but only up to `readBufferSize`. So we do have some maximum limit and our buffer can't balloon indefinitely.

However, this brings me to my second point which is the performance of this. `AsyncStream` is not super fast so I expect this PR to have performance impact when streaming a lot of data to/from a subprocess. It would be good to understand that impact and if the `MultiProducerSingleConsumerAsyncChannel` can improve this.

Lastly, aren't we missing to close the `diskIO` when the iterator is dropped. We probably want to setup a `onTerminationCallback` on the stream or the channel to close the diskIO right?

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -414,37 +415,25 @@ extension DispatchIO {
...
+                continuation.finish(throwing: SubprocessError(
+                    code: .init(.failedToReadFromSubprocess),
+                    underlyingError: .init(rawValue: error)
+                ))
+                return
+            }
+
+            // Treat empty data and nil as the same
+            if let data = data.map({ $0.isEmpty ? nil : $0 }) ?? nil {
+                continuation.yield(AsyncBufferSequence.Buffer(data: data))
```
weissi, line 446: Wait, this strips backpressure. If you do that then a slow consumer and fast producer will OOM kill you. You won't be able to use AsyncStream here (it's a type that should be avoided unless you use `AsyncStream(unfolding: {...})` or set it to drop elements if the buffer is full (but of course that's not possible here because it would lose data).

iCharlesHu: I understand the concern, but I don't think this is the right PR to address this issue since it's trying to solve a different problem. I'll address this in a separate PR.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -414,37 +415,25 @@ extension DispatchIO {
...
+            } else if done {
+                continuation.finish()
```
iCharlesHu: I think we should stick with `switch` just to be explicit that we handle all possible combinations. For example, it's possible `done == true && data != nil` the current implementation won't call `finish()`.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -441,6 +442,12 @@ public struct PlatformOptions: Sendable {
     public init() {}
 }
 
+extension PlatformOptions {
+    internal struct StreamOptions: Sendable {
+        internal init() {}
+    }
+}
+
```
iCharlesHu: I understand this might make code sharing realizer but I don't think it's appropriate to add an empty struct here that will never be used (same with the `streamOption` property on Windows). 

(Sorry I might have missed this in earlier reviews)

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -35,44 +35,68 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
-            let data = try await self.diskIO.readChunk(
-                upToLength: readBufferSize
-            )
-            if data == nil {
-                // We finished reading. Close the file descriptor now
+        public mutating func next() async throws -> Buffer? {
+
+            if bytesRemaining <= 0 {
+                bytesRemaining = readBufferSize
+                diskIO.stream(upToLength: readBufferSize, continuation: continuation)
```
iCharlesHu, line 62: I don't think you need to manually stream only up to `readBufferSize` here. The idea of streaming is that we want to read until the end of file so we should use `.max` here to specify `DispatchIO` should read data until an EOF is reached. You should only need to call `stream` once in the initializer.

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: I'll look into this in a separate PR. I think the `AsyncStream` based implementation is good enough for this PR.

Sources/Subprocess/AsyncBufferSequence.swift
rdingman: @iCharlesHu 

Are we sure we want that? That exacerbates the concerns of @weissi  and @FranzBusch. With the current approach, the easing of back pressure is limited to buffering up to about `readBufferSize` for slow consumers. This suggestion opens that up to the entire output leaving it essentially unbounded.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
iCharlesHu: Created https://github.com/swiftlang/swift-subprocess/issues/51

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: Will investigate under https://github.com/swiftlang/swift-subprocess/issues/51

Sources/Subprocess/Platforms/Subprocess+Windows.swift
rdingman: Rearranged how we use StreamOptions so it is only defined and used on platforms that need it.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
rdingman: I didn't go back to a switch because I re-read the semantics of `DispatchIO.read()` and restructured this code to handle these semantics. Now all code paths either call `fatalError` for an invalid combination of handler arguments, yield non-empty data, or call `finish()` when we've reach the end of the file.

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: I was trying to unblock this PR so I created https://github.com/swiftlang/swift-subprocess/issues/51 to address the `AsyncStream` issue.

However, after some time to think over and studying @FranzBusch’s suggestion (https://github.com/apple/swift-async-algorithms/pull/305.), I’ve come to the conclusion that we can’t make the `read` function “push based” (using `AsyncStream`) if we want to depend on `DispatchIO`. The main reason is that `DispatchIO` doesn’t support back pressure, so any back pressure mechanism we implement would be limited by `DispatchIO` since we can’t instruct it to slow down (blocking the handler block is a no-go).

The current method (AsyncStream + buffer size limit) is better because it prevents the unlimited buffer problem, but it loses the advantage of letting `DispatchIO` determine the optimal buffer size. Instead of calling `.read` once with `.max`, we call it multiple times with a small buffer size. Additionally, as @FranzBusch pointed out, `AsyncStream` has performance issues, making it not worthwhile to use any more.

Therefore, for this PR, could you please:
- Remove the use of `AsyncStream` and `.stream()` and instead use `withCheckedThrowingContinuation` and `.read()` with a simple `async` function to read data up to a max length (similar to my implementation here https://github.com/swiftlang/swift-subprocess/pull/48/files)
- This change means we can’t rely on calling `.setLimit(lowWater:)` / `.setLimit(highWater:)` since `.read()` will always read until the buffer size is reached. Therefore, instead of `PlatformOptions.StreamOptions`, we just need a `let preferredStreamBufferSize: Int` to set the actual buffer size we instruct `.read()` to fetch each time.

I believe the above approach should still allow callers to set how frequent they wish to be updated without needing to introduce `AsyncStream`.

PR #43, opened 2025-05-11
Make the package build with Swift 6.0.3.

ethan-kusters:
There are a couple instances of trailing commas in the project that prevents it from compiling with Swift 6.0.3. This resolves those issues.


iCharlesHu (approved):
Thanks for the fix!

Issue #44, opened 2025-05-14
`run` appears to be stuck (due to buffer limit?)


itingliu:
```swift
let result = try await Subprocess.run(
    .path("/bin/cat"),
    arguments: ["path/to/some/large/text/file.txt]
)
```

This looks like it is stuck and never returns. It is because the file content exceeds the default limit, and `cat` is waiting to write more to the output buffer.

iCharlesHu:
For this one we decided to just remove the default buffer size limit, but still allow you to set one if you choose to.

weissi:
That doesn't sound like the right fix. If something blows past the 'maximum bytes collect' limit, you should throw an error (potentially giving the user what we have so far).

Just raising the buffer won't help, try `cat /dev/zero`

iCharlesHu:
> 1. Should the default value for stdin/stderr be discard or collect?

The current default values for `stdin`, `stdout`, and `stderr` are `.none` (no input), `.string` (collect as String), and `.discarded`. I believe this is the most common use case. Instead of setting the output to `.string` by default, we propose removing this default setting and `.string` itself. This means the caller will have to explicitly choose an output type and buffer size. We chose this approach because it aligns with Swift’s overall design principle: if we can’t provide a reasonable default (and in this case, we truly can’t), we leave it to the caller to decide. This way, you can pass `.string(limit: .max)` if you want the `git status without limit` behavior.


> 2. If “collect” is the default: should the buffer be unbounded or bounded+throw?

This is a good question which is why we decided to not offer a default at all. You'll have to pick one according to your needs.

> 3. When we reach the limit do we throw or return normally?

I think what you’re describing is a choice between a "soft" and a "hard" limit. We opted for a hard limit here because, for many **collected** output types, it’s not possible to arbitrarily limit the output and expect it to still be convertible to the target output type. For instance, converting a truncated buffer to a `String` might not always work due to incorrect grapheme cluster breaks. In my opinion, if you want a soft limit, you should use the custom closure `run`. This way, you can decide how to process potentially partial data and determine what to do (send a signal, etc.) when the limit is reached.


> 4. When do we throw the limit error?

We decided on throwing the error at the exact moment the limit is reached since there's no point return partial data as discussed above. At which point we will attempt to gracefully terminate the child process by running `teardownSequence`.

> Things may go very wrong if we kill the process in the middle of a distributed database update.

[...]

PR #45, opened 2025-05-15
Remove the default output buffer limit

iCharlesHu:
This PR removes the output buffer limit by default when setting `.string` or `.bytes` as `output` or `error`.

Resolves https://github.com/swiftlang/swift-subprocess/issues/44

### DIFF
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -46,7 +46,7 @@ public protocol OutputProtocol: Sendable, ~Copyable {
 #endif
 extension OutputProtocol {
     /// The max amount of data to collect for this output.
-    public var maxSize: Int { 128 * 1024 }
+    public var maxSize: Int { .max }
 }
 
 /// A concrete `Output` type for subprocesses that indicates that
@@ -240,10 +240,9 @@ extension OutputProtocol where Self == FileDescriptorOutput {
 @available(SubprocessSpan, *)
 #endif
 extension OutputProtocol where Self == StringOutput<UTF8> {
-    /// Create a `Subprocess` output that collects output as
-    /// UTF8 String with 128kb limit.
+    /// Create a `Subprocess` output that collects output as UTF8 String
     public static var string: Self {
-        .init(limit: 128 * 1024, encoding: UTF8.self)
+        .init(limit: .max, encoding: UTF8.self)
     }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/Subprocess/IO/Output.swift
```diff
@@ -240,10 +240,9 @@ extension OutputProtocol where Self == FileDescriptorOutput {
 @available(SubprocessSpan, *)
 #endif
 extension OutputProtocol where Self == StringOutput<UTF8> {
-    /// Create a `Subprocess` output that collects output as
-    /// UTF8 String with 128kb limit.
+    /// Create a `Subprocess` output that collects output as UTF8 String
```
parkera, line 243: I think the documentation comment here should say that the buffer is unlimited, so the amount of memory required is proportional to the size of the output.

iCharlesHu: Sounds good

iCharlesHu:
Updated header doc to indicate that `.string` and `.bytes` have unlimited buffer size.

Sources/Subprocess/IO/Output.swift
```diff
@@ -46,7 +46,7 @@ public protocol OutputProtocol: Sendable, ~Copyable {
 #endif
 extension OutputProtocol {
     /// The max amount of data to collect for this output.
-    public var maxSize: Int { 128 * 1024 }
+    public var maxSize: Int { .max }
```
weissi, line 49: @iCharlesHu Why are we fixing a hang with increasing the buffer size to infinity? That's not quite the right fix I think.

if you're not streaming and you blow through the max size you should get an error, not a hang. That's also what AsyncProcess does.

in fact, you will now see a hang (followed by an OOM kill) if you run for example `cat /dev/zero`.

parkera: I understand your point, but struggling a bit to find the right answer here. If we put a limit in, and during testing someone only uses file sizes < 128k (for example), then a customer of the app is the one who hits the error. If we are unlimited, then it'll just use infinite memory, which is also a problem.

weissi: Safe and unexpected is better than unsafe an unexpected IMHO. So a limit is IMHO a must.

weissi: If you get an error telling you "too much data to collect", it's immediately actionable. If your process just grinds to a halt or gets OOM killed, you need to do much more work to even figure out what the problem is.

Issue #15, opened 2025-04-21
Run with Custom Closure sample code not compiling


McNight:
**Describe the bug**
The sample code provided in **Run with Custom Closure** section (requiring `SubprocessSpan` trait) is not compiling.

**To Reproduce**
1. Make sure you have a Swift 6.2 development snapshot installed & configured with `swiftly`
2. `swift package init --type executable`
3. Make `swift-subprocess` a dependency & provide the `SubprocessSpan` trait
4. Use the sample code provided in the **Run with Custom Closure**

**Expected behavior**
It compiles and runs fine.

**Environment (please complete the following information):**
 - OS: macOS 15.4.1
 - Swift version: Apple Swift version 6.2-dev (LLVM 81ab6d9f7e4810f, Swift 9cc1947527bacea)

**Additional context**
I had trouble running the sample code provided in **Run with Custom Closure** section.

First, pls correct me if I'm wrong: I think we should specify that this sample would work only if you specify the `SubprocessSpan` trait.

Then, even with the trait enabled, you have to provide these arguments to the `run` function:

```swift
try await run(
    .path("/bin/dd"),
    arguments: ["if=/path/to/document"],
    input: .none, // <-- here...
    output: .sequence, // <-- here...
    error: .discarded // <-- and here.
)
```

I don't know if it could be interesting to provide default values to these parameters for these **Run with Closure** family of APIs, or if this is intended.

Then, the body of the closure in itself doesn't build, especially this line:

```swift
let string = chunk.withUnsafeBytes { String(decoding: $0, as: UTF8.self) }
```

`withUnsafeBytes` is not available when `SubprocessSpan` is set.

If you change it to:

```swift
let string = chunk.bytes.withUnsafeBytes { String(decoding: $0, as: UTF8.self) } // bytes is the newly introduced RawSpan
```

then you get the following error: **Lifetime-dependent value escapes its scope**.

What's funny though is that I first tried to run this on Asahi Linux Fedora 42 and it worked fine (I didn't get the lifetime value error).

To circumvent this lifetime error on macOS (it wasn't needed on Linux), I added this extension:
```swift
public func stringFromBytes() -> String {
    bytes.withUnsafeBytes { String(decoding: $0, as: UTF8.self) }
}
````
but I don't know if it's correct, I'm not yet familiar with the new `~Escapable` capabilities and `RawSpac` introduced in Swift 6.2...

Finally, with this extension, it built fine but when I tried to run it, I got this runtime issue:
[...]

PR #48, opened 2025-05-19
AsyncBufferSequence.Buffer Improvements

iCharlesHu:
This PR enhances the usability of `AsyncBufferSequence.Buffer` and the closure-based `run()` in several ways:

- Introduced `AsyncBufferSequence.LineSequence`, which lets developers stream standard output and error line by line as `String`. This is the recommended approach for converting `AsyncBufferSequence.Buffer` to `String` because converting each buffer individually might not always work due to potential breaks in the String.
- Updated the implementation of `AsyncBufferSequence.Buffer.bytes` by basing it on `DispatchData.Slice` instead of `DispatchData`. This change ensures that `AsyncBufferSequence.Buffer` holds a contiguous chunk of bytes, which is necessary for the `.bytes` API.
- Removed the unnecessary availability tags for the closure-based `run()`.
- Made several improvements to the documentation.

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -158,9 +158,6 @@ public func run<
 ///   - body: The custom execution body to manually control the running process
 /// - Returns a ExecutableResult type containing the return value
 ///     of the closure.
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result, Input: InputProtocol, Error: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -208,9 +205,6 @@ public func run<Result, Input: InputProtocol, Error: OutputProtocol>(
     }
 }
 
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -258,9 +252,6 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     }
 }
 
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result, Error: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -291,9 +282,6 @@ public func run<Result, Error: OutputProtocol>(
     }
 }
 
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result, Output: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -340,9 +328,6 @@ public func run<Result, Output: OutputProtocol>(
 ///   - body: The custom execution body to manually control the running process
 /// - Returns a ExecutableResult type containing the return value
 ///     of the closure.
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -486,9 +471,6 @@ public func run<
 ///       the running process and write to its standard input.
 /// - Returns a ExecutableResult type containing the return value
 ///     of the closure.
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func run<Result>(
     _ configuration: Configuration,
     isolation: isolated (any Actor)? = #isolation,
@@ -528,9 +510,6 @@ public func run<Result>(
 ///   - output: A file descriptor to bind to the subprocess' standard output.
 ///   - error: A file descriptor to bind to the subprocess' standard error.
 /// - Returns: the process identifier for the subprocess.
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func runDetached(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -564,9 +543,6 @@ public func runDetached(
 ///   - output: A file descriptor to bind to the subprocess' standard output.
 ///   - error: A file descriptor to bind to the subprocess' standard error.
 /// - Returns: the process identifier for the subprocess.
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
 public func runDetached(
     _ configuration: Configuration,
     input: FileDescriptor? = nil,
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -168,7 +168,7 @@ public func run<Result, Input: InputProtocol, Error: OutputProtocol>(
     workingDirectory: FilePath? = nil,
     platformOptions: PlatformOptions = PlatformOptions(),
     input: Input = .none,
-    error: Error,
+    error: Error = .discarded,
     isolation: isolated (any Actor)? = #isolation,
     body: ((Execution, AsyncBufferSequence) async throws -> Result)
 ) async throws -> ExecutionResult<Result> where Error.OutputType == Void {
@@ -267,7 +267,7 @@ public func run<Result, Error: OutputProtocol>(
     environment: Environment = .inherit,
     workingDirectory: FilePath? = nil,
     platformOptions: PlatformOptions = PlatformOptions(),
-    error: Error,
+    error: Error = .discarded,
     isolation: isolated (any Actor)? = #isolation,
     body: ((Execution, StandardInputWriter, AsyncBufferSequence) async throws -> Result)
 ) async throws -> ExecutionResult<Result> where Error.OutputType == Void {
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -152,7 +152,6 @@ public func run<
 ///   - platformOptions: The platform specific options to use
 ///     when running the executable.
 ///   - input: The input to send to the executable.
-///   - output: How to manage the executable standard ouput.
 ///   - error: How to manager executable standard error.
 ///   - isolation: the isolation context to run the body closure.
 ///   - body: The custom execution body to manually control the running process
@@ -205,6 +204,21 @@ public func run<Result, Input: InputProtocol, Error: OutputProtocol>(
     }
 }
 
+/// Run a executable with given parameters and a custom closure
+/// to manage the running subprocess' lifetime and stream its standard error.
+/// - Parameters:
+///   - executable: The executable to run.
+///   - arguments: The arguments to pass to the executable.
+///   - environment: The environment in which to run the executable.
+///   - workingDirectory: The working directory in which to run the executable.
+///   - platformOptions: The platform specific options to use
+///     when running the executable.
+///   - input: The input to send to the executable.
+///   - output: How to manager executable standard output.
+///   - isolation: the isolation context to run the body closure.
+///   - body: The custom execution body to manually control the running process
+/// - Returns a ExecutableResult type containing the return value
+///     of the closure.
 public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -252,6 +266,21 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     }
 }
 
+/// Run a executable with given parameters and a custom closure
+/// to manage the running subprocess' lifetime, write to its
+/// standard input, and stream its standard output.
+/// - Parameters:
+///   - executable: The executable to run.
+///   - arguments: The arguments to pass to the executable.
+///   - environment: The environment in which to run the executable.
+///   - workingDirectory: The working directory in which to run the executable.
+///   - platformOptions: The platform specific options to use
+///     when running the executable.
+///   - error: How to manager executable standard error.
+///   - isolation: the isolation context to run the body closure.
+///   - body: The custom execution body to manually control the running process
+/// - Returns a ExecutableResult type containing the return value
+///     of the closure.
 public func run<Result, Error: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -282,6 +311,21 @@ public func run<Result, Error: OutputProtocol>(
     }
 }
 
+/// Run a executable with given parameters and a custom closure
+/// to manage the running subprocess' lifetime, write to its
+/// standard input, and stream its standard error.
+/// - Parameters:
+///   - executable: The executable to run.
+///   - arguments: The arguments to pass to the executable.
+///   - environment: The environment in which to run the executable.
+///   - workingDirectory: The working directory in which to run the executable.
+///   - platformOptions: The platform specific options to use
+///     when running the executable.
+///   - output: How to manager executable standard output.
+///   - isolation: the isolation context to run the body closure.
+///   - body: The custom execution body to manually control the running process
+/// - Returns a ExecutableResult type containing the return value
+///     of the closure.
 public func run<Result, Output: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -313,8 +357,8 @@ public func run<Result, Output: OutputProtocol>(
 }
 
 /// Run a executable with given parameters and a custom closure
-/// to manage the running subprocess' lifetime and write to its
-/// standard input via `StandardInputWriter`
+/// to manage the running subprocess' lifetime, write to its
+/// standard input, and stream its standard output and standard error.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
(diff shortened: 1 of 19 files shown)


Sources/Subprocess/API.swift
```diff
@@ -267,7 +267,7 @@ public func run<Result, Error: OutputProtocol>(
     environment: Environment = .inherit,
     workingDirectory: FilePath? = nil,
     platformOptions: PlatformOptions = PlatformOptions(),
-    error: Error,
+    error: Error = .discarded,
```
parkera, line 290: Is this part of the same change?

Sources/Subprocess/Buffer.swift
```diff
@@ -88,44 +80,46 @@ extension AsyncBufferSequence.Buffer {
...
+                let empty = UnsafeRawBufferPointer(start: nil, count: 0)
+                let span = RawSpan(_unsafeBytes: empty)
+                return _overrideLifetime(of: span, to: self)
+            }
+            switch backing {
+            case .pointer(let ptr):
+                let span = RawSpan(_unsafeElements: ptr)
+                return _overrideLifetime(of: span, to: self)
+            case .array(let array):
+                let span = array.bytes
```
glessard:
```suggestion
                let span = array.span.bytes
```

Sources/Subprocess/Buffer.swift
```diff
@@ -88,44 +80,46 @@ extension AsyncBufferSequence.Buffer {
...
+                let empty = UnsafeRawBufferPointer(start: nil, count: 0)
+                let span = RawSpan(_unsafeBytes: empty)
+                return _overrideLifetime(of: span, to: self)
+            }
+            switch backing {
+            case .pointer(let ptr):
+                let span = RawSpan(_unsafeElements: ptr)
+                return _overrideLifetime(of: span, to: self)
+            case .array(let array):
+                let span = array.span.bytes
```
jakepetroules: Swift 6.2 on Windows: error: value of type '[UInt8]' has no member 'span'

iCharlesHu: Hmm that's strange... should be available on `6.2`. Let me check

iCharlesHu:
In stead of directly expose `String` initializer, we decided to introduce `AsyncBufferSequence.LineSequence` which transforms `AsyncBufferSequence` to `AsyncSequence<String>` to allow line-by-line streaming

Sources/Subprocess/API.swift
```diff
@@ -152,23 +152,19 @@ public func run<
...
-#endif
 public func run<Result, Input: InputProtocol, Error: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
     environment: Environment = .inherit,
     workingDirectory: FilePath? = nil,
     platformOptions: PlatformOptions = PlatformOptions(),
     input: Input = .none,
-    error: Error,
+    error: Error = .discarded,
```
iCharlesHu, line 167: This was accidentally left out from the `~Copyable` PR. I discovered it as I was testing closure based runs

Sources/Subprocess/API.swift
```diff
@@ -258,16 +266,28 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     }
 }
 
-#if SubprocessSpan
-@available(SubprocessSpan, *)
-#endif
+/// Run a executable with given parameters and a custom closure
```
itingliu: Ditto in other places where you use "executable"
```suggestion
/// Run an executable with given parameters and a custom closure
```

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -37,22 +34,23 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+            // If we have more left in buffer, use that
+            guard self.buffer.isEmpty else {
+                return self.buffer.removeFirst()
+            }
+            // Read more data
+            let data = try await self.diskIO.read(
                 upToLength: readBufferSize
             )
-            if data == nil {
+            guard let data = data else {
```
itingliu: Nit: I think you can just do
```suggestion
            guard let data else {
```

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -61,7 +59,15 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
                 #endif
                 return nil
             }
-            return data
+            let createdBuffers = Buffer.createFrom(data)
+            // Most (all?) cases there should be only one buffer
+            // because DispatchData are motsly contiguous
```
itingliu:
```suggestion
            // because DispatchData are mostly contiguous
```

Sources/Subprocess/Buffer.swift
```diff
@@ -150,12 +104,53 @@ extension AsyncBufferSequence.Buffer: Equatable, Hashable {
...
+    /// Unfortunitely `DispatchData.Region` is not available on Linux, hence our own wrapper
+    internal struct Slice: @unchecked Sendable, RandomAccessCollection {
+        typealias Element = UInt8
+
+        internal let bytes: UnsafeBufferPointer<UInt8>
+
+        internal var startIndex: Int { self.bytes.startIndex }
+        internal var endIndex: Int { self.bytes.endIndex }
+
+        internal init(bytes: UnsafeBufferPointer<UInt8>) {
```
itingliu, line 125: Is this missing a startIndex argument? without that, 
```
internal var startIndex: Int { self.bytes.startIndex }
```
This will always return 0 since `UnsafeBufferPointer.startIndex` is always 0. 

And without that it would not match the semantics of the other Slice-like types, where the index of the containing collection is preserved in the subsequence (like `ArraySlice` and `Array`) right?

Sources/Subprocess/Buffer.swift
```diff
@@ -150,12 +104,53 @@ extension AsyncBufferSequence.Buffer: Equatable, Hashable {
...
+            hasher.combine(bytes: ptr)
         }
     }
     #endif
 }
+
+// MARK: - DispatchData.Block
+#if canImport(Darwin) || canImport(Glibc) || canImport(Android) || canImport(Musl)
+extension DispatchData {
+    /// Unfortunitely `DispatchData.Region` is not available on Linux, hence our own wrapper
```
itingliu: Nit: either fix the typo "Unfortunitely" -> "Unfortunately" or just remove the word

```suggestion
    /// `DispatchData.Region` is not available on Linux, hence our own wrapper
```

BTW this doesn't explain why we don't use `Dispatch.Region` on Darwin, and only use our own types for non-Darwin platforms

Sources/Subprocess/Buffer.swift
iCharlesHu: Here we are intentionally keeping a 1 to 1 `Buffer` to `Slice` mapping. We actually don't really care the relative position of the Slices within the source `DispatchData`, as long as the slices are contiguous. This is the same conceptual design with the original `Buffer` to `DispatchData` mapping (as in we only care about individual chunk read). We only switched to slice instead of `DispatchData` because `DispatchData` is not contiguous and we need a contiguous storage in order to vend the `bytes` property.

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -74,6 +80,239 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
     public func makeAsyncIterator() -> Iterator {
         return Iterator(diskIO: self.diskIO)
     }
+
+    public func lines<Encoding: _UnicodeEncoding>(
+        encoding: Encoding.Type = UTF8.self,
+        bufferingPolicy: LineSequence<Encoding>.BufferingPolicy = .unbounded
```
parkera: Maybe we should label all 'new' API since 0.0.1 (including the ones I worked on) with some kind of comment so we know which ones we need to have further discussion about in review. Here I would like to talk about the tradeoff when defaulting to an unbounded buffer, with no arguments...

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -74,6 +80,239 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+                    let actualBytesCopied = buffer.data.copyBytes(
+                        to: temporary,
+                        count: buffer.data.count
+                    )
+
+                    // Calculate how many CodePoint elements we have
+                    let elementCount = actualBytesCopied / MemoryLayout<Encoding.CodeUnit>.stride
+
+                    // Create array by copying from the buffer reinterpreted as CodePoint
+                    let result: Array<Encoding.CodeUnit> = Array(
```
parkera: Isn't this two copies? Can we create the array directly from the buffer instead?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -74,6 +80,239 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+                        self.buffer += nextBuffer
+                    }
+                    // Step 2: Iterate through buffer to find next line
+                    var currentIndex: Int = self.startIndex
+                    for index in self.startIndex ..< self.buffer.count {
+                        currentIndex = index
+                        // Early return if we exceed max line length
+                        if case .maxLineLength(let maxLength) = self.bufferingPolicy,
+                           currentIndex >= maxLength {
+                            return yield(at: currentIndex)
```
parkera: Could this break in the middle of a UTF8 character (let alone a grapheme cluster)

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -74,6 +80,239 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+                            guard let result = yield(at: currentIndex + 1) else {
+                                continue
+                            }
+                            return result
+                        case newLine:
+                            var targetIndex = currentIndex
+                            if Encoding.CodeUnit.self is UInt8.Type {
+                                // For UTF8, look for the next 0x85 byte
+                                guard (targetIndex + 1) < self.buffer.count,
+                                      self.buffer[targetIndex + 1] == Encoding.CodeUnit(0x85) else {
```
parkera: Should this use the list of possible new line characters above?

Sources/Subprocess/IO/Output.swift
```diff
@@ -151,50 +137,34 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol {
...
-                    continuation.resume(returning: data)
-                    #else
-                    continuation.resume(returning: data.array())
-                    #endif
-                case .failure(let error):
-                    continuation.resume(throwing: error)
-                }
-            }
+        guard let diskIO = diskIO else {
+            // Show not happen due to type system constraints
```
parkera: If this shouldn't happen due to type system constraints, is there a way to make that expressed in the type system instead of a fatal error at runtime?

Sources/Subprocess/Teardown.swift
```diff
@@ -88,7 +85,7 @@ extension Execution {
         using sequence: some Sequence<TeardownStep> & Sendable,
         on processIdentifier: ProcessIdentifier
     ) async {
-        await withUncancelledTask {
+        await withUncanceledTask {
```
parkera: This was actually the preferred spelling with two `l`s.

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: Currently I track them by tagging the PRs with `API Change` label. I'll add some comments and when we are ready for `0.0.1` I'll list out all the changes.

iCharlesHu: I chose `.unbounded` as the default argument because that's what `AsyncStream` do today.

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: This is a great point. I talked to @jmschonfeld and now we lean towards `throw` an error instead of returning here for two reasons:
- Since we can't break anywhere (in the middle of multi-byte encoding), we won't be able to honor `maxLength` exactly (for example, we might have to return 99 bytes instead of 100 because the last byte is the leading byte of a multi-byte character).
- More importantly, we might return the "wrong" result. If the string is broken between grapheme clusters, then we are technically returning the wrong value.

Sources/Subprocess/Teardown.swift
iCharlesHu: Interestingly Xcode thinks `Uncancelled` is misspelled

iCharlesHu:
Resolves https://github.com/swiftlang/swift-subprocess/issues/15

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -74,6 +80,259 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
...
+
+                func yield(at endIndex: Int) -> String? {
+                    defer {
+                        self.buffer.removeFirst(endIndex)
+                        self.startIndex = 0
+                    }
+                    if self.buffer.isEmpty {
+                        return nil
+                    }
+                    return String(decoding: self.buffer[0 ..< endIndex], as: Encoding.self)
```
parkera, line 171: I'm still a little unclear on why we need to copy this into an Array, and then copy it into a String. Why can't String read from the buffer data directly?

Sources/Subprocess/Buffer.swift
itingliu: Bikeshed: As we talked previously, since this is not really representing a sub-sequence like a `Substring` or `ArraySlice`, can we not call it `Slice`? It sounds like we're really interested in ensuring that it's contiguous, so perhaps `ContiguousRegion` or something like that?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -61,7 +59,15 @@ public struct AsyncBufferSequence: AsyncSequence, Sendable {
                 #endif
                 return nil
             }
-            return data
+            let createdBuffers = Buffer.createFrom(data)
```
parkera, line 62: Style nitpick, but this is just `init`.

parkera: Or, in this case it sort of goes the other way since it vends buffers from data... I guess we can clean it up later.

Sources/Subprocess/Buffer.swift
```diff
@@ -150,12 +104,55 @@ extension AsyncBufferSequence.Buffer: Equatable, Hashable {
...
         }
     }
     #endif
 }
+
+// MARK: - DispatchData.Block
+#if canImport(Darwin) || canImport(Glibc) || canImport(Android) || canImport(Musl)
+extension DispatchData {
+    /// Unfortunately `DispatchData.Region` is not available on Linux, hence our own wrapper
+    internal struct Slice: @unchecked Sendable, RandomAccessCollection {
```
itingliu: Why is this `@unchecked Sendable`?

Sources/Subprocess/Buffer.swift
```diff
@@ -150,12 +104,55 @@ extension AsyncBufferSequence.Buffer: Equatable, Hashable {
...
+        internal init(bytes: UnsafeBufferPointer<UInt8>) {
+            self.bytes = bytes
+        }
+
+        internal func withUnsafeBytes<ResultType>(_ body: (UnsafeRawBufferPointer) throws -> ResultType) rethrows -> ResultType {
+            return try body(UnsafeRawBufferPointer(self.bytes))
+        }
+
+        @discardableResult
+        internal func copyBytes<DestinationType>(
```
itingliu: Is this used at all? I'm not seeing any callsites?

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -929,10 +953,124 @@ extension SubprocessUnixTests {
...
+            var result = ""
+            for testCase in testCases {
+                result += testCase.value
+            }
+            try result.write(to: url, atomically: true, encoding: .utf8)
+            #endif
+        }
+
+        let testCaseCount = 60
+        let testFilePath = URL.temporaryDirectory.appending(path: "NewLines.txt")
```
parkera: Throw a `UUID` in here, or test cases that are running in parallel will reuse the same file.

Sources/Subprocess/AsyncBufferSequence.swift
itingliu: If we decide to throw an error for https://github.com/swiftlang/swift-subprocess/pull/45 
can we please have an issue tracking updating this to the same behavior?

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: We have to keep an `[UInt8]` buffer in order to scan through and find new line characters. We can't use `String` directly as the buffer type because every time when we receive a chunk of data, we can't directly convert it to String since the chunk might end in between byte encodings.

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: And also it creates multiple. I didn't use initializer because it's `DispatchData -> [Buffer]`.

Sources/Subprocess/Buffer.swift
iCharlesHu: Because it holds a `let bytes: UnsafeBufferPointer<UInt8>` property. `DispatchData.Region`'s backing storage is also `@unchecked Sendable`, I suspect for the same reason.

PR #53, opened 2025-06-03
Fix sample usage of closure

giginet:
I found the README is not correct for the current main.


iCharlesHu:
Hi @giginet thanks for the fix! Unfortunately this example is no longer correct with https://github.com/swiftlang/swift-subprocess/pull/48 and I'll make sure to update the example with that PR.

PR #61, opened 2025-06-06
Add an overload that allows supplying custom stdout and stderr destinations AND a body

jakepetroules:
Every other combination of stdin/stdout/stderr with a custom body was present except for this one.

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -152,6 +152,67 @@ public func run<
 ///   - platformOptions: The platform specific options to use
 ///     when running the executable.
 ///   - input: The input to send to the executable.
+///   - output: How to manager executable standard output.
+///   - error: How to manager executable standard error.
+///   - isolation: the isolation context to run the body closure.
+///   - body: The custom execution body to manually control the running process
+/// - Returns an executableResult type containing the return value
+///     of the closure.
+public func run<Result, Input: InputProtocol, Output: OutputProtocol, Error: OutputProtocol>(
+    _ executable: Executable,
+    arguments: Arguments = [],
+    environment: Environment = .inherit,
+    workingDirectory: FilePath? = nil,
+    platformOptions: PlatformOptions = PlatformOptions(),
+    input: Input = .none,
+    output: Output = .discarded,
+    error: Error = .discarded,
+    isolation: isolated (any Actor)? = #isolation,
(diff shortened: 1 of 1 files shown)


jakepetroules:
My use case is that I want to be able to point stdout/stderr at the same file descriptor, AND I want to support cancellation. With the current set of overloads I can only do one or the other.

Sources/Subprocess/API.swift
```diff
@@ -142,6 +142,74 @@ public func run<
...
+    _ executable: Executable,
+    arguments: Arguments = [],
+    environment: Environment = .inherit,
+    workingDirectory: FilePath? = nil,
+    platformOptions: PlatformOptions = PlatformOptions(),
+    input: Input = .none,
+    output: Output,
+    error: Error,
+    isolation: isolated (any Actor)? = #isolation,
+    body: ((Execution, AsyncBufferSequence, AsyncBufferSequence) async throws -> Result)
```
iCharlesHu: Since you are accepting output and error explicitly, the body closure shouldn't have AsyncBufferSequence` because that's fundamentally incompatible with explicit output types. Here `body` should be just `(Execution) async throws -> Result`

jakepetroules: Done!

Sources/Subprocess/API.swift
```diff
@@ -142,6 +142,67 @@ public func run<
...
+///     of the closure.
+public func run<Result, Input: InputProtocol, Output: OutputProtocol, Error: OutputProtocol>(
+    _ executable: Executable,
+    arguments: Arguments = [],
+    environment: Environment = .inherit,
+    workingDirectory: FilePath? = nil,
+    platformOptions: PlatformOptions = PlatformOptions(),
+    input: Input = .none,
+    output: Output,
+    error: Error,
```
iCharlesHu: Could you add default values so it's consistent with the rest of `run()` family?

iCharlesHu (approved):

Sources/Subprocess/API.swift
jakepetroules: However, how do you reconcile that with the fact that other overloads DO include the AsyncBufferSequence in the body closure? Should those be removed, then?

Sources/Subprocess/API.swift
jakepetroules: Done

Issue #47, opened 2025-05-16
Use `epoll` and non-blocking read/write instead of DispatchIO on Linux


iCharlesHu:
(manually cloned from https://github.com/iCharlesHu/Subprocess/pull/61)

We are getting bad pointer deref on Linux deep within `libdispatch`:
```
*** Signal 11: Backtracing from 0xffff8d4ccdf0... done ***

*** Program crashed: Bad pointer dereference at 0x0000000000000050 ***

Thread 0:

0  0x0000ffff8dad43f4 initializeBufferWithCopyOfBuffer for Range in libswiftCore.so

Thread 1 crashed:

0      0x0000ffff8d4ccdf0 _dispatch_event_loop_drain + 1216 in libdispatch.so
1 [ra] 0x0000ffff8d4c1e38 _dispatch_mgr_invoke + 123 in libdispatch.so
2 [ra] 0x0000ffff8d4c1dac _dispatch_mgr_thread + 131 in libdispatch.so
3 [ra] 0x0000ffff8d4c5118 _dispatch_worker_thread + 431 in libdispatch.so
4 [ra] 0x0000ffff8c27595c <unknown> in libc.so.6

Thread 2:

0  0x0000ffff78022900


Registers:

 x0 0x0000000000000001  1
 x1 0x0000000000000081  129
 x2 0x0000000000000001  1
 x3 0x0000000000000000  0
 x4 0x0000000000000000  0
 x5 0x0000000000000008  8
 x6 0x0000000000000000  0
 x7 0x0000000000000004  4
 x8 0x0000000000000062  98
 x9 0x001ffe1c80000000  9005122638053376
x10 0x001ffe1c80000000  9005122638053376
x11 0x00007ff840000000  140704202358784
x12 0x0000000000000000  0
x13 0x0000000000000000  0
x14 0x000006cb6e4c0bf9  7470798605305
x15 0x0000000029aacefe  699059966
x16 0x0000000000000001  1
x17 0x0000ffff8c27da50  fd 7b bd a9 fd 03 00 91 f3 53 01 a9 f4 03 00 aa  ý{½©ý···óS·©ô··ª
x18 0x000000000000001d  29
x19 0x00000000ffffffff  4294967295
x20 0x0000000000000030  48
x21 0x0000000000000000  0
x22 0x0000000000000000  0
x23 0x0000000000000003  3
[...]

PR #64, opened 2025-06-10
Create platform specific AsyncIO

iCharlesHu:
This PR introduces platform specific `AsyncIO`:

- Darwin: based on DispatchIO
- Linux: based on epoll
- (TODO) Windows: based on IOCP with OVERLAPPED

The `epoll` based `AsyncIO` replaces `DispatchIO` on Linux.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/47


Sources/Subprocess/Buffer.swift
```diff
@@ -95,9 +95,7 @@ extension AsyncBufferSequence.Buffer {
 
 // MARK: - Hashable, Equatable
 extension AsyncBufferSequence.Buffer: Equatable, Hashable {
-    #if os(Windows)
-    // Compiler generated conformances
-    #else
+    #if canImport(Darwin)
```
itingliu, line 95: I know these aren't not new changes, but why do we have custom implementation only for Darwin?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,784 @@
...
+        return try await self.read(from: diskIO.fileDescriptor, upTo: maxLength)
+    }
+
+    func read(
+        from fileDescriptor: FileDescriptor,
+        upTo maxLength: Int
+    ) async throws -> [UInt8]? {
+        // If we are reading until EOF, start with readBufferSize
+        // and gradually increase buffer size
+        let bufferLength = maxLength == .max ? readBufferSize : maxLength
```
itingliu, line 348: Can we just respect the passed-in `maxLength` as-is and not impose this silent cap? 
This feels deceiving from callsites perspective and I imagine it'd make it difficult to figure out what's going on

Sources/Subprocess/IO/Output.swift
```diff
@@ -266,12 +266,11 @@ extension OutputProtocol {
             return () as! OutputType
         }
 
-        #if os(Windows)
-        let result = try await diskIO?.fileDescriptor.read(upToLength: self.maxSize)
-        return try self.output(from: result ?? [])
-        #else
-        let result = try await diskIO!.dispatchIO.read(upToLength: self.maxSize)
+        let result = try await AsyncIO.shared.read(from: diskIO!, upTo: self.maxSize)
```
itingliu: Is this force unwrap correct?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,784 @@
...
+            }
+        }
+
+        // Create thread data
+        let context = MonitorThreadContext(
+            epollFileDescriptor: self.epollFileDescriptor,
+            shutdownFileDescriptor: self.shutdownFileDescriptor
+        )
+        let threadContext = Unmanaged.passRetained(context)
+        #if os(macOS) || os(FreeBSD) || os(OpenBSD)
```
itingliu: This whole block is `#if canImport(Glibc) || canImport(Android) || canImport(Musl)`. Would we ever need to handle `os(macOS)`?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,784 @@
...
+        self.monitorThread = thread
+        #endif
+        if rc != 0 {
+            maybeSetupError = SubprocessError(
+                code: .init(.asyncIOFailed("Failed to create monitor thread")),
+                underlyingError: .init(rawValue: rc)
+            )
+        }
+
+        self.setupError = maybeSetupError
```
itingliu: Why can't we throw this error right here in `init`?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,784 @@
...
+import _SubprocessCShims
+import Synchronization
+
+private typealias SignalStream = AsyncThrowingStream<Bool, any Error>
+private let _epollEventSize = 256
+private let _registration: Mutex<
+    [PlatformFileDescriptor : SignalStream.Continuation]
+> = Mutex([:])
+
+final class AsyncIO: Sendable {
```
itingliu, line 40: On Darwin, `class AsyncIO` doesn't have any stored variables. That means `AsyncIO.shared.read()`  can also be a static function, and you would just write `AsyncIO.read(...)`

I think you left it this way because on linux it has stored variables so they can't be static funcs, but we still want to share the same source at callsites? If so would it make more sense to move the `#if platform` guards into the shared functions? So instead of

```
#if linux
class AsyncIO {
    func read(...)
}
#elif mac 
class AsyncIO {
    func read(...)
}
#endif
```

Can we have
```
class AsyncIO {
    func read(...) {
        #if linux
         ...
        #elif mac 
         ...
        #endif
    }
}
```

iCharlesHu:
> What's the strategy for other Unix platforms like FreeBSD?

We don't currently have a plan to support `FreeBSD`, however, we are definitely open to it. Although I'm not sure logistically how this would work because in order to officially support a platform, we'd have to have CI for it. Since FreeBSD isn't an officially supported platform for Swift, I'm not sure if we can do that just yet (@shahmishal please correct me if I'm wrong).

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: I decided to go with the first approach because the implementation for each platform is vastly different. So having `#if #else` inside of each function makes this file overall not readable. This is especially true now I'm adding Windows-specific implementation on top of this. It's much more maintainable if each platform has its own "section" of code so the overall flow is more clear.

Sources/Subprocess/Buffer.swift
iCharlesHu: `DispatchData` is unfortunately not `Hashable`

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: Here we need to preallocate buffer of a certain length. We can't just pass in `.max` to allocate infinite buffer so we have to grow it manually similar like an array (all this is so we can avoid copying and write into the buffer directly from `read()`).

Sources/Subprocess/IO/Output.swift
iCharlesHu: Yes. Line 262 above 
```
if OutputType.self == Void.self {
    return () as! OutputType
}
```
Take cares of the `nil` possibility. I'll add a comment.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: Ahh good catch. Thanks!

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: `AsyncIO` is designed to be a singleton `static let` so we only spin up one additional thread for monitoring. Unfortunately this also means we can't have failable initializer...

iCharlesHu:
Rebased

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -155,7 +146,16 @@ extension AsyncBufferSequence {
                             UnsafeBufferPointer(start: ptr.baseAddress?.assumingMemoryBound(to: Encoding.CodeUnit.self), count: elementCount)
                         )
                     }
-
+                    #else
+                    // Cast data to CodeUnit type
+                    let result = buffer.withUnsafeBytes { ptr in
```
cthielen, line 153: What about:

```swift
Buffer.withUnsafeBytes { ptr in
    ptr.withMemoryRebound(to: Encoding.CodeUnit.self) { ptr in
        return Array(ptr)
    }
}
```

I believe it's the preferred approach for this sort of thing and also bypasses the `.stride` vs. `.size` issue.

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,781 @@
...
+    }
+
+    private let epollFileDescriptor: CInt
+    private let shutdownFileDescriptor: CInt
+    private let setupError: SubprocessError?
+
+    private let monitorThread: pthread_t
+
+    private init() {
+        var maybeSetupError: SubprocessError? = nil
```
cthielen: Could `self.setupError` be Optional to avoid the need for `maybeSetupError`?

cthielen: Is it possible for multiple errors to occur and start overriding `maybeSetupError`?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,781 @@
...
+        self.setupError = maybeSetupError
+        atexit {
+            AsyncIO.shared.shutdown()
+        }
+    }
+
+    private func shutdown() {
+        var one: UInt64 = 1
+        // Wake up the thread for shutdown
+        _ = _SubprocessCShims.write(self.shutdownFileDescriptor, &one, MemoryLayout<UInt64>.size)
```
cthielen: I think you want `.stride` instead of `.size`.

It doesn't matter for meta types, but `.stride` is more what you're using conceptually.

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,781 @@
...
+
+    func _write<Bytes: _ContiguousBytes>(
+        _ bytes: Bytes,
+        to diskIO: borrowing TrackedPlatformDiskIO
+    ) async throws -> Int {
+        let fileDescriptor = diskIO.fileDescriptor
+        let signalStream = self.registerFileDescriptor(fileDescriptor, for: .write)
+        var writtenLength: Int = 0
+        for try await _ in signalStream {
+            while true {
```
cthielen, line 433: Do we need any form of sleeping or waiting here? e.g. if the OS returns EAGAIN or EWOULDBLOCK, does this while loop eat CPU cycles?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,781 @@
...
+    func write(
+        _ span: borrowing RawSpan,
+        to diskIO: borrowing TrackedPlatformDiskIO
+    ) async throws -> Int {
+        let fileDescriptor = diskIO.fileDescriptor
+        let signalStream = self.registerFileDescriptor(fileDescriptor, for: .write)
+        var writtenLength: Int = 0
+        for try await _ in signalStream {
+            while true {
+                let written = span.withUnsafeBytes { ptr in
```
cthielen, line 477: Same comment here about whether a sleep is appropriate.

Sources/Subprocess/IO/Output.swift
```diff
@@ -142,14 +142,12 @@ public struct BytesOutput: OutputProtocol {
     internal func captureOutput(
         from diskIO: consuming TrackedPlatformDiskIO?
     ) async throws -> [UInt8] {
-        #if os(Windows)
-        let result = try await diskIO?.fileDescriptor.read(upToLength: self.maxSize) ?? []
-        try diskIO?.safelyClose()
-        return result
-        #else
-        let result = try await diskIO!.dispatchIO.read(upToLength: self.maxSize)
+        let result = try await AsyncIO.shared.read(from: diskIO!, upTo: self.maxSize)
```
cthielen: If it's safe to force unwrap here, should this function be taking an optional?

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: I'm changing to be `Result` so it's more concise.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: No. So if we see `EAGAIN` or `EWOULDBLOCK` we break out of the inner while loop:
```
if errno == EAGAIN || errno == EWOULDBLOCK {
    // No more data for now wait for the next signal
    break
}
```
Then we wait for the next item in the signalStream to come `for try await _ in signalStream` which tells us the item is ready to be read. Basically we are replying on Swift concurrency to `await`.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: Ditto. The key is the `break` which break out of the inner `while` loop. The outer `for try await` loop is naturally `await`ed based on epoll

Sources/Subprocess/IO/Output.swift
iCharlesHu: Yes, this is taken care of at type level. Only `DiscardedOutput` have `nil` diskIO so we can safely unwrap here.

Sources/Subprocess/IO/AsyncIO.swift
itingliu: Can we throw an error in that case?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,808 @@
...
+    ) async throws -> [UInt8]? {
+        return try await self.read(from: diskIO.fileDescriptor, upTo: maxLength)
+    }
+
+    func read(
+        from fileDescriptor: FileDescriptor,
+        upTo maxLength: Int
+    ) async throws -> [UInt8]? {
+        return try await withCheckedThrowingContinuation { continuation in
+            DispatchQueue.global(qos: .userInitiated).async {
```
itingliu, line 690: Can we use a Task/TaskGroup instead of `DispatchQueue`?

Sources/Subprocess/IO/AsyncIO.swift
itingliu: Which function sets `errno` here?

I think @cthielen has a point because it's not super obvious that we'd be guaranteed to break the loop.

Sources/Subprocess/IO/Output.swift
itingliu: This is code smell though... if we never expect it to be nil then we should mark it that way. That's probably out of scope of this PR, but it's not the first time this question came up.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: I added a comment to explain the flow in more detail.

Sources/Subprocess/IO/Output.swift
iCharlesHu: I refactored the code to centralize "unwrapping" to one location and used `fatalError()` to make it clear it is a programmer error to have nil diskIO.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: In case maxLength is `.max`? No that's the intended use case. This API is designed to be able to read arbitrary length until we reach EOF. Since we don't know how many bytes the child process will produce, somewhere down the call stack we have to similar thing (grow buffer as needed in order to read until EOF).

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: This is old code I just moved it. The replacement is incoming (actually already ready just need this PR to land first). The `tl;dr` is we have to use `DispatchQueue.aync` in order to not block Swift Task's corporative thread pool, since line below `ReadFile` is a blocking API. This was my temporary solution while building up the API.

Sources/Subprocess/Buffer.swift
```diff
@@ -92,26 +92,23 @@ extension AsyncBufferSequence.Buffer {
...
-            hasher.combine(bytes: ptr)
-        }
+        hasher.combine(self.data)
     }
     #endif
+    // else Compiler generated conformances
 }
 
 // MARK: - DispatchData.Block
 #if canImport(Darwin) || canImport(Glibc) || canImport(Android) || canImport(Musl)
```
jakepetroules, line 108: Shouldn't this be ONLY `#if canImport(Darwin)` if we aren't using Dispatch on other platforms?

itingliu (approved):

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -157,7 +148,13 @@ extension AsyncBufferSequence {
                             UnsafeBufferPointer(start: ptr.baseAddress?.assumingMemoryBound(to: Encoding.CodeUnit.self), count: elementCount)
                         )
                     }
-
+                    #else
+                    // Cast data to CodeUnitg type
```
jakepetroules, line 152: nit: spelling error "CodeUnitg"

Sources/Subprocess/IO/Output.swift
```diff
@@ -140,16 +140,14 @@ public struct BytesOutput: OutputProtocol {
...
     ) async throws -> [UInt8] {
-        #if os(Windows)
-        let result = try await diskIO?.fileDescriptor.read(upToLength: self.maxSize) ?? []
-        try diskIO?.safelyClose()
-        return result
-        #else
-        let result = try await diskIO!.dispatchIO.read(upToLength: self.maxSize)
-        try diskIO?.safelyClose()
+        let result = try await AsyncIO.shared.read(from: diskIO, upTo: self.maxSize)
+        try diskIO.safelyClose()
```
jakepetroules, line 146: I know this is an existing issue but note that safelyClose will fail to get called if read throws an error, which is probably going to cause a file descriptor leak?

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -524,4 +524,30 @@ internal func monitorProcessTermination(
...
+    internal consuming func createPlatformDiskIO() -> TrackedPlatformDiskIO {
+        // Transferring out the ownership of fileDescriptor means we don't have go close here
+        let shouldClose = self.closeWhenDone
+        let closeFd = self.fileDescriptor
+        let dispatchIO: DispatchIO = DispatchIO(
+            type: .stream,
+            fileDescriptor: self.platformDescriptor(),
+            queue: .global(),
+            cleanupHandler: { error in
+                // Close the file descriptor
```
jakepetroules, line 539: Not new code and not a blocker for this PR but just wanted to point out #106 again since you're making changes in this area. I've actually had a lot of issues with this particular DispatchIO initializer (the cleanup handler semantics are somewhat unclear) and you may be better served by calling the single-shot API inside of TrackedFileDescriptor; see what I've done in https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/Dispatch%2BAsync.swift for example.

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,826 @@
...
+                    }
+                }
+                if let lastError = lastError {
+                    let windowsError = SubprocessError(
+                        code: .init(.failedToReadFromSubprocess),
+                        underlyingError: .init(rawValue: lastError)
+                    )
+                    continuation.resume(throwing: windowsError)
+                } else {
+                    continuation.resume(returning: values)
```
jakepetroules, line 740: This line causes all I/O consumption to hang, which is why the tests are hanging; you need to return `nil` to indicate EOF if the return value is empty.

If I make that change, all the Windows tests pass successfully.

```suggestion
                    continuation.resume(returning: values.isEmpty ? nil : values)
```

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -0,0 +1,826 @@
...
+        let shutdownFileDescriptor: CInt
+        let monitorThread: pthread_t
+    }
+
+    static let shared: AsyncIO = AsyncIO()
+
+    private let state: Result<State, SubprocessError>
+
+    private init() {
+        // Create main epoll fd
```
jakepetroules, line 73: Note that `epoll` is a Linux-specific API.

I don't think it should block your initial PR here, but would you be able to look at a pure POSIX path (based on regular `poll` instead of `epoll`?) that we could fall back to for other non-Linux POSIX platforms like FreeBSD?

(Note that as a BSD, FreeBSD also has kqueue which is probably preferable there, but that could come later)

jakepetroules (changes requested):

iCharlesHu:
Superseded by https://github.com/swiftlang/swift-subprocess/pull/117

Issue #49, opened 2025-05-21
Use with-style methods instead of `~Copyable` types for file descriptors management


FranzBusch:
In a recent PR, the code changed over to model file descriptors as `~Copyable` types. There are two problems with this:
1. Closing a file descriptor can be potentially asynchronous
2. Closing a file descriptor can fail

Neither of those two things can be modeled within a `deinit`. Currently, the best practice in the ecosystem is to use `with-style` methods to handle such resources e.g.:

```
func withFileDescriptor(body: (FileDescriptor) async throws -> Void) async throws -> Void
```

PR #72, opened 2025-06-12
Fatal error if TrackedFileDescriptor and TrackedDispatchIO have not been closed in deinit

iCharlesHu:
We cannot depend on calling .close() in deinit since it’s non-failable. As a result, Subprocess needs to manually call .close() at the conclusion of the read/write operation. fatalError in deinit to ensures the FileDescriptor/DispatchIO will not leak.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/49

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -632,26 +632,7 @@ internal struct TrackedFileDescriptor: ~Copyable {
             return
         }
 
-        do {
-            try fileDescriptor.close()
-        } catch {
-            guard let errno: Errno = error as? Errno else {
-                return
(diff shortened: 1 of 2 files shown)


parkera:
cc @FranzBusch @weissi

Sources/Subprocess/Configuration.swift
```diff
@@ -632,26 +632,7 @@ internal struct TrackedFileDescriptor: ~Copyable {
             return
         }
```
weissi, line 634: there's still this `guard self.closeWhenDone` here which is probably not intended as we're never closing here (which is great)

PR #73, opened 2025-06-12
Add CMake build

etcwilde:
Adding a CMake build so that projects that use CMake to bootstrap, like SwiftPM and swift-build, can use this library.
This is an initial pass. Future work should include a mechanism to use a pre-built Swift-System and an installation story that generates a `SwiftSubprocessConfig.cmake` file. Changes are necessary in the Swift-System CMake files in order to have a proper installation story here though because the SwiftSystem library is not properly exported.

Unlike SwiftPM, CMake does not automatically generate modulemaps for C headers, so we need to include one.

I'm not sure how the `mach/vm_page_size.h` header was getting included, but I was getting errors about the `vm_page_size` variable not being declared. If the header exists, include it so that variable is defined.

Also added a workflow to verify that the CMake build continues to work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,9 @@
+cmake_minimum_required(VERSION 3.29)
+project(Subprocess LANGUAGES C Swift)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name ${PROJECT_NAME}>")
+
+include(InstallExternalDependencies)
+
+add_subdirectory(Sources)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.29)
+cmake_minimum_required(VERSION 3.26...3.29)
 project(Subprocess LANGUAGES C Swift)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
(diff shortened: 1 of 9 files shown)


cmake/modules/InstallExternalDependencies.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+include_guard()
+
+# TODO: Use find_package to find a pre-built SwiftSystem
+
+include(FetchContent)
+
+FetchContent_Declare(SwiftSystem
+  GIT_REPOSITORY https://github.com/apple/swift-system.git
+  GIT_TAG a34201439c74b53f0fd71ef11741af7e7caf01e1 # 1.4.2
```
jakepetroules, line 20: @shahmishal @owenv FYI since we're currently trying to get this to 1.5.0.

jakepetroules (approved):
Thanks!

CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+##
+## Copyright (c) 2025 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See https://swift.org/LICENSE.txt for license information
+##
+##===----------------------------------------------------------------------===##
+
+cmake_minimum_required(VERSION 3.26...3.29)
+project(SwiftSubprocess LANGUAGES C Swift)
```
iCharlesHu: In `Package.swift` we use `Subprocess` as the library and package name. Should we keep it consistent?

CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+##===----------------------------------------------------------------------===##
```
iCharlesHu, line 1: Could you also update `Package.swift` to explicitly exclude all `CMakeLists.txt` files?

PR #74, opened 2025-06-12
Make Configuration.workingDirectory optional

jakepetroules:
On all platforms, retrieving the current working directory is a failable operation. When a Configuration is given a null working directory, instead of resolving that to the current working directory at initialization time (which can't be done safely because Configuration.init is non-failable), make Configuration.workingDirectory optional and preserve any nil input through to process creation time. Both CreateProcess (Windows) and posix_spawn or fork/exec (POSIX) inherit the working directory of the calling process if it is not specified, so this doesn't change any behavior.

It also better matches the design of other properties on the Configuration struct like the Executable, which defer resolution of their concrete value rather than doing so at Configuration.init time.

Lastly, it removes incorrect code: the implementation of FilePath.currentWorkingDirectory could crash on POSIX platforms if the current directory was no longer accessible or getcwd failed for any other reason, and on Windows it would not properly support non-ASCII characters in path names nor path names longer than 260 characters.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -39,7 +39,8 @@ public struct Configuration: Sendable {
     /// The environment to use when running the executable.
     public var environment: Environment
     /// The working directory to use when running the executable.
-    public var workingDirectory: FilePath
+    /// If this property is `nil`, the subprocess will inherit the working directory from the parent process.
+    public var workingDirectory: FilePath?
     /// The platform specific options to use when
     /// running the subprocess.
     public var platformOptions: PlatformOptions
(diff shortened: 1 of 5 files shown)


Sources/Subprocess/Configuration.swift
```diff
@@ -39,7 +39,7 @@ public struct Configuration: Sendable {
     /// The environment to use when running the executable.
     public var environment: Environment
     /// The working directory to use when running the executable.
```
iCharlesHu, line 41: Could you update the the doc to mention something like "if nil, the child process will inherit working directory from parent process"

iCharlesHu (approved):

Sources/Subprocess/Configuration.swift
jakepetroules: Done!

Issue #46, opened 2025-05-16
fork+exec fails to close other file descriptors


iCharlesHu:
https://github.com/iCharlesHu/swift-experimental-subprocess/blob/7204d94bc468c0669a326fe584bf240e1eecbb2b/Sources/_SubprocessCShims/process_shims.c#L322

UNIX programs typically need to close all other open fds that they don't want to inherit into the child process after fork() but before execve(). This code fails to do so, this is security-relevant.

Implementation here: https://github.com/swiftlang/swift-sdk-generator/pull/156/files#diff-456340af052c33f5e1551237a5b2b7a6f20c374fea29d0d352c931b328a319a4R134-R140

(manually cloned from https://github.com/iCharlesHu/Subprocess/issues/44)

PR #79, opened 2025-06-13
Emulate POSIX_SPAWN_CLOEXEC_DEFAULT in fork/exec

iCharlesHu:
POSIX_SPAWN_CLOEXEC_DEFAULT is only available on Darwin. Emulate POSIX_SPAWN_CLOEXEC_DEFAULT on other platforms by calling close after fork, before exec.

This commit also removes _subprocess_posix_spawn_fallback because we can't emulate POSIX_SPAWN_CLOEXEC_DEFAULT in a thread-safe manner while using posix_spawn.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/46
Resolves: https://github.com/swiftlang/swift-subprocess/issues/132

### DIFF
--- a/Sources/_SubprocessCShims/process_shims.c
+++ b/Sources/_SubprocessCShims/process_shims.c
@@ -34,8 +34,12 @@
 #include <string.h>
 #include <fcntl.h>
 #include <pthread.h>
-
+#include <dirent.h>
 #include <stdio.h>
+
+#if __has_include(<linux/close_range.h>)
+#include <linux/close_range.h>
+#endif
 
 #if __has_include(<crt_externs.h>)
 #include <crt_externs.h>
@@ -50,6 +54,7 @@ extern char **environ;
 #include <mach/vm_page_size.h>
 #endif
 
+#if !TARGET_OS_WINDOWS
 int _was_process_exited(int status) {
     return WIFEXITED(status);
 }
@@ -70,8 +75,7 @@ int _was_process_suspended(int status) {
     return WIFSTOPPED(status);
 }
 
-#if TARGET_OS_LINUX
-#include <stdio.h>
+#endif // !TARGET_OS_WINDOWS
 
 #ifndef SYS_pidfd_send_signal
 #define SYS_pidfd_send_signal 424
@@ -98,40 +102,6 @@ vm_size_t _subprocess_vm_size(void) {
     // This shim exists because vm_page_size is not marked const, and therefore looks like global mutable state to Swift.
     return vm_page_size;
 }
-#endif
-
-// MARK: - Private Helpers
-static pthread_mutex_t _subprocess_fork_lock = PTHREAD_MUTEX_INITIALIZER;
-
-static int _subprocess_block_everything_but_something_went_seriously_wrong_signals(sigset_t *old_mask) {
-    sigset_t mask;
-    int r = 0;
-    r |= sigfillset(&mask);
-    r |= sigdelset(&mask, SIGABRT);
-    r |= sigdelset(&mask, SIGBUS);
-    r |= sigdelset(&mask, SIGFPE);
-    r |= sigdelset(&mask, SIGILL);
-    r |= sigdelset(&mask, SIGKILL);
-    r |= sigdelset(&mask, SIGSEGV);
-    r |= sigdelset(&mask, SIGSTOP);
-    r |= sigdelset(&mask, SIGSYS);
-    r |= sigdelset(&mask, SIGTRAP);
-
-    r |= pthread_sigmask(SIG_BLOCK, &mask, old_mask);
-    return r;
-}
-
-#define _subprocess_precondition(__cond) do { \
-    int eval = (__cond); \
-    if (!eval) { \
-        __builtin_trap(); \
-    } \
-} while(0)
-
-#if __DARWIN_NSIG
-#  define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
-#else
-#  define _SUBPROCESS_SIG_MAX 32
 #endif
 
 
@@ -374,6 +344,100 @@ static int _clone3(int *pidfd) {
     return syscall(SYS_clone3, &args, sizeof(args));
 }
 
+static pthread_mutex_t _subprocess_fork_lock = PTHREAD_MUTEX_INITIALIZER;
+
+static int _subprocess_make_critical_mask(sigset_t *old_mask) {
+    sigset_t mask;
+    int r = 0;
+    r |= sigfillset(&mask);
+    r |= sigdelset(&mask, SIGABRT);
+    r |= sigdelset(&mask, SIGBUS);
+    r |= sigdelset(&mask, SIGFPE);
+    r |= sigdelset(&mask, SIGILL);
+    r |= sigdelset(&mask, SIGKILL);
+    r |= sigdelset(&mask, SIGSEGV);
+    r |= sigdelset(&mask, SIGSTOP);
+    r |= sigdelset(&mask, SIGSYS);
+    r |= sigdelset(&mask, SIGTRAP);
+
+    r |= pthread_sigmask(SIG_BLOCK, &mask, old_mask);
+    return r;
+}
+
+#define _subprocess_precondition(__cond) do { \
+    int eval = (__cond); \
+    if (!eval) { \
+        __builtin_trap(); \
+    } \
+} while(0)
+
+#if __DARWIN_NSIG
+#  define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
+#else
+#  define _SUBPROCESS_SIG_MAX 32
+#endif
+
+int _shims_snprintf(
+    char * _Nonnull str,
+    int len,
+    const char * _Nonnull format,
+    char * _Nonnull str1,
+    char * _Nonnull str2
+) {
+    return snprintf(str, len, format, str1, str2);
+}
+
+static int _positive_int_parse(const char *str) {
+    char *end;
+    long value = strtol(str, &end, 10);
+    if (end == str) {
+        // No digits found
+        return -1;
+    }
+    if (errno == ERANGE || val <= 0 || val > INT_MAX) {
+        // Out of range
+        return -1;
+    }
+    return (int)value;
+}
+
+static int _highest_possibly_open_fd_dir(const char *fd_dir) {
+    int highest_fd_so_far = 0;
+    DIR *dir_ptr = opendir(fd_dir);
+    if (dir_ptr == NULL) {
+        return -1;
+    }
+
+    struct dirent *dir_entry = NULL;
+    while ((dir_entry = readdir(dir_ptr)) != NULL) {
+        char *entry_name = dir_entry->d_name;
+        int number = _positive_int_parse(entry_name);
+        if (number > (long)highest_fd_so_far) {
+            highest_fd_so_far = number;
+        }
+    }
+
+    closedir(dir_ptr);
+    return highest_fd_so_far;
+}
+
+static int _highest_possibly_open_fd(void) {
+#if defined(__APPLE__)
+    int hi = _highest_possibly_open_fd_dir("/dev/fd");
+    if (hi < 0) {
+        hi = getdtablesize();
+    }
+#elif defined(__linux__)
+    int hi = _highest_possibly_open_fd_dir("/proc/self/fd");
+    if (hi < 0) {
+        hi = getdtablesize();
+    }
+#else
+    int hi = getdtablesize();
+#endif
+    return hi;
+}
+
 int _subprocess_fork_exec(
     pid_t * _Nonnull pid,
     int * _Nonnull pidfd,
@@ -433,7 +497,7 @@ int _subprocess_fork_exec(
     _subprocess_precondition(rc == 0);
     // Block all signals on this thread
     sigset_t old_sigmask;
-    rc = _subprocess_block_everything_but_something_went_seriously_wrong_signals(&old_sigmask);
+    rc = _subprocess_make_critical_mask(&old_sigmask);
     if (rc != 0) {
         close(pipefd[0]);
         close(pipefd[1]);
@@ -556,20 +620,22 @@ int _subprocess_fork_exec(
         if (rc < 0) {
             write_error_and_exit;
         }
-
-        // Close parent side
-        if (file_descriptors[1] >= 0) {
-            rc = close(file_descriptors[1]);
-        }
-        if (file_descriptors[3] >= 0) {
-            rc = close(file_descriptors[3]);
-        }
-        if (file_descriptors[5] >= 0) {
-            rc = close(file_descriptors[5]);
-        }
-
-        if (rc < 0) {
-            write_error_and_exit;
+        // Close all other file descriptors
+        rc = -1;
+        errno = ENOSYS;
+#if __has_include(<linux/close_range.h>) || defined(__FreeBSD__)
+        // We must NOT close pipefd[1] for writing errors
+        rc = close_range(STDERR_FILENO + 1, pipefd[1] - 1, 0);
+        rc |= close_range(pipefd[1] + 1, ~0U, 0);
+#endif
+        if (rc != 0) {
+            // close_range failed (or doesn't exist), fall back to close()
+            for (int fd = STDERR_FILENO + 1; fd < _highest_possibly_open_fd(); fd++) {
+                // We must NOT close pipefd[1] for writing errors
+                if (fd != pipefd[1]) {
+                    close(fd);
+                }
+            }
         }
 
         // Run custom configuratior
@@ -646,8 +712,6 @@ int _subprocess_fork_exec(
 }
 
 #endif // TARGET_OS_LINUX
-
-#endif // !TARGET_OS_WINDOWS
 
 #pragma mark - Environment Locking
 
--- a/Sources/_SubprocessCShims/process_shims.c
+++ b/Sources/_SubprocessCShims/process_shims.c
@@ -36,6 +36,7 @@
 #include <pthread.h>
 #include <dirent.h>
 #include <stdio.h>
+#include <limits.h>
 
 #if __has_include(<linux/close_range.h>)
 #include <linux/close_range.h>
@@ -54,7 +55,6 @@ extern char **environ;
 #include <mach/vm_page_size.h>
 #endif
 
-#if !TARGET_OS_WINDOWS
 int _was_process_exited(int status) {
     return WIFEXITED(status);
 }
@@ -73,26 +73,6 @@ int _get_signal_code(int status) {
 
 int _was_process_suspended(int status) {
     return WIFSTOPPED(status);
-}
-
-#endif // !TARGET_OS_WINDOWS
-
-#ifndef SYS_pidfd_send_signal
-#define SYS_pidfd_send_signal 424
-#endif 
-
-int _shims_snprintf(
-    char * _Nonnull str,
-    int len,
-    const char * _Nonnull format,
-    char * _Nonnull str1,
-    char * _Nonnull str2
-) {
-    return snprintf(str, len, format, str1, str2);
-}
-
-int _pidfd_send_signal(int pidfd, int signal) {
-    return syscall(SYS_pidfd_send_signal, pidfd, signal, NULL, 0);
 }
 
 #endif
@@ -304,6 +284,16 @@ int _pidfd_open(pid_t pid) {
     return syscall(SYS_pidfd_open, pid, 0);
 }
 
+// SYS_pidfd_send_signal is only defined on Linux Kernel 5.1 and above
+// Define our dummy value if it's not available
+#ifndef SYS_pidfd_send_signal
+#define SYS_pidfd_send_signal 424
+#endif
+
+int _pidfd_send_signal(int pidfd, int signal) {
+    return syscall(SYS_pidfd_send_signal, pidfd, signal, NULL, 0);
+}
+
 // SYS_clone3 is only defined on Linux Kernel 5.3 and above
 // Define our dummy value if it's not available (as is the case with Musl libc)
 #ifndef SYS_clone3
@@ -344,6 +334,18 @@ static int _clone3(int *pidfd) {
     return syscall(SYS_clone3, &args, sizeof(args));
 }
 
+struct linux_dirent64 {
+    unsigned long d_ino;
+    unsigned long d_off;
+    unsigned short d_reclen;
+    unsigned char d_type;
+    char d_name[];
+};
+
+static int _getdents64(int fd, struct linux_dirent64 *dirp, size_t nbytes) {
+    return syscall(SYS_getdents64, fd, dirp, nbytes);
+}
+
 static pthread_mutex_t _subprocess_fork_lock = PTHREAD_MUTEX_INITIALIZER;
 
 static int _subprocess_make_critical_mask(sigset_t *old_mask) {
@@ -371,10 +373,18 @@ static int _subprocess_make_critical_mask(sigset_t *old_mask) {
     } \
 } while(0)
 
-#if __DARWIN_NSIG
-#  define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
-#else
-#  define _SUBPROCESS_SIG_MAX 32
+#if defined(NSIG_MAX)           /* POSIX issue 8 */
+# define _SUBPROCESS_SIG_MAX NSIG_MAX
+#elif defined(__DARWIN_NSIG)    /* Darwin */
+# define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
+#elif defined(_SIG_MAXSIG)      /* FreeBSD */
+# define _SUBPROCESS_SIG_MAX _SIG_MAXSIG
+#elif defined(_SIGMAX)          /* QNX */
+# define _SUBPROCESS_SIG_MAX (_SIGMAX + 1)
+#elif defined(NSIG)             /* 99% of everything else */
+# define _SUBPROCESS_SIG_MAX NSIG
+#else                           /* Last resort */
+# define _SUBPROCESS_SIG_MAX (sizeof(sigset_t) * CHAR_BIT + 1)
 #endif
 
 int _shims_snprintf(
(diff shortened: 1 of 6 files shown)


Sources/_SubprocessCShims/process_shims.c
```diff
@@ -620,40 +562,46 @@ int _subprocess_fork_exec(
...
-        }
-        if (file_descriptors[5] >= 0) {
-            rc = close(file_descriptors[5]);
+        if (rc < 0) {
+            write_error_and_exit;
         }
+        // Close all other file descriptors
+        rc = -1;
+        errno = ENOSYS;
+#if __has_include(<linux/close_range.h>)
```
jakepetroules: Here we should add `|| defined(__FreeBSD__)` (FreeBSD also has close_range, defined in unistd.h).

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -364,93 +370,57 @@ static int _subprocess_addchdir_np(
...
+    if (hi < 0) {
+        hi = getdtablesize();
+    }
+#elif defined(__linux__)
+    int hi = _highest_possibly_open_fd_dir("/proc/self/fd");
+    if (hi < 0) {
+        hi = getdtablesize();
+    }
+#else
+    int hi = 1024;
```
jakepetroules: Where does '1024' come from and how safe is it to default to this vs making this #error and requiring a specific implementation for new platforms?

Also, the following additional platforms support getdtablesize: FreeBSD, OpenBSD, NetBSD, Solaris, QNX.

lin72h:
FreeBSD [PR #1698](https://github.com/freebsd/freebsd-src/pull/1698) adds support for POSIX O_CLOFORK, which could simplify this implementation.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Instead of `#error`, how about if I just call `getdtablesize` unconditionally here and have it naturally error on platforms that don't support it.

iCharlesHu:
> FreeBSD [PR #1698](https://github.com/freebsd/freebsd-src/pull/1698) adds support for POSIX O_CLOFORK, which could simplify this implementation.

Ahh thanks. Good to know.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,100 @@ int _subprocess_spawn(
...
-    return 1;
-#else
-    // Standardized posix_spawn_file_actions_addchdir version (POSIX.1-2024, June 2024) available in:
-    //  - Solaris 11.4 (August 2018)
-    //  - NetBSD 10.0 (March 2024)
-    return 1;
-#endif
+static pthread_mutex_t _subprocess_fork_lock = PTHREAD_MUTEX_INITIALIZER;
+
+static int _subprocess_block_everything_but_something_went_seriously_wrong_signals(sigset_t *old_mask) {
```
cthielen: `_subprocess_make_critical_mask` ? `_subprocess_critical_mask` ?

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,100 @@ int _subprocess_spawn(
...
-    }
-    // Setup working directory
-    if (working_directory != NULL) {
-        rc = _subprocess_addchdir_np(&file_actions, working_directory);
-        if (rc != 0) {
-            return rc;
+    return snprintf(str, len, format, str1, str2);
+}
+
+static int _positive_int_parse(const char *str) {
```
cthielen, line 400: I may be misunderstanding but would it be better to do:

```c
static int _positive_int_parse(const char *str) {
    int out = atoi(str);
    return out >= 0 ? out : -1;
}
```

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,100 @@ int _subprocess_spawn(
...
 
-static int _subprocess_posix_spawn_fallback(
-    pid_t * _Nonnull pid,
-    const char * _Nonnull exec_path,
-    const char * _Nullable working_directory,
-    const int file_descriptors[_Nonnull],
-    char * _Nullable const args[_Nonnull],
-    char * _Nullable const env[_Nullable],
-    gid_t * _Nullable process_group_id
+int _shims_snprintf(
```
parkera, line 390: Why do we need this shim?

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,100 @@ int _subprocess_spawn(
...
-    // Flags
-    short flags = POSIX_SPAWN_SETSIGMASK | POSIX_SPAWN_SETSIGDEF;
-    if (process_group_id != NULL) {
-        flags |= POSIX_SPAWN_SETPGROUP;
-        rc = posix_spawnattr_setpgroup(&spawn_attr, *process_group_id);
-        if (rc != 0) { return rc; }
+    struct dirent *dir_entry = NULL;
+    while ((dir_entry = readdir(dir_ptr)) != NULL) {
+        char *entry_name = dir_entry->d_name;
+        int number = _positive_int_parse(entry_name);
```
parkera: Is this really the right way to do this? Seems very ad-hoc.

parkera: (and racy)

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -968,6 +968,35 @@ extension SubprocessUnixTests {
...
+    @Test func testDoesNotInheritRandomFileDescriptorsByDefault() async throws {
+        // This tests makes sure POSIX_SPAWN_CLOEXEC_DEFAULT works on all platforms
+        let pipe = try FileDescriptor.ssp_pipe()
+        defer {
+            close(pipe.readEnd.rawValue)
+            close(pipe.writeEnd.rawValue)
+        }
+        let writeFd = pipe.writeEnd.rawValue
+        let result = try await Subprocess.run(
+            .path("/bin/bash"),
```
cthielen: I _think_ it's safer to assume a Unix system has `/bin/sh` instead of `/bin/bash`.

jakepetroules: Agree. FWIW, FreeBSD doesn't have bash installed by default, and it's located at /usr/local/bin/bash.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: On Linux this function isn't exposed via `import Glibc`

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Yes. So there are generally two ways to determine the highest fd number on Linux:
1) Use `getrlimit`
2) Read `/proc/self/fd`
The difference between the two is that 1) only tells you the limit whereas 2) tells you the actual highest fd opened. I decided to go for 2) because it's more accurate. Also since this code is executed between `fork()` and `exec()`, there is no threading issues here.

weissi: > (and racy)

It's not racy because you're in a new process with just one thread. It's not ideal because it uses `malloc` and in theory the malloc lock might be in a bad state because it's after fork

weissi: And yes, the problem with the limit is that the limit is often _VERY_ high (i.e. 1M) which means you need to issue A LOT of closes

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,99 @@ int _subprocess_spawn(
...
+} while(0)
+
+#if __DARWIN_NSIG
+#  define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
 #else
-    // Standardized posix_spawn_file_actions_addchdir version (POSIX.1-2024, June 2024) available in:
-    //  - Solaris 11.4 (August 2018)
-    //  - NetBSD 10.0 (March 2024)
-    return posix_spawn_file_actions_addchdir(file_actions, path);
+#  define _SUBPROCESS_SIG_MAX 32
```
jakepetroules: POSIX 2024 added NSIG_MAX:

> {NSIG_MAX}
Some historical implementations provided compile-time constants NSIG or SIGMAX to define the maximum number of signals the implementation supported, but these values did not necessarily reflect the number of signals that could be handled using a sigset_t. With the addition of real-time signals and the desire by some applications to be able to allocate additional real-time signals at run-time, neither of these constants provided a useable, portable value. {NSIG_MAX} was added to the standard to allow applications to determine the maximum number of signals that an implementation will support based on the size of the sigset_t type (defined in [<signal.h>](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/signal.h.html)).

32 for Darwin, 64 for Linux, 128 for FreeBSD.

Here is what Ruby's vm_core.h does, I suggest we replicate that for maximum portability:

```c
#if defined(NSIG_MAX)           /* POSIX issue 8 */
# undef NSIG
# define NSIG NSIG_MAX
#elif defined(_SIG_MAXSIG)      /* FreeBSD */
# undef NSIG
# define NSIG _SIG_MAXSIG
#elif defined(_SIGMAX)          /* QNX */
# define NSIG (_SIGMAX + 1)
#elif defined(NSIG)             /* 99% of everything else */
# /* take it */
#else                           /* Last resort */
# define NSIG (sizeof(sigset_t) * CHAR_BIT + 1)
#endif
```
https://github.com/ruby/ruby/blob/c584cc079eef2f4e314a97eff310c9947e1d7010/vm_core.h#L144-L156

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,99 @@ int _subprocess_spawn(
...
-    posix_spawn_file_actions_destroy(&file_actions);
-    posix_spawnattr_destroy(&spawn_attr);
-    return rc;
+#elif defined(__linux__)
+    int hi = _highest_possibly_open_fd_dir("/proc/self/fd");
+    if (hi < 0) {
+        hi = getdtablesize();
+    }
+#else
+    int hi = getdtablesize();
```
jakepetroules: Linux man page recommends that portable applications call sysconf(_SC_OPEN_MAX) instead (which is defined by POSIX).

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,99 @@ int _subprocess_spawn(
...
-    if (rc != 0) { return rc; }
-    rc = posix_spawnattr_setsigdefault(&spawn_attr, &all_signals);
-    if (rc != 0) { return rc; }
-    // Flags
-    short flags = POSIX_SPAWN_SETSIGMASK | POSIX_SPAWN_SETSIGDEF;
-    if (process_group_id != NULL) {
-        flags |= POSIX_SPAWN_SETPGROUP;
-        rc = posix_spawnattr_setpgroup(&spawn_attr, *process_group_id);
-        if (rc != 0) { return rc; }
+static int _highest_possibly_open_fd(void) {
```
jakepetroules, line 465: Why make this platform-conditional? On Linux, /dev/fd exists as symlinks, is there any reason not to always use /dev/fd? That should work on pretty much all platforms.

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -969,6 +969,35 @@ extension SubprocessUnixTests {
...
+            close(pipe.writeEnd.rawValue)
+        }
+        let writeFd = pipe.writeEnd.rawValue
+        let result = try await Subprocess.run(
+            .path("/bin/sh"),
+            arguments: ["-c", "echo hello from child >&\(writeFd); echo wrote into \(writeFd), echo exit code $?"],
+            output: .string,
+            error: .string
+        )
+        close(pipe.writeEnd.rawValue)
```
jakepetroules: This will end up closing writeEnd twice because of the close in the defer block above, which looks unsafe. And I can't see how that behavior is relevant to the test. Instead use the `closeAfter` API from FileDescriptor. It doesn't provide an `async` overload but here's a polyfill I wrote:

```swift
extension FileDescriptor {
    /// Runs a closure and then closes the FileDescriptor, even if an error occurs.
    ///
    /// - Parameter body: The closure to run.
    ///   If the closure throws an error,
    ///   this method closes the file descriptor before it rethrows that error.
    ///
    /// - Returns: The value returned by the closure.
    ///
    /// If `body` throws an error
    /// or an error occurs while closing the file descriptor,
    /// this method rethrows that error.
    public func closeAfter<R>(_ body: () async throws -> R) async throws -> R {
        // No underscore helper, since the closure's throw isn't necessarily typed.
        let result: R
        do {
            result = try await body()
        } catch {
            _ = try? self.close() // Squash close error and throw closure's
            throw error
        }
        try self.close()
        return result
    }
}
```

jakepetroules (changes requested):

lin72h:
Just a heads up - the PR has [merged](https://github.com/freebsd/freebsd-src/commit/02ffca404e65b2058720e1e9d5a5bc8bb2867113). I think the timing means it'll likely make it into the upcoming FreeBSD 15 release in December.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,154 +265,99 @@ int _subprocess_spawn(
...
+    if (end == str) {
+        // No digits found
+        return -1;
     }
-    if (file_descriptors[2] >= 0) {
-        rc = posix_spawn_file_actions_adddup2(
-            &file_actions, file_descriptors[2], STDOUT_FILENO
-        );
-        if (rc != 0) { return rc; }
+    if (errno == ERANGE || val <= 0 || val > INT_MAX) {
```
itingliu: I think this is what's causing the Linux CI failure

```
[7/22] Compiling CSystem shims.c
/__w/swift-subprocess/swift-subprocess/Sources/_SubprocessCShims/process_shims.c:318:28: error: use of undeclared identifier 'val'
  318 |     if (errno == ERANGE || val <= 0 || val > INT_MAX) {
```

Should be `value` instead?

Sources/_SubprocessCShims/process_shims.c
weissi: @iCharlesHu I implemented a version of this in `AsyncProcess` which should be (v)fork safe because it doesn't allocate: https://github.com/swiftlang/swift-sdk-generator/blob/3f4a6a800414010da0057985f80b2aba5dd4fc60/Sources/CProcessSpawnSync/internal-helpers.h#L57-L98

(glibc (LGPL!) also has an implementation: https://github.com/bminor/glibc/blob/6afece738c2b8408585272a95090ce5d5345dd19/sysdeps/unix/sysv/linux/closefrom_fallback.c#L31)

ricardobranco777:
Now that O_CLOFORK is merged in all *BSD (OpenBSD is pending), work can be started on support O_CLOFORK flags.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Thanks @weissi!

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -351,6 +319,140 @@ static int _clone3(int *pidfd) {
...
+#define _subprocess_precondition(__cond) do { \
+    int eval = (__cond); \
+    if (!eval) { \
+        __builtin_trap(); \
+    } \
+} while(0)
+
+#if __DARWIN_NSIG               /* Darwin */
+#  define _SUBPROCESS_SIG_MAX __DARWIN_NSIG
+#elif defined(NSIG_MAX)         /* POSIX issue 8 */
```
jakepetroules: super nit: the POSIX one should go first, being the standard thing, shouldn't it?

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -351,6 +319,140 @@ static int _clone3(int *pidfd) {
...
+    }
+
+error:
+    close(dir_fd);
+    return highest_fd_so_far;
+}
+
+static int _highest_possibly_open_fd(void) {
+#if defined(__linux__)
+    int hi = _highest_possibly_open_fd_dir_linux("/dev/fd");
```
jakepetroules, line 467: nit: might be worth a comment that this is only needed for the Linux-older-than-5.9 path where we don't have close_range which makes this more efficient.

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -952,9 +952,64 @@ extension SubprocessUnixTests {
...
+        }
+
+        #expect(result.terminationStatus.isSuccess)
+        #expect(
+            result.standardOutput?.trimmingCharacters(in: .whitespacesAndNewlines) ==
+            "wrote into \(writeFd), echo exit code 1"
+        )
+        // Depending on the platform, standard error should be something like
+        // `/bin/bash: 7: Bad file descriptor
+        #expect(!result.standardError!.isEmpty)
```
jakepetroules: nit: avoid force unwraps:

```suggestion
        let standardError = try #require(result.standardError)
        #expect(!standardError.isEmpty)
```

jakepetroules (approved):

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -530,20 +632,22 @@ int _subprocess_fork_exec(
...
-        if (file_descriptors[5] >= 0) {
-            rc = close(file_descriptors[5]);
-        }
-
-        if (rc < 0) {
-            write_error_and_exit;
+        // Close all other file descriptors
+        rc = -1;
+        errno = ENOSYS;
+#if __has_include(<linux/close_range.h>) || defined(__FreeBSD__)
```
jakepetroules: Sorry, one more thing. This breaks the build on Android because it doesn't have close_range.

/Volumes/Sources/OpenSource/swift-subprocess/Sources/_SubprocessCShims/process_shims.c:640:14: error: call to undeclared function 'close_range'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  640 |         rc = close_range(STDERR_FILENO + 1, pipefd[1] - 1, 0);
      |              ^
1 error generated.
[8/22] Compiling CSystem shims.c

It was introduced in NDK r34 so you need an availability guard...

jakepetroules (changes requested):

Sources/_SubprocessCShims/process_shims.c
jakepetroules: This will fix it:

```suggestion
#if (__has_include(<linux/close_range.h>) && (!defined(__ANDROID__) || __ANDROID_API__  >= 34)) || defined(__FreeBSD__)
```

iCharlesHu: Thanks for the fix!

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -374,6 +334,142 @@ static int _clone3(int *pidfd) {
...
+static int _highest_possibly_open_fd_dir_linux(const char *fd_dir) {
+    int highest_fd_so_far = 0;
+    int dir_fd = open(fd_dir, O_RDONLY);
+    if (dir_fd < 0) {
+        // errno set by `open`.
+        return -1;
+    }
+
+    // Buffer for directory entries - allocated on stack, no heap allocation
+    char buffer[4096] = {0};
```
jakepetroules, line 425: BLOCKING: I think this is missing some error handling, right? If the buffer isn't of sufficient size because there are > 4096 directory entries, _getdents64 should return -1 and set errno to EINVAL, but it seems we'll return 0 from _highest_possibly_open_fd_dir_linux in that case (the initial value of highest_fd_so_far) because the condition of the while loop is checked to be > 0.

I think we need to return -1 in that case so that in _highest_possibly_open_fd we fall back to _SC_OPEN_MAX. As it stands, exceeding the buffer size will result in closing zero file descriptors.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -374,6 +334,142 @@ static int _clone3(int *pidfd) {
...
+                highest_fd_so_far = -1;
+                goto error;
+            }
+        }
+        long offset = 0;
+        while (offset < bytes_read) {
+            struct linux_dirent64 *entry = (struct linux_dirent64 *)(buffer + offset);
+
+            // Skip "." and ".." entries
+            if (entry->d_name[0] != '.') {
```
jakepetroules, line 448: Should this check for "." and ".." exactly? Right now it's a prefix check. Not sure if there can ever be any dotfiles in here.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -556,20 +652,22 @@ int _subprocess_fork_exec(
...
-
-        if (rc < 0) {
-            write_error_and_exit;
+        // Close all other file descriptors
+        rc = -1;
+        errno = ENOSYS;
+        #if (__has_include(<linux/close_range.h>) && (!defined(__ANDROID__) || __ANDROID_API__  >= 34)) || defined(__FreeBSD__)
+        // We must NOT close pipefd[1] for writing errors
+        rc = close_range(STDERR_FILENO + 1, pipefd[1] - 1, 0);
+        rc |= close_range(pipefd[1] + 1, ~0U, 0);
```
jakepetroules, line 665: On platforms with closefrom (OpenBSD is a relevant one), we could also prefer that.

e.g. loop over `3`..<`pipefd[1]` then call `closefrom(pipdfd[1] + 1)`

Also, Linux has this "libbsd" library which provides polyfills of closefrom; might be a useful reference: https://cgit.freedesktop.org/libbsd/tree/src/closefrom.c

Tests/SubprocessTests/LinuxTests.swift
```diff
@@ -64,56 +63,31 @@ struct SubprocessLinuxTests {
...
-                    }
-                } finally: { error in
-                    // Now kill the process
-                    try subprocess.send(signal: error != nil ? .kill : .terminate)
+        ) { subprocess, standardOutput in
+            try await tryFinally {
+                // First suspend the process
+                try subprocess.send(signal: .suspend)
+                try await waitForCondition(timeout: .seconds(30)) {
+                    let state = try subprocess.state()
```
jakepetroules: Why change this? I found it necessary to check for the additional states in order to make the test more reliable.

Looks like CI already had a failure:

✘ Test testSuspendResumeProcess() recorded an issue at LinuxTests.swift:65:6: Caught error: Timed out waiting for condition to be true
✘ Test testSuspendResumeProcess() failed after 30.059 seconds with 1 issue.

jakepetroules (changes requested):

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -556,20 +652,22 @@ int _subprocess_fork_exec(
...
+        #endif
+        if (rc != 0) {
+            // close_range failed (or doesn't exist), fall back to close()
+            for (int fd = STDERR_FILENO + 1; fd < _highest_possibly_open_fd(); fd++) {
```
jakepetroules: BLOCKING: Think you have an off-by-one error here and need this:

```suggestion
            for (int fd = STDERR_FILENO + 1; fd <= _highest_possibly_open_fd(); fd++) {
```

Might explain why testSubprocessDoesNotInheritVeryHighFileDescriptors is failing with one fd still open? Especially because it occurs on amazonlinux2 and focal, both of which have a glibc older than 2.34 (and therefore no close_range).

jakepetroules:
✘ Test testDoesNotReapUnrelatedChildProcess() recorded an issue at ProcessMonitoringTests.swift:229:6: Caught error: Failed to spawn the new process with underlying error: UnderlyingError(rawValue: 11)
✘ Test testDoesNotReapUnrelatedChildProcess() failed after 0.006 seconds with 1 issue.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: No `_getdents64` doesn't work like that. It's akin to `readdir` in the sense that it's an iterator. You keep calling `_getdents64` on the same `dir_fd` with the fixed-sized buffer; `_getdents64` will fill the buffer and return the size read. You keep calling `_getdents64` until it returns 0, which means all entries are read.

https://linux.die.net/man/2/getdents64 has a sample program that uses fixed buffer:
```c
#define BUF_SIZE 1024

int
main(int argc, char *argv[])
{
    int fd, nread;
    char buf[BUF_SIZE];
    struct linux_dirent *d;
    int bpos;
    char d_type;

   fd = open(argc > 1 ? argv[1] : ".", O_RDONLY | O_DIRECTORY);
    if (fd == -1)
        handle_error("open");

   for ( ; ; ) {
        nread = syscall(SYS_getdents, fd, buf, BUF_SIZE);
        if (nread == -1)
            handle_error("getdents");

       if (nread == 0)
            break;

       printf("--------------- nread=%d ---------------\n", nread);
        printf("i-node#  file type  d_reclen  d_off   d_name\n");
        for (bpos = 0; bpos < nread;) {
            d = (struct linux_dirent *) (buf + bpos);
            printf("%8ld  ", d->d_ino);
            d_type = *(buf + bpos + d->d_reclen - 1);
            printf("%-10s ", (d_type == DT_REG) ?  "regular" :
                             (d_type == DT_DIR) ?  "directory" :
                             (d_type == DT_FIFO) ? "FIFO" :
                             (d_type == DT_SOCK) ? "socket" :
                             (d_type == DT_LNK) ?  "symlink" :
                             (d_type == DT_BLK) ?  "block dev" :
                             (d_type == DT_CHR) ?  "char dev" : "???");
            printf("%4d %10lld  %s\n", d->d_reclen,
                    (long long) d->d_off, d->d_name);
            bpos += d->d_reclen;
        }
    }

   exit(EXIT_SUCCESS);
}
```

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: I don't think that's necessary because we know this code is only called on `/dev/fd`, which only has fd numbers. The only other dot file would be `.` and `..`.

Tests/SubprocessTests/LinuxTests.swift
iCharlesHu: I'm updating this entire test. I don't think we should poll `/proc/self`, which is inherently racy. Instead, we should just call `waitpid` like the Darwin one.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: When will `close_range` be not available but `closefrom` is? https://man.freebsd.org/cgi/man.cgi?closefrom(2) seems to imply they have the same availability.

Sources/_SubprocessCShims/process_shims.c
jakepetroules: Got it, thanks for clarifying RE the buffer size. However I think is still some missing error handling: if _getdents64 returns -1, the loop terminates but returns the highest_fd_so_far.

This block should be _outside_ the while loop:

```c
if (bytes_read < 0) {
    if (errno == EINTR) {
        continue;
    } else {
        // `errno` set by _getdents64.
        highest_fd_so_far = -1;
        goto error;
    }
}
```

The condition of the while loop is that bytes_read>0, so bytes_read<0 can never be true inside the loop.

Sources/_SubprocessCShims/process_shims.c
jakepetroules: FreeBSD is the only platform that has _both_ functions. Here, this table should clarify:

| Operating System | `close_range` Function | `closefrom` Function |
|------------------|-----------------------|---------------------|
| macOS            | No                    | No                 |
| FreeBSD          | [Yes](https://man.freebsd.org/cgi/man.cgi?close_range(2))                    | [Yes](https://man.freebsd.org/cgi/man.cgi?closefrom(2)) (returns `void`)                |
| OpenBSD          | No                    | [Yes](https://man.openbsd.org/closefrom) (returns `int`)                |
| Linux/Android    | [Yes](https://man7.org/linux/man-pages/man2/close_range.2.html)                   | No                  |

I meant to say that I think we should use close_range and/or closefrom whenever possible since it minimizes the number of syscalls needed to close all the fds. Specifically, we should prefer closefrom over looping, not that we should prefer closefrom over close_range. Thus closefrom instead of looping would be an OpenBSD-specific optimization.

ricardobranco777: I'm working on enabling `close_range` on NetBSD.  Will also do DragonflyBSD.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Ahh good catch. Thanks!

Sources/_SubprocessCShims/process_shims.c
jakepetroules: Nice to hear! Would be great if OpenBSD could get it too.

closefrom is not so nice as it has incompatible signatures on FreeBSD vs OpenBSD.

PR #81, opened 2025-06-14
withAsyncTaskCleanupHandler calls its cleanup handler twice if body throws

### LINKED ISSUE #80: withAsyncTaskCleanupHandler calls its cleanup handler twice if body throws
If the body closure given to withAsyncTaskCleanupHandler throws, its cleanup handler is called twice: once in the catch block where body is called, and then again in the task group task once the Task.sleep throws due to cancellation, which swallows the error and then continues to call the handler as well.

This results in the teardown sequence being invoked twice.

jakepetroules:
If the body closure given to withAsyncTaskCleanupHandler throws, its cleanup handler is called twice: once in the catch block where body is called, and then again in the task group task once the Task.sleep throws due to cancellation, which swallows the error and then continues to call the handler as well.

This results in the teardown sequence being invoked twice.

Closes #80


Sources/Subprocess/Configuration.swift
```diff
@@ -752,31 +752,156 @@ extension Optional where Wrapped == String {
...
 
         do {
             let result = try await body()
-            group.cancelAll()
+            runCancellationHandlerPromise.fulfill(with: false)
             return result
         } catch {
-            await handler()
+            runCancellationHandlerPromise.fulfill(with: true)
             throw error
```
iCharlesHu, line 790: It seems like the issue is simply we don't have to call `handler()` here and let the one above run. Why do we need `Promise`?

On a related note, we are really trying very very hard to move away from `LockedState` like types like this at least on Darwin (this is why we decided to move `.standardOutput` and `.standardError` out from the `Execution` type, so we don't have to hold a `LockedState`).

jakepetroules: I cannot see any other way to ensure the cancellation handler runs only once and only on cancellation or failure. The group.addTask is ALWAYS going to run to completion no matter what happens, but we specifically don't want to run the teardown sequence if no cancellation was invoked or error was thrown. Otherwise you may incorrectly attempt to terminate a still-running process after the body closure runs.

jakepetroules:
Rolled into #88

PR #83, opened 2025-06-15
Process termination monitoring implementation on Linux conflicts with processes spawned by other means

### LINKED ISSUE #82: Process termination monitoring implementation on Linux conflicts with processes spawned by other means
The approach used by Subprocess to monitor subprocess termination on Linux is fundamentally flawed as it calls waitid with P_ALL, and WEXITED without WNOWAIT, which will end up reaping pids that were spawned outside the Subprocess library.

We need to use an implementation more like https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/ExitTests/WaitFor.swift, which doesn't suffer from this issue.

jakepetroules:
The approach used by Subprocess to monitor subprocess termination on Linux is fundamentally flawed as it calls waitid with P_ALL, and WEXITED without WNOWAIT, which will end up reaping pids that were spawned outside the Subprocess library.

Use an implementation more like swift-testing does for wait tests, which doesn't suffer from this issue.

Closes #82

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Linux.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Linux.swift
@@ -37,6 +37,7 @@ extension Configuration {
         outputPipe: consuming CreatedPipe,
         errorPipe: consuming CreatedPipe
     ) throws -> SpawnResult {
+        // Ensure the waiter thread is running.
         _setupMonitorSignalHandler()
 
         // Instead of checking if every possible executable path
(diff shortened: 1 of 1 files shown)


Sources/Subprocess/Locked.swift
```diff
@@ -0,0 +1,138 @@
+//===----------------------------------------------------------------------===//
```
iCharlesHu: We really would rather not keep the `Lockable` (aka `LockedState`) pattern. Since this is `Linux` only change, we should just use `Mutex`.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -316,67 +313,71 @@ private extension siginfo_t {
 }
 
 private let setup: () = {
-    signalSource.setEventHandler {
-        while true {
-            var siginfo = siginfo_t()
-            guard waitid(P_ALL, id_t(0), &siginfo, WEXITED) == 0 || errno == EINTR else {
```
iCharlesHu, line 322: Can't we simply add `WEXITED | WNOWAIT` here? swift-testing's Linux implementation also uses
```
waitid(P_ALL, 0, &siginfo, WEXITED | WNOWAIT)
```

jakepetroules: Explained in DM: you need to first call with WNOWAIT to be notified that the process has exited, then you call again with the specific pid and without WNOWAIT to “release” the pid back to the system for reuse. Without that second waitid call it’s a “leak” where you have a bunch of zombie subprocesses lying around.

Sources/Subprocess/Locked.swift
jakepetroules: I cannot use Mutex because I need the semantics of pthread_cond_signal and pthread_cond_wait, the latter of which requires a pthread_mutex_t.

Mutex's underlying implementation on Linux is a futex, and you can't access the underlying lock instance via the Mutex API anyways.

So I don't see another way to do this, unless you want me to just inline the code. But that would just mean directly calling pthread_mutex_init/pthread_mutex_lock/pthread_mutex_unlock, and still not using Mutex.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -266,32 +266,56 @@ extension String {
 internal func monitorProcessTermination(
     forProcessWithIdentifier pid: ProcessIdentifier
 ) async throws -> TerminationStatus {
+    // Ensure the waiter thread is running.
+    _setupMonitorSignalHandler()
```
iCharlesHu: This method is already called in `Configuration.spawn`, we don't need to call it again here.

jakepetroules: Done

PR #87, opened 2025-06-17
Use of Execution and teardown sequences race with process termination

jakepetroules:
Process termination monitoring is started asynchronously with the closure which receives the Execution, meaning any use of Execution in the body to send signals, etc., may send them to a pid which no longer refers to the original process. The cleanup handler may also run after the process has already terminated. This compounds with the fix for #82 (#83) to fully resolve the race conditions.


jakepetroules:
This needs #83 to merge first to avoid introducing a _different_ PID reuse issue.

jakepetroules:
Rolled into #88

PR #90, opened 2025-06-23
Ensure monitorProcessTermination is ALWAYS called after body runs

### LINKED ISSUE #89: Subprocess.run can hang if body throws
This is a regression introduced by #88. We need to ensure monitorProcessTermination is ALWAYS called after body runs, regardless of whether it threw an error or not.

jakepetroules:
This fixes a regression introduced by #88. We need to ensure monitorProcessTermination is ALWAYS called after body runs, regardless of whether it threw an error or not.

Closes #89

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -78,25 +78,34 @@ public struct Configuration: Sendable {
 
         let processIdentifier = _spawnResult.execution.processIdentifier
 
-        let result = try await withAsyncTaskCleanupHandler {
-            let inputIO = _spawnResult.inputWriteEnd()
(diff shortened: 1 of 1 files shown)


Sources/Subprocess/Configuration.swift
```diff
@@ -779,15 +788,15 @@ internal func withAsyncTaskCleanupHandler<Result>(
             // so _immediately_ if the failure scenario is due to parent task
             // cancellation. We do so in a detached Task to prevent cancellation
             // of the parent task from interrupting enumeration of the stream itself.
-            await Task.detached {
+            await withUncancelledTask {
```
jakepetroules, line 791: Already had an abstraction for this

PR #91, opened 2025-06-23
Fix issues with AsyncBufferSequence.LineSequence

### LINKED ISSUE #78: testLineSequence is flaky, at least on Linux
testLineSequence occasionally fails on Linux. See attached test output which was from a run with nightly-main jammy on GitHub actions CI.

[job-logs.txt](https://github.com/user-attachments/files/20732129/job-logs.txt)

jakepetroules:
The main problem is that the internal buffer used by AsyncBufferSequence.LineSequence may end on the boundary of a line ending sequence -- this impacts primarily the UTF-8 encoding, but also impacts UTF-16 and UTF-32 with the \r\n sequence specifically. When this condition occurs, the range check prevents the peek-ahead to the next 1 or 2 bytes and prevents a complete line from being returned to the client.

The buffer size is based on the page size by default, which on my macOS and Linux systems respectively are 16384 and 4096. This led to testLineSequence failing frequently on Linux due to the greater likelihood of a line ending being split across a multiple of the buffer size.

To fix this, we always load more bytes into the buffer until the buffer no longer ends with a potential partial line ending sequence (unless we hit EOF which correctly causes an early return).

Additionally, the testLineSequence test could generate empty lines, which meant that it was possible to have a line ending with \r followed by an (empty) line ending with \n, indistinguishable from a single line ending with \r\n. This is a problem for any line ending sequences where one is a prefix of another -- \r and \r\n are the only ones which meet this criteria. To fix that, prevent the test from ever generating an empty buffer, and since it does so by restricting the character set to Latin, will never again produce the problematic sequence.

Also switch testTeardownSequence to use AsyncBufferSequence.LineSequence instead of its custom line splitting logic. This ensures the test works correctly regardless of buffer size, even with a contrived buffer size of 1.

Closes #78


jakepetroules:
To test this, I hardcoded the buffer size to 1 to help ensure it was working correctly including in cases where it needed to read from the underlying buffer multiple times in a row for the UTF-8 line separator and paragraph separator sequences.

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -881,17 +874,21 @@ extension SubprocessUnixTests {
...
+                length = Int.random(in: 1 ..< 16)
             }
 
             var buffer: [UInt8] = Array(repeating: 0, count: length)
             for index in 0 ..< length {
                 buffer[index] = UInt8.random(in: range)
             }
+            // Buffer cannot be empty or a line with a \r ending followed by an empty one with a \n ending would be indistinguishable.
+            // This matters for any line ending sequences where one line ending sequence is the prefix of another. \r and \r\n are the
+            // only two which meet this criteria.
```
itingliu, line 890: Are you saying this could result in a paragraph that is
```
\r
\n
```
which is no different from
```
\r\n
```
?

And the problem with this is that the test assumes two lines, so it breaks the test assumption?

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -214,14 +214,35 @@ extension AsyncBufferSequence {
...
+                                return true
+                            default:
+                                return false
+                            }
+                        }
+                    }
                     // Step 1: Load more buffer if needed
-                    if self.startIndex >= self.buffer.count {
+                    var partialLineSequence = bufferHasPotentialPartialLineSequence()
+                    while partialLineSequence || self.startIndex >= self.buffer.count {
```
itingliu: If the line sequence is just `...text...\r... text...\r` where there is never a `\n`, does this mean we'd just read in all buffered data at once (since `partialLineSequence` will keep being `true`)? 

Would it work to just unconditionally read one more byte to capture the potential `\n`, regardless of the set buffer size?

Tests/SubprocessTests/SubprocessTests+Unix.swift
jakepetroules: Right, consider the following two lines:

```swift
"xxx\r"
"\n"
```

That is indistinguishable from the single line:

```swift
"xxx\r\n"
```

since the parser will always greedily look for a subsequent \n if \r is seen. And both are valid; \r is also a valid newline sequence on its own.

Sources/Subprocess/AsyncBufferSequence.swift
jakepetroules: I ended up restructuring the code to make it much easier to follow and validate.

jakepetroules:
Following some feedback from @itingliu, I ended up restructuring the code to make it much easier to follow and validate. The previous iteration of this patch also still failed in rare cases, but the latest version I've had running the test in a loop for probably 12 hours straight with no issues, so this should be quite solid now.

owenv (approved):

PR #93, opened 2025-06-24
Store the Windows process HANDLE in Execution to avoid pid reuse races

### LINKED ISSUE #92: ProcessIdentifier should store the HANDLE internally on Windows, not just the numeric PID
Unlike Unix, Windows doesn't have a "zombie" state where the pid remains valid in a zombie state after the subprocess has exited. The numeric process ID becomes invalid immediately once the process has exited.

On Windows, instead of closing/discarding the process HANDLE immediately after CreateProcess returns, it should be retained in ProcessIdentifier and only closed once monitorProcessTermination returns. Otherwise, the period between which the original HANDLE returned from CreateProcessW is closed and OpenProcess is subsequently called to regain the handle from the system using the numeric pid is subject to race conditions.

jakepetroules:
Unlike Unix, Windows doesn't have a "zombie" state where the numeric pid remains valid after the subprocess has exited. Instead, the numeric process ID becomes invalid immediately once the process has exited. Per the Windows documentation of GetProcessId and related APIs, "Until a process terminates, its process identifier uniquely identifies it on the system."

On Windows, instead of closing/discarding the process HANDLE immediately after CreateProcess returns, retain the HANDLE in Execution and only close it once monitorProcessTermination returns. This resolve a race condition where the process could have exited between the period where CloseProcess was called and monitorProcessTermination is called. Now we simply use the original process handle to wait for the process to exit and retrieve its exit code, rather than "reverse engineering" its HANDLE from its numeric pid using OpenProcess.

Closes #92


jakepetroules:
@compnerd Any thoughts on this or the associated GitHub issue?

compnerd (commented):
This feels a lot more safe to me. I think that the execution needs to hold on to the process handle. Once the execution is torn down, we can close the handle.

jakepetroules:
> This feels a lot more safe to me. I think that the execution needs to hold on to the process handle. Once the execution is torn down, we can close the handle.

Yep, that's what it's doing. `monitorProcessTermination` calls CloseHandle once the process is terminated and we've collected its output.

Sources/Subprocess/Execution.swift
```diff
@@ -35,13 +35,16 @@ public struct Execution: Sendable {
     public let processIdentifier: ProcessIdentifier
 
     #if os(Windows)
+    internal nonisolated(unsafe) let processHandle: HANDLE
```
iCharlesHu: I agree we should save the process HANDLE but I don't think we should put it on Execution. Why not leave it in `ProcessIdentifier` since it serves that purpose? Therefore you don't have to make other changes.

iCharlesHu: I understand you used `nonisolated(unsafe)` here because `HANDLE` isn't `Sendable`, and we want `Execution` and `ProcessIdentifier` to be. @compnerd  do you know what the best approach is? Should we use `nonisolated(unsafe)` or mark the struct `@unchecked Sendable` (or other approaches)?

jakepetroules: I looked at that, but ProcessIdentifier is Codable and Hashable, so HANDLE shouldn't go in there because it's a pointer (which we should not be serializing to disk)

jakepetroules: HANDLE is an opaque pointer, so it's effectively Sendable in practice in the logical context of how it's used.

jakepetroules: Also, since the HANDLE is only valid for the lifetime of the process and becomes invalid once CloseHandle is called, it seemed a better for to be on Execution.

jakepetroules: Also, you meant @compnerd not @ compend

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -473,15 +445,8 @@ internal func monitorProcessTermination(
...
-    guard
-        let processHandle = OpenProcess(
-            DWORD(PROCESS_QUERY_INFORMATION | SYNCHRONIZE),
-            false,
-            pid.value
-        )
-    else {
-        return .exited(1)
+
+        _ = CloseHandle(execution.processHandle)
```
owenv: My understanding is that a CloseHandle error is _usually_ programmer error, but should this assert/throw if it does return one?

jakepetroules: Good point, I added a fatalError.

owenv (approved):

Issue #85, opened 2025-06-16
`self.spawn` is a blocking function but is run on `async` threads


weissi:
https://github.com/swiftlang/swift-subprocess/blob/80bd50f91abf49164156b61c1071821927242c62/Sources/Subprocess/Configuration.swift#L69

`execve`/`posix_spawn` will block the child process (which is fine) but we also need to communicate information back into the parent process (to figure out if `execve` worked). That however may block, for example if we spawn an executable on a network'd file system. But in the code today, `try self.spawn(...)` is run on `async` threads which is a problem as it might block all asynchronous work from happening potentially indefinitely.

`self.spawn` should be offloaded onto a thread pool where blocking is okay.

jakepetroules:
Unfortunately this will also technically require an API change, since runDetached is currently not async.

jakepetroules:
#95 proposes to remove `runDetached`, so that would also unblock resolving this.

jakepetroules:
Removed API change tag since the removal of runDetached in #95 means this change can now be done without further API changes.

PR #95, opened 2025-06-24
Remove runDetached API

### LINKED ISSUE #94: Design of runDetached API is problematic on Windows
This function returns a ProcessIdentifier, which on Unix is a pid and on Windows is a numeric process identifier. Similar to #92, there is a Windows-specific race condition here. On Unix the pid will be valid until someone waitid's it. But on Windows the ProcessIdentifier may already be invalid by the time runDetached returns.

Should we consider an API change to return a different type like DetachedProcessIdentifier which would provide access to the underlying HANDLE? and would NOT be Codable, unlike ProcessIdentifier. This would still fulfill the API contract that the caller is responsible for freeing the subprocess resources and monitoring any state, but you need the original HANDLE to be able to do that without race conditions.

Marking as blocker because I think the question needs to be resolved before the API becomes immutable.

--

One approach could be to have runDetached return an Execution, and make the processHandle property I'm adding in #93, public instead of internal. That way we don't need to introduce new types.

jakepetroules:
Currently runDetached returns a ProcessIdentifier, which contains a pidfd on Linux and HANDLE on Windows. These handles are closed as soon as runDetached returns, making the values effectively useless to the caller. Similar to https://github.com/swiftlang/swift-subprocess/issues/92, there is a Windows-specific race condition here. On Unix the pid will be valid until someone waitid's it, so a caller _could_ use pidfd_open to get a new pidfd. But on Windows the ProcessIdentifier may already be invalid by the time runDetached returns and therefore OpenProcess can't be used to recover a handle. On FreeBSD one can't get a process descriptor from a PID at all (PDs are only for child processes).

In order to avoid these race conditions, runDetached would need to either return a type which manages the lifetime of the various platform-specific handles, or leak the handles and leave them to be managed by the user. Due to questionable utility of these APIs in the first place (the closure based API can be used in conjunction with a Task to do pretty much everything runDetached can, without the downsides), we simply remove them.

Closes #94


jakepetroules:
@compnerd Any thoughts on this or the associated GitHub issue?

jakepetroules:
I thought about that...

I think we'd need to make Execution non-Copyable and add a deinit that calls CloseHandle? I seem to recall there was previous discussion about doing making Execution non-Copyable, but it didn't happen, maybe @iCharlesHu can provide some info on why that path wasn't chosen at the time?

And then I wonder if we would want some method like "takeHandle" which would allow the client to explicitly take ownership of the HANDLE away from the Execution.

jakepetroules:
> However, it was specifically rejected due to the issue raised by @compnerd: we don’t want to return something and rely on the user to close it. 

Doesn't `~Copyable` + `deinit` allow us to close the handle without requiring the user to do so?

Also, I would argue that returning a numeric pid on Unix is still a resource leak of sorts, since the zombie will remain in the process table until the user calls `waitid` on it. So how's that much different from failing to call CloseHandle on Windows?

jakepetroules:
@iCharlesHu I ended up splitting part of this change into #101 (which also introduces new API but is independently useful of these runDetached-specific issues), and now instead of returning Execution, runDetached returns a new non-Copyable type ProcessHandle which guarantees that the underlying platform handle is released (so there can be no resource leaks) while still solving the underlying race condition since the caller now has access to the platform handles from the return value of runDetached.

Happy to still start a thread on the Swift forums if you think it's needed but hopefully this removes the resource leakage controversy in an elegant way!

iCharlesHu (changes requested):
This API change needs to be discussed.

iCharlesHu (approved):

iCharlesHu:
We decided to remove this API for two reasons:
1) We realized we can't safely offer this API. Subprocesses now operate on process file descriptors (process handles on Windows) that need to be closed;
2) As #85 pointed out, spawning processes (posix_spawn) is really a blocking operation, which means the spawn call that runDetached relies on will have to be async, forcing runDetached to be async as well.

jakepetroules:
Rebased to apply cleanly on main.

PR #99, opened 2025-06-29
Add support for FreeBSD

jakepetroules:
Also includes speculative support for OpenBSD, which I haven't tested, but which is likely correct as it tends to require the same code paths as FreeBSD.


Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -310,11 +316,16 @@ private final class ChildProcessContinuations: Sendable {
 private let _childProcessContinuations = ChildProcessContinuations()
 
 private nonisolated(unsafe) let _waitThreadNoChildrenCondition = {
+    #if os(FreeBSD) || os(OpenBSD)
+    let result = UnsafeMutablePointer<pthread_cond_t?>.allocate(capacity: 1)
```
owenv, line 320: any reason not to use the MutexType type alias here?

owenv (approved):

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -285,8 +285,14 @@ internal func monitorProcessTermination(
 
 // Small helper to provide thread-safe access to the child process to continuations map as well as a condition variable to suspend the calling thread when there are no subprocesses to wait for. Note that Mutex cannot be used here because we need the semantics of pthread_cond_wait, which requires passing the pthread_mutex_t instance as a parameter, something the Mutex API does not provide access to.
 private final class ChildProcessContinuations: Sendable {
+    #if os(FreeBSD) || os(OpenBSD)
+    typealias MutexType = pthread_mutex_t?
```
parkera, line 289: This is unfortunate. Why would we support an optional lock?

michael-yuji: For Linux/Glibc, pthread_mutex_t is a union, but on OpenBSD/FreeBSD, it is a pointer.
Unfortunately in the view of swift this could potentially be a null pointer and hence becomes imported as an optional pointer type.

etcwilde: Long-term, once we have a separate module map and platform overlay, we should API-note it with the appropriate nullability information.
https://github.com/swiftlang/swift/issues/81407

PR #101, opened 2025-06-30
Expose platform-specific process file descriptors

jakepetroules:
Expose the pidfd on Linux and process/thread HANDLEs on Windows. This is necessary for interop with certain platform-native APIs, especially on Windows. pidfds are also gaining the ability to be used for more and more with each Linux kernel release.

One test has been introduced which shows how this API could be used to associate a subprocess with a Job Object using the Windows API, and for Linux, a "test" that shows one use case for access to the raw fd, which is to read the globally unique (to the boot session) identifier for the process, allowing processes sharing the same PID to be distinguished.

FreeBSD also has the concept of process file descriptors, and would expect to expose the same ProcessIdentifier API as for Linux. This may spread to other platforms in the future.

--

Somewhat related article with some good info: https://gaultier.github.io/blog/way_too_many_ways_to_wait_for_a_child_process_with_a_timeout.html

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Linux.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Linux.swift
@@ -191,7 +191,7 @@ extension Configuration {
 public struct ProcessIdentifier: Sendable, Hashable {
     /// The platform specific process identifier value
     public let value: pid_t
-    internal let processDescriptor: PlatformFileDescriptor
+    public let processDescriptor: CInt
 
     internal init(value: pid_t, processDescriptor: PlatformFileDescriptor) {
         self.value = value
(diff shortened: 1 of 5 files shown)


iCharlesHu (changes requested):
This API changes needs to be discussed

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -191,7 +191,7 @@ extension Configuration {
 public struct ProcessIdentifier: Sendable, Hashable {
     /// The platform specific process identifier value
     public let value: pid_t
-    internal let processDescriptor: PlatformFileDescriptor
+    public let processDescriptor: CInt
```
iCharlesHu, line 194: nit: since we are using it else where might as well keep it here as well

jakepetroules: `PlatformFileDescriptor` is internal (just like `ProcessIdentifierProtocol`)

PR #113, opened 2025-07-04
Disable Windows Test

iCharlesHu:
https://github.com/swiftlang/swift-subprocess/pull/104 was unreviewed. Reverting

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -24,6 +24,7 @@ jobs:
           dnf install -y procps
         fi
       windows_swift_versions: '["6.1", "nightly-main"]'
+      windows_build_command: 'swift build'
       enable_macos_checks: true
       macos_xcode_versions: '["16.3"]'
 


jakepetroules (changes requested):
Are you sure the test hangs are not a regression from the changes being introduced in #64? The Windows tests have been consistently passing since #104 was merged, including on all the PRs open for review. Only #64 saw a hang.

If I pull the main branch and run `swift test` on my machine several times in a row, I don't see any hangs. However, if I pull your `iCharlesHu:charles/linux-async-io` branch from #64, the tests hang for me immediately.

I think we should not merge this since the evidence points to #64 being the cause of the issue.

jakepetroules:
Also, in case you're referring to:

> Test / Windows (nightly-main - windows-2019)

...that's not hanging because of a code issue. It's hanging because GitHub removed Windows 2019 builders from their infrastructure, and the repository is misconfigured to still require successful builds of that job. We need @shahmishal to change the requirement from that one to:

> Pull request / Test / Windows (6.1 - windows-2022) (pull_request)

Issue #105, opened 2025-06-30
Zombies won't be reaped if spawn() throws in Configuration.run


jakepetroules:
Under certain circumstances if spawn() throws in Configuration.run, process resources will never be freed (zombies won't be reaped on Unix) if the call to spawn() throws an error after posix_spawn or fork/exec has successfully returned a pid.

This "resource leak" doesn't affect Windows because the only point at which spawnDirect/spawnAsUser can throw after a successful process creation explicitly releases the handles if closing file handles fails, but the process may still continue running.

On Unix (both Darwin and non-Darwin implementations are affected), failing to close file handles after a successful spawn, or failing to access the working directory, can trigger this bug. In this case we should probably SIGKILL the process and perform termination monitoring to wait for exit.

A good way to do this could be to remove Execution from SpawnResult and then change the return type of spawn to something like `Result<(Execution, SpawnResult), (Execution?, Error)>` (where the Execution nullability indicates whether the failure occurred pre- or post- spawn of the actual process) while turning spawn into a non-throwing function.

iCharlesHu:
I believe his issue only affects Linux and only under specific scenarios:

On Darwin, we use `posix_spawn()`. `posix_spawn()` is the last step in `spawn()`, and if it succeeds, `spawn()` will succeed. According to the `posix_spawn()` man page: 

> Upon successful completion, posix_spawn() and posix_spawnp() place
       the PID of the child process in pid and return 0.  If there is an
       error during the fork() step, then no child is created, the
       contents of *pid are unspecified, and these functions return an
       error number as described below.

(see [https://man7.org/linux/man-pages/man3/posix_spawn.3.html](https://man7.org/linux/man-pages/man3/posix_spawn.3.html))

`posix_spawn` is "atomic" in the sense that if it does not return zero (spawn() throws), no child is created, so we won't need to reap.

On Linux: this issue happens if fork() succeeds (child created) but any setup code fails between `fork` and `exec` (including `exec()` itself). I'll add a Linux-specific handler for this.

Issue #110, opened 2025-07-02
Implement IOCP for AsyncIO on Windows


iCharlesHu:
We are introducing `epoll` based `AsyncIO` for `Subprocess` with https://github.com/swiftlang/swift-subprocess/pull/64 . We should do the same for Windows but based it on `IOCP`.

iCharlesHu:
`AsyncIO` was implemented with https://github.com/swiftlang/swift-subprocess/pull/117

Issue #97, opened 2025-06-26
Unify unit tests on all platforms and add more


iCharlesHu:
Currently Subprocess is under tested, especially on Windows. We should make most tests available on all platforms and add more tests in general.

PR #117, opened 2025-07-07
Introduce AsyncIO for Windows and Linux

iCharlesHu:
This PR finishes `AsyncIO` by introducing Windows IO Completion Port based implementation.

This PR is staged on top of https://github.com/swiftlang/swift-subprocess/pull/64 and should be considered ready to be review as soon as that one lands.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/105


iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/110

itingliu:
Talked to @iCharlesHu offline and he confirmed that if we merge this one, we won't need #64

Sources/Subprocess/Configuration.swift
```diff
@@ -581,34 +581,47 @@ internal enum StringOrRawBytes: Sendable, Hashable {
...
+        /// new registration to proceed if the handle is reused.
+        AsyncIO.shared.removeRegistration(for: handle)
+        guard CloseHandle(handle) else {
+            let error = GetLastError()
+            // Getting `ERROR_INVALID_HANDLE` suggests that the file descriptor
+            // might have been closed unexpectedly. This can pose security risks
+            // if another part of the code inadvertently reuses the same file descriptor
+            // number. This problem is especially concerning on Unix systems due to POSIX’s
+            // guarantee of using the lowest available file descriptor number, making reuse
+            // more probable. We use `fatalError` upon receiving `.badFileDescriptor`
```
jakepetroules: nit: this comment still refers to the non-Windows path

Sources/Subprocess/Configuration.swift
```diff
@@ -630,84 +643,248 @@ internal struct TrackedFileDescriptor: ~Copyable {
...
+    mutating func writeFileDescriptor() -> IODescriptor? {
         return self._writeFileDescriptor.take()
     }
 
-    internal init(closeWhenDone: Bool) throws {
-        let pipe = try FileDescriptor.ssp_pipe()
+    internal init(closeWhenDone: Bool, purpose: Purpose) throws {
+        #if canImport(WinSDK)
+        // On Windows, we need to create a named pipe
+        let pipeName = "\\\\.\\pipe\\subprocess-\(purpose)-\(Int.random(in: .min ..< .max))"
```
jakepetroules: I would be a bit worried about running into collisions here, there is a good chance this could repeat and lead to subtle bugs. Instead of making it random, what about using a monotonically increasing unsigned 64-bit integer stored in a static variable? Then we can guarantee this is never a problem.

Something like this maybe?

```swift
func nextPipeNumber() -> UInt64 {
    struct Static: ~Copyable {
        static let pipeNumber = Atomic<UInt64>(0)
    }
    return Static.pipeNumber.add(1, ordering: .sequentiallyConsistent).newValue
}
```

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -673,156 +678,376 @@ final class AsyncIO: Sendable {
...
+            return
+        }
+        PostQueuedCompletionStatus(
+            ioPort,
+            0,
+            shutdownPort,
+            nil
+        )
+        // Wait for monitor thread to exit
+        WaitForSingleObject(monitorThreadHandle, INFINITE);
```
jakepetroules: nit: stray semicolon

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -673,156 +678,376 @@ final class AsyncIO: Sendable {
...
+                let completionKey = UInt64(UInt(bitPattern: handle))
+                // Windows does not offer an API to remove a handle
+                // from given ioCompletionPort. If this handle has already
+                // been registered we simply need to update the continuation
+                let registrationFound = _registration.withLock { storage in
+                    if storage[completionKey] != nil {
+                        // Old registration found. This means this handle has
+                        // already been registered. We simply need to update
+                        // the continuation saved
+                        storage[completionKey] = continuation
```
jakepetroules: Are there scenarios where this is expected, and couldn't this leave continuations that never get resumed? Should this fatalError?

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -673,156 +678,376 @@ final class AsyncIO: Sendable {
...
+        var readLength: Int = 0
+        var signalStream = self.registerHandle(handle).makeAsyncIterator()
+
+        while true {
+            var overlapped = _OVERLAPPED()
+            let succeed = try resultBuffer.withUnsafeMutableBufferPointer { bufferPointer in
+                // Get a pointer to the memory at the specified offset
+                // Windows ReadFile uses DWORD for target count, which means we can only
+                // read up to DWORD (aka UInt32) max.
+                let targetCount = min(bufferPointer.count - readLength, Int(UInt32.max))
```
jakepetroules: Swift also supports x86 (32-bit) Windows, `Int(UInt32.max)` will overflow and crash because `Int` is `Int32` and not `Int64` there.

Sources/Subprocess/IO/AsyncIO.swift
```diff
@@ -673,156 +678,376 @@ final class AsyncIO: Sendable {
...
+        let handle = diskIO.channel
+        var signalStream = self.registerHandle(diskIO.channel).makeAsyncIterator()
+        var writtenLength: Int = 0
+        while true {
+            var overlapped = _OVERLAPPED()
+            let succeed = try span.withUnsafeBytes { ptr in
+                // Windows WriteFile uses DWORD for target count
+                // which means we can only write up to DWORD max
+                let remainingLength = min(
+                    ptr.count - writtenLength, Int(DWORD.max)
```
jakepetroules: `Int(DWORD.max)` will also crash on x86-32. There's a few more occurrences of this later in the file as well.

jakepetroules:
Oh, one more suggestion, could AsyncIO.swift be split up into one file per backend? IOCP, epoll, Dispatch. That would make it easier to add others in future, like kqueue (BSD), poll (non-Linux non-BSD Unix), without the one file getting too big.

Sources/Subprocess/Configuration.swift
iCharlesHu: How about UUID? It feels heavy to introduce lock and atomics just for a random value.

Also: is this really a concern? I would be really worried if a random number between `.min` and `.max` starts to repeat fairly soon.

iCharlesHu: Never mind UUID is not available in Subprocess. I still feel like the chances of us running into collisions are fairly low, and introducing atomic just for this purpose doesn't seem to justify the additional safety it brings.

@cthielen do you have any thoughts on this?

iCharlesHu: I believe random number collisions follows the Birthday Paradox rule, which according to (multiple) ChatGPT/Claude calculations puts us around 3.5 ^ 10 ^9 range for full 64 bit signed int. This means we have to run 3.5 * 10^9 times for it to have more than 50% chance repeating. To me that's statistically negligible.

Sources/Subprocess/IO/AsyncIO.swift
iCharlesHu: No for several reasons:
- Once a handle is registered with we will get notified by `GetQueuedCompletionStatus` to resume continuation (that's the API contract)
- We only have this issue because Windows does not offer "remove from completion queue" API. Normally after each read we would remove the handle from the port so it can be re-registered again. But since we can't, we have to leave this entry in to signify that this handle has already been registered. The second read to the same handle only happens after the first one returns and continuations resumed.

Sources/Subprocess/Configuration.swift
```diff
@@ -589,36 +589,45 @@ internal enum StringOrRawBytes: Sendable, Hashable {
...
 
-    internal mutating func safelyClose() throws {
-        guard self.closeWhenDone else {
-            return
+internal func _safelyClose(_ target: _CloseTarget) throws {
+    switch target {
+    #if canImport(WinSDK)
+    case .handle(let handle):
+        /// Windows does not provide a “deregistration” API (the reverse of
+        /// `CreateIoCompletionPort`) for handles and it it reuses HANDLE
```
cthielen, line 605: Minor: "it it reuses"

cthielen:
```suggestion
        /// `CreateIoCompletionPort`) for handles and it reuses HANDLE
```

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+                }
+
+                // Breakout the monitor loop if we received shutdown from the shutdownFD
+                if targetFileDescriptor == shutdownPort {
+                    break
+                }
+                // Notify the continuations
+                _registration.withLock { store in
+                    if let continuation = store[targetFileDescriptor] {
+                        continuation.yield(bytesTransferred)
```
cthielen: Continuation + lock issue

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+        }
+    }
+
+    private func shutdown() {
+        // Post status to shutdown HANDLE
+        guard case .success(let ioPort) = ioCompletionPort,
+              case .success(let monitorThreadHandle) = monitorThread else {
+            return
+        }
+        PostQueuedCompletionStatus(
```
cthielen, line 168: This could benefit from argument labels if possible.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+    private func registerHandle(_ handle: HANDLE) -> SignalStream {
+        return SignalStream { continuation in
+            switch self.ioCompletionPort {
+            case .success(let ioPort):
+                // Make sure thread setup also succeed
+                if case .failure(let error) = monitorThread {
+                    continuation.finish(throwing: error)
+                    return
+                }
+                let completionKey = UInt64(UInt(bitPattern: handle))
```
cthielen, line 189: Is there an advantage to this over `UInt64(biPattern:)`?

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+        var signalStream = self.registerHandle(handle).makeAsyncIterator()
+
+        while true {
+            var overlapped = _OVERLAPPED()
+            let succeed = try resultBuffer.withUnsafeMutableBufferPointer { bufferPointer in
+                // Get a pointer to the memory at the specified offset
+                // Windows ReadFile uses DWORD for target count, which means we can only
+                // read up to DWORD (aka UInt32) max.
+                let targetCount: DWORD
+                if MemoryLayout<Int>.size == MemoryLayout<Int32>.size {
```
cthielen: Looks like this appears 3x in `AsyncIO+Windows`. Maybe an opportunity for a helper?

Sources/Subprocess/Configuration.swift
jakepetroules: Why is using an Atomic for this "heavy"? (Also, not a lock _and_ atomics -- just an atomic. No need for a lock as well) How do you know Int.random is not actually more expensive? I did a quick benchmark and atomics seem to outperform Int.random by an order of magnitude...

```swift
import Synchronization

let pipeNumber = Atomic<UInt64>(0)

func main() {
    let clock = ContinuousClock()
    let count = 100_000
    let random = clock.measure {
        for _ in 0..<count {
            Int.random(in: .min ..< .max)
        }
    }

    let atomic = clock.measure {
        for _ in 0..<count {
            pipeNumber.add(1, ordering: .sequentiallyConsistent).newValue
        }
    }

    print("random = \(random)")
    print("atomic = \(atomic)")
}

main()
```

random = 0.0332772 seconds (33.2 ms)
atomic = 0.0006226 seconds (0.6 ms)

Sources/Subprocess/IO/AsyncIO.swift
jakepetroules: Thanks for the explanation. I also didn't realize this was an AsyncStream.Continuation when I skimmed over it before, so this makes more sense now.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+                        underlyingError: .init(rawValue: lastError)
+                    )
+                    throw error
+                }
+
+            }
+            // Now wait for read to finish
+            let bytesWritten: DWORD = try await signalStream.next() ?? 0
+
+            writtenLength += Int(bytesWritten)
```
jakepetroules: Here's another place where a DWORD (UInt32) to Int conversion could crash on 32-bit Windows with large-enough values

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -157,7 +152,13 @@ extension AsyncBufferSequence {
                             UnsafeBufferPointer(start: ptr.baseAddress?.assumingMemoryBound(to: Encoding.CodeUnit.self), count: elementCount)
                         )
                     }
-
+                    #else
+                    // Cast data to CodeUnitg type
```
jakepetroules: typo

```suggestion
                    // Cast data to CodeUnit type
```

Sources/Subprocess/Configuration.swift
```diff
@@ -640,84 +649,248 @@ internal struct TrackedFileDescriptor: ~Copyable {
...
+        let shouldClose = self.closeWhenDone
+        self.closeWhenDone = false
+        #if canImport(Darwin)
+        // Transferring out the ownership of fileDescriptor means we don't have go close here
+        let closeFd = self.descriptor
+        let dispatchIO: DispatchIO = DispatchIO(
+            type: .stream,
+            fileDescriptor: self.platformDescriptor(),
+            queue: .global(),
+            cleanupHandler: { error in
```
jakepetroules: I think you made some additional changes since I mentioned this in https://github.com/swiftlang/swift-subprocess/pull/64/files#r2184424762, so I'd suggest adding the `@Sendable` annotation here (since from my reading it looks like it should work now) to avoid any potential build errors if that annotation is corrected at some point.

```suggestion
            cleanupHandler: { @Sendable error in
```

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+                if error != 0 {
+                    continuation.resume(
+                        throwing: SubprocessError(
+                            code: .init(.failedToReadFromSubprocess),
+                            underlyingError: .init(rawValue: error)
+                        )
+                    )
+                    return
+                }
+                if let data = data {
```
jakepetroules: nit

```suggestion
                if let data {
```

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+                    deallocator: .custom(
+                        nil,
+                        {
+                            // noop
+                        }
+                    )
+                )
+            }
+            self.write(dispatchData, to: diskIO) { writtenLength, error in
+                if let error = error {
```
jakepetroules: nit

```suggestion
                if let error {
```

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
```
jakepetroules: nit

```suggestion
                if let error {
```

Sources/Subprocess/IO/Output.swift
```diff
@@ -140,16 +150,14 @@ public struct BytesOutput: OutputProtocol {
...
     ) async throws -> [UInt8] {
-        #if os(Windows)
-        let result = try await diskIO?.fileDescriptor.read(upToLength: self.maxSize) ?? []
-        try diskIO?.safelyClose()
-        return result
-        #else
-        let result = try await diskIO!.dispatchIO.read(upToLength: self.maxSize)
-        try diskIO?.safelyClose()
+        let result = try await AsyncIO.shared.read(from: diskIO, upTo: self.maxSize)
+        try diskIO.safelyClose()
```
jakepetroules, line 167: safelyClose will fail to get called if read throws an error, which is probably going to cause a file descriptor leak? should this use some sort of try-finally abstraction?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+            return
+        }
+
+        // Create thread data
+        let context = MonitorThreadContext(
+            epollFileDescriptor: epollFileDescriptor,
+            shutdownFileDescriptor: shutdownFileDescriptor
+        )
+        let threadContext = Unmanaged.passRetained(context)
+#if os(FreeBSD) || os(OpenBSD)
```
jakepetroules: No need for FreeBSD/OpenBSD checks here since they'd use a kqueue implementation (and a POSIX poll implementation is probably different enough that we'd want its own file rather than ifdef'ing in the Linux one)

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+        guard rc == 0 else {
+            let error = SubprocessError(
+                code: .init(.asyncIOFailed("Failed to create monitor thread")),
+                underlyingError: .init(rawValue: rc)
+            )
+            self.state = .failure(error)
+            return
+        }
+
+#if os(FreeBSD) || os(OpenBSD)
```
jakepetroules: ditto https://github.com/swiftlang/swift-subprocess/pull/117/files#r2227584217

Sources/Subprocess/Configuration.swift
weissi: Agree that a single atomic makes more sense here. Doesn't even need to be `.sequentiallyConsistent` as you're not using it as a memory barrier, `.relaxed` is enough.


---

Also a nit: Ideally this should be

```
enum Static {
    static let number ...
}
```

over

```
struct Static: ~Copyable { ...}
```

because not even planning to ever instantiate a value. `enum`s make better namespaces than `struct`s.

Sources/Subprocess/Buffer.swift
```diff
@@ -92,26 +92,23 @@ extension AsyncBufferSequence.Buffer {
...
     #endif
+    // else Compiler generated conformances
 }
 
 // MARK: - DispatchData.Block
 #if canImport(Darwin) || canImport(Glibc) || canImport(Android) || canImport(Musl)
 extension DispatchData {
     /// Unfortunately `DispatchData.Region` is not available on Linux, hence our own wrapper
-    internal struct _ContiguousBufferView: @unchecked Sendable, RandomAccessCollection {
+    internal struct _ContiguousBufferView: @unchecked Sendable, RandomAccessCollection, Hashable {
```
weissi: this is a very dangerous type. If it really has to exist it should be `~Copyable & ~Escapable` and/or `*Unsafe*`

Sources/Subprocess/Configuration.swift
```diff
@@ -589,36 +589,45 @@ internal enum StringOrRawBytes: Sendable, Hashable {
...
+    switch target {
+    #if canImport(WinSDK)
+    case .handle(let handle):
+        /// Windows does not provide a “deregistration” API (the reverse of
+        /// `CreateIoCompletionPort`) for handles and it it reuses HANDLE
+        /// values once they are closed. Since we rely on the handle value
+        /// as the completion key for `CreateIoCompletionPort`, we should
+        /// remove the registration when the handle is closed to allow
+        /// new registration to proceed if the handle is reused.
+        AsyncIO.shared.removeRegistration(for: handle)
```
weissi, line 612: is this really a synchronous function? CC @al45tair , will this synchronously & without blocking remove a registration?

Sources/Subprocess/Configuration.swift
```diff
@@ -640,84 +649,248 @@ internal struct TrackedFileDescriptor: ~Copyable {
...
+    #else
+    typealias Descriptor = FileDescriptor
+    #endif
+
+    internal var closeWhenDone: Bool
+    internal let descriptor: Descriptor
+
+    internal init(
+        _ descriptor: Descriptor,
+        closeWhenDone: Bool
```
weissi, line 677: we shouldn't have this parameter, anything but `true` here is bad.

For the cases where you need to yank out ownership, you can introduce a `consuming func takeOwnership() -> FileDescriptor`

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+                        throwing: SubprocessError(
+                            code: .init(.failedToReadFromSubprocess),
+                            underlyingError: .init(rawValue: error)
+                        )
+                    )
+                    return
+                }
+                if let data = data {
+                    if buffer.isEmpty {
+                        buffer = data
```
weissi, line 62: threading bug here, you can't capture a mutable `buffer`. Why does the compiler not error here?

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    #if SubprocessSpan
+    internal func write(
+        _ span: borrowing RawSpan,
+        to diskIO: borrowing IOChannel
+    ) async throws -> Int {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Int, any Error>) in
+            let dispatchData = span.withUnsafeBytes {
+                return DispatchData(
+                    bytesNoCopy: $0,
```
weissi, line 86: this is very dangerous because it potentially escapes the pointer that shouldn't escape. This undoes all the benefits that span gives you

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+    #endif  // SubprocessSpan
+
+    internal func write(
+        _ array: [UInt8],
+        to diskIO: borrowing IOChannel
+    ) async throws -> Int {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Int, any Error>) in
+            let dispatchData = array.withUnsafeBytes {
+                return DispatchData(
+                    bytesNoCopy: $0,
```
weissi, line 114: dangerous too, escaping an inner pointer

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+                    if targetFileDescriptor == context.shutdownFileDescriptor {
+                        var buf: UInt64 = 0
+                        _ = _SubprocessCShims.read(context.shutdownFileDescriptor, &buf, MemoryLayout<UInt64>.size)
+                        break monitorLoop
+                    }
+
+                    // Notify the continuation
+                    _registration.withLock { store in
+                        if let continuation = store[targetFileDescriptor] {
+                            continuation.yield(true)
```
weissi: high deadlock potential here: callout under lock.

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+        let monitorThread: pthread_t
+    }
+
+    static let shared: AsyncIO = AsyncIO()
+
+    private let state: Result<State, SubprocessError>
+
+    private init() {
+        // Create main epoll fd
+        let epollFileDescriptor = epoll_create1(CInt(EPOLL_CLOEXEC))
```
weissi, line 73: where are we closing this?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+    private func shutdown() {
+        guard case .success(let currentState) = self.state else {
+            return
+        }
+
+        var one: UInt64 = 1
+        // Wake up the thread for shutdown
+        _ = _SubprocessCShims.write(currentState.shutdownFileDescriptor, &one, MemoryLayout<UInt64>.stride)
+        // Cleanup the monitor thread
+        pthread_join(currentState.monitorThread, nil)
```
weissi, line 216: this will go badly wrong if we do this more than once. `shutdown` should protect against that. After `pthread_join` the `pthread_t` is a dangling pointer

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+        to diskIO: borrowing IOChannel
+    ) async throws -> Int {
+        return try await self._write(array, to: diskIO)
+    }
+
+    func _write<Bytes: _ContiguousBytes>(
+        _ bytes: Bytes,
+        to diskIO: borrowing IOChannel
+    ) async throws -> Int {
+        let fileDescriptor = diskIO.channel
```
weissi, line 424: do we have a _guarantee_ that `fileDescriptor` is _not_ an actual file? Because on actual files, `epoll` will always return writable even if you can't write them and `write` will block (e.g. on NFS or other network FSes).

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+import _SubprocessCShims
+import Synchronization
+
+private typealias SignalStream = AsyncThrowingStream<Bool, any Error>
+private let _epollEventSize = 256
+private let _registration: Mutex<
+    [PlatformFileDescriptor : SignalStream.Continuation]
+> = Mutex([:])
+
+final class AsyncIO: Sendable {
```
weissi, line 39: this whole file has very high complexity code that seems fairly adhoc, do we have a rigorous test suite for this?

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+            )
+            self.ioCompletionPort = .failure(error)
+            self.monitorThread = .failure(error)
+            return
+        }
+        self.ioCompletionPort = .success(port)
+        // Create monitor thread
+        let threadContext = MonitorThreadContext(ioCompletionPort: port)
+        let threadContextPtr = Unmanaged.passRetained(threadContext)
+        let threadHandle = CreateThread(nil, 0, { args in
```
al45tair: I'm a bit concerned about the use of `CreateThread` here. I don't think that's safe; the Swift runtime library uses the C library, and the C library _requires_ that you use [`_beginthreadex`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=msvc-170&redirectedfrom=MSDN) rather than [`CreateThread`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createthread), in order that its data structures are properly initialised.

To quote the `CreateThread` documentation:

> A thread in an executable that calls the C run-time library (CRT) should use the [_beginthreadex](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex) and [_endthreadex](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/endthread-endthreadex) functions for thread management rather than **CreateThread** and [ExitThread](https://learn.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-exitthread); this requires the use of the multithreaded version of the CRT. If a thread created using **CreateThread** calls the CRT, the CRT may terminate the process in low-memory conditions.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -0,0 +1,437 @@
...
+
+                let offsetAddress = bufferPointer.baseAddress!.advanced(by: readLength)
+                // Read directly into the buffer at the offset
+                return ReadFile(
+                    handle,
+                    offsetAddress,
+                    DWORD(truncatingIfNeeded: targetCount),
+                    nil,
+                    &overlapped
+                )
```
al45tair: The `OVERLAPPED` struct appears to be all zero at this point. This is fine for non-seekable handles, like pipes, but for seekable handles (like file handles), this code is wrong as in that case you need to set the offset at which the operation should take place (Win32 doesn't keep track of the file pointer for you for overlapped I/O).

Given the context, I'm pretty sure we're using it with pipes, so this isn't a bug per-se, but we should be cautious about reusing it elsewhere. It would be good to have a comment explaining this above the `ReadFile` and `WriteFile` calls in this file.

Sources/Subprocess/Configuration.swift
al45tair: On Windows, there's no way to remove a registration anyway, so this is just manipulating an internal data structure.

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
+                )
+            }
+            self.write(dispatchData, to: diskIO) { writtenLength, error in
+                if let error = error {
+                    continuation.resume(throwing: error)
+                } else {
+                    continuation.resume(returning: writtenLength)
+                }
+            }
+        }
```
al45tair: As @weissi says above, you _mustn't_ escape the pointer from the span out of the `withUnsafeBytes`. You should rearrange the code as below:

```suggestion
            span.withUnsafeBytes {
                let dispatchData = DispatchData(
                    bytesNoCopy: $0,
                    deallocator: .custom(
                        nil,
                        {
                            // noop
                        }
                    )
                )
                self.write(dispatchData, to: diskIO) { writtenLength, error in
                    if let error {
                        continuation.resume(throwing: error)
                    } else {
                        continuation.resume(returning: writtenLength)
                    }
                }
            }
        }
```

Sources/Subprocess/IO/AsyncIO+Darwin.swift
```diff
@@ -0,0 +1,165 @@
...
```
al45tair: Same problem here, with the same kind of rearrangement as the solution:

```suggestion
            array.withUnsafeBytes {
                let dispatchData = DispatchData(
                    bytesNoCopy: $0,
                    deallocator: .custom(
                        nil,
                        {
                            // noop
                        }
                    )
                )
                self.write(dispatchData, to: diskIO) { writtenLength, error in
                    if let error {
                        continuation.resume(throwing: error)
                    } else {
                        continuation.resume(returning: writtenLength)
                    }
                }
            }
        }
```

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+                } else if bytesRead == 0 {
+                    // We reached EOF. Return whatever's left
+                    try self.removeRegistration(for: fileDescriptor)
+                    guard readLength > 0 else {
+                        return nil
+                    }
+                    resultBuffer.removeLast(resultBuffer.count - readLength)
+                    return resultBuffer
+                } else {
+                    if errno == EAGAIN || errno == EWOULDBLOCK {
```
al45tair: What about `EINTR`?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+                }
+                if written > 0 {
+                    writtenLength += written
+                    if writtenLength >= bytes.count {
+                        // Wrote all data
+                        try self.removeRegistration(for: fileDescriptor)
+                        return writtenLength
+                    }
+                } else {
+                    if errno == EAGAIN || errno == EWOULDBLOCK {
```
al45tair: `EINTR`?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -0,0 +1,507 @@
...
+                }
+                if written > 0 {
+                    writtenLength += written
+                    if writtenLength >= span.byteCount {
+                        // Wrote all data
+                        try self.removeRegistration(for: fileDescriptor)
+                        return writtenLength
+                    }
+                } else {
+                    if errno == EAGAIN || errno == EWOULDBLOCK {
```
al45tair: `EINTR`?

Sources/Subprocess/IO/AsyncIO+Windows.swift
al45tair: (Specifically, the continuation may cause other code to run with the lock still held, and the lock will only be released once execution gets back here afterwards. You don't want that, because that can lead to deadlocks.)

iCharlesHu: Ahh thanks for point this one out. Totally missed it

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Unfortunately not since this is a Windows API. But I can add comments.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: While this one appears it would happen but it won't:

We calculate the write length to `WriteFile()` by `buffer.count - readLength`. Since both are `Int`, when they are `Int32` the resulting write length will never exceed `Int32.max` so the `bytesWritten` should fit just fine.

However, just to be extra sure I'll switch to `truncatingIfNeeded`.

Sources/Subprocess/Buffer.swift
iCharlesHu: I understand the concern, but this is exactly what `DispatchData.Region` looks like on Darwin. I simply copied it over because it's not available on Linux.

Sources/Subprocess/Configuration.swift
iCharlesHu: Ditto

Sources/Subprocess/Configuration.swift
iCharlesHu: This is to support the use case where user can pass a `FileDescriptor` directly to be bonded with child process. They can choose to not close the `fd` after subprocess spawns. For example, you can bind your stdout to child process and you wouldn't want Subprocess to close it for you.

Sources/Subprocess/IO/AsyncIO+Darwin.swift
iCharlesHu: Why not? Are you suggesting we have to use Mutex here?

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: Ahh good catch. Added to `shutdown()`.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: `shutdown` is only called in `atexit`, which by definition should only run once.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: Yes. The only scenario that `FileDescriptor` might be a file is when user pass in FD directly. In those cases we bind them to child process directly and we don't read from / write to them.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: > fairly adhoc

How so? Is there a more canonical way of doing AsyncIO on Linux? This new implementation passes all existing tests and I'll be adding more later: https://github.com/swiftlang/swift-subprocess/issues/97

We need to finish up the implementation first.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Thanks for letting me know! Today I learned 😆

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Yes it's only used with pipes (non-seekable). I'll add a comment.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: `UInt64(biPattern:)` requires Int64 so I had to convert twice. Otherwise I get the error message `error: cannot convert value of type 'HANDLE' (aka 'UnsafeMutableRawPointer') to expected argument type 'Int64'`

Sources/Subprocess/Configuration.swift
jakepetroules: Oh yeah, this is important for my use case in swiftlang/swift-build#638

Sources/Subprocess/IO/AsyncIO+Darwin.swift
jakepetroules: First, note that this package is still building in Swift 5 language mode (we have not updated the build system's default, even when using a Swift 6 toolchain and tools version), so that can be a cause of missing diagnostics. In this case, it's because of the many missing annotations throughout Dispatch, in particular the read handler is not marked @Sendable even though it should be based on its semantics.

I think _technically_ this is safe because this closure is guaranteed to be called on the same serial queue and is not otherwise accessed while those blocks are executing, so there is no actual concurrent access. That said, we should express this in a way that the compiler can guarantee as safe, especially if these DispatchIO annotations get corrected.

jakepetroules: I'd suggest you stick @Sendable on the handler here, similar to what I mentioned in https://github.com/swiftlang/swift-subprocess/pull/117#discussion_r2227553343

Sources/Subprocess/IO/AsyncIO+Windows.swift
jakepetroules: I doubt that matters anymore with modern CRT and Windows, a lot of those types of requirements have been removed. A couple 15 year old posts on Stack Overflow appear to make the same argument: https://stackoverflow.com/a/2101036 and https://stackoverflow.com/a/20901790

I also looked at the implementation of _beginthread (ucrt sources are distributed with the Windows SDK) and it doesn't appear to do any meaningful initialization before calling into CreateThread, as far as I can tell.

@compnerd do you happen to know more on this topic?

Sources/Subprocess/Configuration.swift
al45tair: Atomics aren't sufficient here. Named pipe names are system-wide.

Why are we using a named pipe here anyway? What's wrong with `CreatePipe()`? Using a Windows named pipe is an odd thing to do unless you want to open it from some other process (or across the network).

If we absolutely must use a named pipe for some reason, we should be trying to create it _in a loop_, with the `CreateNamedPipe` function's `dwOpenMode` set to include `FILE_FLAG_FIRST_PIPE_INSTANCE`, and then adjust the name on each pass through the loop so that we get a unique name. An additional gotcha here is that in some circumstances named pipes are restricted to the same container and must start their names with "`\\.\pipe\LOCAL\`". I think it's very likely the wrong choice for this code…

Sources/Subprocess/IO/AsyncIO+Windows.swift
al45tair: Using `_beginthreadex` is still documented as best practice. I think you're right that it's better now than it used to be, which is why the docs now comment that

> If a thread created using *CreateThread* calls the CRT, the CRT may terminate the process in low-memory conditions.

which I _think_ is newer language (the implication being that it will lazily create the state it needs, rather than just crashing).

Sources/Subprocess/Buffer.swift
weissi: Right, but let's not perpetuate bad APIs into new code. This is extremely dangerous and just because DispatchData does dangerous things, we shouldn't also do that, right?

Sources/Subprocess/IO/AsyncIO+Linux.swift
weissi: Using epoll/kqueue correctly is notoriously difficult, so typically you would create an abstraction for it and rigorously test that. For example, NIO has a `Selector` (https://github.com/apple/swift-nio/blob/main/Sources/NIOPosix/SelectorGeneric.swift) implementation which abstracts a lot of the difficult stuff.

Not abstracting that usually leads to very-hard-to-debug stalls or CPU spins etc.

Sources/Subprocess/Configuration.swift
weissi: > Named pipe names are system-wide

I'm glad you looked over this. Had no idea this is a named pipe :P.

Sources/Subprocess/IO/AsyncIO+Darwin.swift
weissi: > First, note that this package is still building in Swift 5 language mode

Hang on, `swift-tools-version:6.0` will make it `-swift-version 6` (unlike just invoking `swiftc` on 6.0+). Anyway, we should be in language mode 6.

> be called on the same serial queue

it's on `.global()` which is concurrent.

Sources/Subprocess/Configuration.swift
iCharlesHu: @al45tair we have to use named pipe here because Microsoft's documentation states that:

> Asynchronous (overlapped) read and write operations are not supported by anonymous pipes.

See https://learn.microsoft.com/en-us/windows/win32/ipc/anonymous-pipe-operations

Sources/Subprocess/Buffer.swift
iCharlesHu: It's this way because we have to use `DisaptchData.Region` on Darwin so this is to make it consistent across platforms. You can argue we shouldn't use `DispatchData.Region` on Darwin either but I'd rather make that change separately since this one is big enough as it is.

iCharlesHu: I just realized that since now Linux no longer use `DispatchData` we can remove this type. Darwin will continue to use `DispatchData.Region`

Sources/Subprocess/IO/AsyncIO+Darwin.swift
iCharlesHu: I switched to a dedicated serial queue.

I agree with @jakepetroules that this is safe because all calls are made on the same serial queue. I could mark the closure as `@Sendable` but that would require us to actually use an unnecessary lock. This is analogues to the old days of `actor` approach where you just basically use a serial queue to make sure there's no concurrent access.

jakepetroules: > Hang on, swift-tools-version:6.0 will make it -swift-version 6 (unlike just invoking swiftc on 6.0+).

Sorry, you're right, I might've been thinking of an old bug.

> I agree with @jakepetroules that this is safe because all calls are made on the same serial queue. I could mark the closure as @Sendable but that would require us to actually use an unnecessary lock. This is analogues to the old days of actor approach where you just basically use a serial queue to make sure there's no concurrent access.

At the same time, it's a known issue that the closure is not marked @Sendable. Fixing that bug WILL break the build of Swift Subprocess.

So I suggest writing the read method like this, which is compliant with full Sendable checking (without any Mutex) and insulates us against a potential fix there:

```swift
internal func read(
    from dispatchIO: DispatchIO,
    upTo maxLength: Int
) async throws -> DispatchData? {
    let stream = AsyncThrowingStream<DispatchData, any Error> { continuation in
        dispatchIO.read(
            offset: 0,
            length: maxLength,
            queue: .global()
        ) { @Sendable done, data, error in
            if error != 0 {
                continuation.finish(
                    throwing: SubprocessError(
                        code: .init(.failedToReadFromSubprocess),
                        underlyingError: .init(rawValue: error)
                    )
                )
                return
            }
            if let data {
                continuation.yield(data)
            }
            if done {
                continuation.finish()
            }
        }
    }
    var buffer: DispatchData = .empty
    for try await chunk in stream {
        buffer.append(chunk)
    }
    return buffer.isEmpty ? nil : buffer
}
```

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -222,6 +214,8 @@ final class AsyncIO: Sendable {
         _ = _SubprocessCShims.write(currentState.shutdownFileDescriptor, &one, MemoryLayout<UInt64>.stride)
         // Cleanup the monitor thread
         pthread_join(currentState.monitorThread, nil)
+        close(currentState.epollFileDescriptor)
```
jakepetroules: Add fatalError's if close returns nonzero?

Package.swift
```diff
@@ -6,7 +6,7 @@ import PackageDescription
 var dep: [Package.Dependency] = [
     .package(
         url: "https://github.com/apple/swift-system",
-        from: "1.5.0"
+        exact: "1.5.0"
```
jakepetroules, line 11: I'm guessing you're pinning to 1.5.0 to avoid the regression introduced on Focal? Suggest adding a comment to that effect if so; "exact" is generally discouraged

jakepetroules:
This is coming along really nicely! It looks pretty good to me at this point, and the last thing I'd like to see addressed before I'd feel comfortable giving a formal ✅ is the @sendable issues with DispatchIO detailed in https://github.com/swiftlang/swift-subprocess/pull/117#discussion_r2232205381

Tests/SubprocessTests/SubprocessTests+Unix.swift
```diff
@@ -668,14 +668,11 @@ extension SubprocessUnixTests {
         var platformOptions = PlatformOptions()
         platformOptions.supplementaryGroups = Array(expectedGroups)
         let idResult = try await Subprocess.run(
-            .name("swift"),
+            .path("/usr/bin/swift"),
```
jakepetroules: Does this really still hang without this change, considering the brand-new I/O backend? Could we revert this change? (it's useful for platforms where swift is not located in /usr/bin)

Sources/Subprocess/IO/AsyncIO+Darwin.swift
iCharlesHu: `AsyncStream` uses a lock internally. This is a hot code path and it's continuously called when streaming. We really shouldn't put a lock here unless we are sure it is absolutely necessary. IMO, this is the same `~Copyable` optional hack we use everywhere. The compiler isn't able to reason that even though we are sending this value to another execution context, it's done sequentially with no overlapping.

If the concern is that when `Dispatch` adds the `@Sendable` annotation, Subprocess will break, I'd rather figure out a (`@unchecked Sendable`) solution then than prematurely optimize it now.

Of course, I could be wrong about this being safe so I'm definitely all ears. I just don't want to change our approach simply because the compiler isn't able to reason our specific use case.

Sources/Subprocess/Configuration.swift
al45tair: Ah, good point, though the actual problem is really just that `CreatePipe()` doesn't have a flags argument for you to specify `FILE_FLAG_OVERLAPPED`. Interestingly, Windows apparently used to use a named pipe anyway to implement anonymous pipes (see https://stackoverflow.com/questions/60645/overlapped-i-o-on-anonymous-pipe).

iCharlesHu: @al45tair thanks for linking that! Thanks to @jakepetroules's counter suggestion our current implementation of `CreatedPipe` is (almost) the same as the one listed above.

Sources/Subprocess/Configuration.swift
```diff
@@ -640,84 +651,277 @@ internal struct TrackedFileDescriptor: ~Copyable {
...
+                    &saAttributes
+                )
+            }
+            guard let parentEnd, parentEnd != INVALID_HANDLE_VALUE else {
+                // Since we created the pipe with `FILE_FLAG_FIRST_PIPE_INSTANCE`,
+                // if there's already a pipe with the same name, GetLastError()
+                // will be set to FILE_FLAG_FIRST_PIPE_INSTANCE. In this case,
+                // try again with a different name.
+                let errorCode = GetLastError()
+                guard errorCode != FILE_FLAG_FIRST_PIPE_INSTANCE else {
```
al45tair: Shouldn't this be `ERROR_ACCESS_DENIED`?

Tests/SubprocessTests/SubprocessTests+Unix.swift
iCharlesHu: It's actually due to https://github.com/swiftlang/swift-subprocess/issues/105 which I addressed. I reverted it back to `.name()`.

iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/105

Sources/Subprocess/Configuration.swift
iCharlesHu: Ahhh copy and paste issue. Thank you so much for catching this!!!

Tests/SubprocessTests/SubprocessTests+Unix.swift
jakepetroules: Oh, nice!

Sources/Subprocess/IO/AsyncIO+Darwin.swift
jakepetroules: I can see your argument for performance and agree `@unchecked Sendable` might be a good compromise for now (unless @weissi or @FranzBusch knows something we don't w.r.t. memory safety) - would you mind throwing that annotation on the closure with a note about why it's there? That way we're insulated against future Dispatch changes and it's also explicit that we're knowingly bypassing compiler checking.

jakepetroules (approved):
Sticking the `@unchecked @Sendable` on the DisptatchIO handler as discussed would be a nice final bit of polish, but I think this is ready to land either way from my perspective.

iCharlesHu:
> Sticking the `@unchecked @Sendable` on the DisptatchIO handler as discussed would be a nice final bit of polish, but I think this is ready to land either way from my perspective.

Unfortunately, you can't use `@unchecked @Sendable` on closures, only types -- this is largely why I decided to not bother now. We'd have to come up with some hack that involves us moving that closure to some type with `@unchecked Sendable`, which I'm not sure it will work until we see the actual error message.

iCharlesHu:
Appears to also resolve: https://github.com/swiftlang/swift-subprocess/issues/119

Sources/Subprocess/Configuration.swift
```diff
@@ -640,84 +651,277 @@ internal struct TrackedFileDescriptor: ~Copyable {
...
+                    pipeNameW,
+                    openMode,
+                    DWORD(PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT),
+                    1, // Max instance,
+                    DWORD(readBufferSize),
+                    DWORD(readBufferSize),
+                    0,
+                    &saAttributes
+                )
+            }
```
al45tair, line 869: OK, one final thing about this code (sorry). I think we should at least add `PIPE_REJECT_REMOTE_CLIENTS` to the flags.

Ideally I think we'd specify an ACL for the named pipe, to make it harder to interfere with this code by having some other process call `CreateFile` between the `CreateNamedPipeW` and `CreateFileW` calls here, but the window for mischief is quite small and `PIPE_REJECT_REMOTE_CLIENTS` would at least restrict it to things already running on the current machine.

PR #121, opened 2025-07-14
Subprocess: avoid some `strdup` deprecation warnings on Windows

compnerd:
Use the POSIX compliant `_strdup` on Windows instead of `strdup` to avoid some warnings when building.


jakepetroules (approved):
Thanks. Please make sure to get @iCharlesHu's approval too.

PR #122, opened 2025-07-14
build: replicate the SPM style build, address TODO

compnerd:
This restructures the internal modules to match what we do in SPM. This is primarily to help make it easier to mirror behaviour.

Address the TODO, adding support to build against a swift-system build tree which is going to be required for integrating this into the Windows toolchain build. If one is not provided, continue the ability to vendor swift-system for ease of local development.


jakepetroules (approved):
Nice! Please make sure to get @iCharlesHu's approval too.

PR #124, opened 2025-07-16
build: add support for installing the image

compnerd:
This replicates much of the logic from the new runtimes build to support installing the Swift interface and the runtime components. Additionally, we are now able to generate the targets configuration to allow dependencies between projects.


jakepetroules (approved):

cmake/modules/EmitSwiftInterface.cmake
```diff
@@ -0,0 +1,60 @@
...
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule/${${PROJECT_NAME}_MODULE_TRIPLE}.swiftmodule>)
+
+  # Generate textual swift interfaces is library-evolution is enabled
+  if(${PROJECT_NAME}_ENABLE_LIBRARY_EVOLUTION)
+    target_compile_options(${target} PRIVATE
+      $<$<COMPILE_LANGUAGE:Swift>:-emit-module-interface-path$<SEMICOLON>${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule/${${PROJECT_NAME}_MODULE_TRIPLE}.swiftinterface>
+      $<$<COMPILE_LANGUAGE:Swift>:-emit-private-module-interface-path$<SEMICOLON>${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule/${${PROJECT_NAME}_MODULE_TRIPLE}.private.swiftinterface>)
+    if(${PROJECT_NAME}_VARIANT_MODULE_TRIPLE)
+      target_compile_options(${target} PRIVATE
+        "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-emit-variant-module-interface-path ${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule/${${PROJECT_NAME}_VARIANT_MODULE_TRIPLE}.swiftinterface>"
```
etcwilde: When is this being built using CMake in an environment where we have ABI stability?

cmake/modules/EmitSwiftInterface.cmake
```diff
@@ -0,0 +1,60 @@
...
+  # Account for an existing swiftmodule file generated with the previous logic
+  if(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule"
+     AND NOT IS_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule")
+    message(STATUS "Removing regular file ${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule to support nested swiftmodule generation")
+    file(REMOVE "${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule")
+  endif()
+
+  target_compile_options(${target} PRIVATE
+    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-emit-module-path ${CMAKE_CURRENT_BINARY_DIR}/${module_name}.swiftmodule/${${PROJECT_NAME}_MODULE_TRIPLE}.swiftmodule>")
+  if(${PROJECT_NAME}_VARIANT_MODULE_TRIPLE)
```
etcwilde: Does this need catalyst support? I would rather we keep it simpler if we can.

.github/workflows/pull_request.yml
```diff
@@ -40,7 +40,7 @@ jobs:
   cmake_build:
     name: CMake Build
     runs-on: ubuntu-latest
-    container: swift:6.1-noble
+    container: swiftlang/swift:nightly-noble
```
etcwilde, line 43: Was this bump intentional? I had used 6.1 to keep it in sync with the required version in `Package.swift`.

jakepetroules: > we don't need the CMake build to be compatible with 6.1, since it's only used for the build of the toolchain itself

and it would require more effort to make this compatible with 6.1; no reason to add support for something we will never use

etcwilde: oh, this was for the `platform` and `arch` subdirectories wasn't it?

etcwilde: > since it's only used for the build of the toolchain itself.

It isn't though. I have projects that are non-SwiftPM that use this build today.

etcwilde: We could set `Subprocess_INSTALL_NESTED_SUBDIR` to `OFF` when using something earlier than 6.2 and the arch and platform subdirs aren't set. Then we wouldn't need to request that info from the compiler.

cmake/modules/EmitSwiftInterface.cmake
compnerd: It is not - but its easier to copy-pasta the file than to have different versions per repository.

cmake/modules/EmitSwiftInterface.cmake
compnerd: Likewise

.github/workflows/pull_request.yml
compnerd: Windows will want that, so we will need to set that to true for CI

cmake/modules/EmitSwiftInterface.cmake
compnerd: Spoke offline, removing the catalyst support.

cmake/modules/EmitSwiftInterface.cmake
compnerd: The library evolution I would like to keep as the intent is to try to enable that for Windows.

.github/workflows/pull_request.yml
jakepetroules: > > since it's only used for the build of the toolchain itself.
> 
> It isn't though. I have projects that are non-SwiftPM that use this build today.

Ah, that's unfortunate...

iCharlesHu (approved):

Issue #111, opened 2025-07-02
Remove PROCESS_INFORMATION from Execution


iCharlesHu:
https://github.com/swiftlang/swift-subprocess/pull/93 added `PROCESS_INFORMATION` as a property to `Execution`, making it no longer "stateless". We want to eventually make `Execution` `~Copyable` so we should remove `PROCESS_INFORMATION` from `Execution` and find another way to save the process handle.

jakepetroules:
Why is the presence of `PROCESS_INFORMATION` incompatible with `Execution` being `~Copyable`?

compnerd:
I think that adding a retroactive `Sendable` conformance would be the right thing to do. As @jakepetroules mentioned, `PROCESS_INFORMATION` shouldn't be incompatible with `~Copyable`.

PR #125, opened 2025-07-16
Introduce process file descriptor (pidfd) based process monitoring for Linux

iCharlesHu:
The current process monitoring code for Linux has a flaw that makes it susceptible to infinite hangs under specific conditions:
- The parent process uses any other method (other than Subprocess itself) to spawn new processes in addition to spawning with `Subprocess`.
- The parent process fails to properly reap the non-Subprocess-spawned process, leaving it as a zombie in the process table.

This is because currently, we rely on running `waitid()` with `P_ALL` and `WNOWAIT` in an infinite loop to detect possible child process state transitions. However, we don’t reap the child process (by specifying `WNOWAIT`) unless we (Subprocess) actually spawned the process.

Here’s a simplified pseudo-code to illustrate the issue:

```swift
while true {
    var siginfo = siginfo_t()
    // We’re not reaping the child process
    if waitid(P_ALL, id_t(0), &siginfo, WEXITED | WNOWAIT) == 0 {
        guard let c = savedContinuation else {
            // If there’s no saved continuation, we didn’t spawn the process
            // In this case, we don’t reap the child process
            continue
        }

        siginfo = siginfo_t()
        waitid(P_PID, numericCast(pid), &siginfo, WEXITED) // We’re actually reaping the child process
    }
}
```

With this setup, if there are zombie children in the process table without reaping, `waitid(P_ALL)` will repeatedly return the same (non-Subprocess-spawned) PID with every call, causing an infinite loop.

You can observe this behavior with the following sample code:

```swift
let arguments = "\"\""

let pid = arguments.withCString { args in
    var pid: pid_t = -1
    let status = posix_spawn(&pid, "/bin/echo", nil, nil, [strdup(args)] + [nil], environ)
    guard status == 0 else {
        fatalError("posix_spawn: \(status), errno: \(errno)")
    }
    return pid
}
print("echo pid: \(pid)")

    .path("/bin/cat"),
    arguments: ["Package.swift"],
    output: .string(limit: .max, encoding: UTF8.self),
    error: .discarded
)
print("cat finished: \(result.terminationStatus)")
print("cat output: \(result.standardOutput ?? "")")
```

After running this example, you’ll notice that the parent process seems to be stuck, and the “cat finished” message is never printed. This is because the parent process never calls `waitid` on the `echo` call, leaving it in the process table. Consequently, the monitor thread runs in an infinite loop.

While some may argue that this is not a bug in `Subprocess`, but rather an issue with the parent code, since the POSIX standard mandates that the process spawning child process must reap the child process via `waitid`. However, Subprocess should still not hang due to someone else’s bug.

To resolve this issue, switch to a **Linux-specific** process monitoring method by creating and observing the process file descriptor (pidfd) using epoll. This approach is similar to the epoll implementation introduced in https://github.com/swiftlang/swift-subprocess/pull/117, with the only difference being that we’re polling pidfd instead of a regular file descriptor.

As part of this change, I also unified the “process handle” design to make it easier to expose process handles to clients later (after the 1.0 release, as requested by https://github.com/swiftlang/swift-subprocess/pull/101). We chose to use `ProcessIdentifier` to host platform-specific process file descriptors and process handles because it perfectly aligns with the original use case. To ensure flexibility, we opted for a concrete `ProcessIdentifier` type instead of just a number, allowing us to add more information if necessary.

### DIFF
--- a/Sources/Subprocess/Execution.swift
+++ b/Sources/Subprocess/Execution.swift
@@ -34,23 +34,11 @@ public struct Execution: Sendable {
     /// The process identifier of the current execution
     public let processIdentifier: ProcessIdentifier
 
-    #if os(Windows)
-    internal let consoleBehavior: PlatformOptions.ConsoleBehavior
-
-    init(
-        processIdentifier: ProcessIdentifier,
-        consoleBehavior: PlatformOptions.ConsoleBehavior
-    ) {
-        self.processIdentifier = processIdentifier
-        self.consoleBehavior = consoleBehavior
-    }
-    #else
     init(
         processIdentifier: ProcessIdentifier
     ) {
         self.processIdentifier = processIdentifier
     }
-    #endif  // os(Windows)
 }
 
 // MARK: - Output Capture
--- a/Sources/Subprocess/Execution.swift
+++ b/Sources/Subprocess/Execution.swift
@@ -35,16 +35,13 @@ public struct Execution: Sendable {
     public let processIdentifier: ProcessIdentifier
 
     #if os(Windows)
-    internal nonisolated(unsafe) let processInformation: PROCESS_INFORMATION
     internal let consoleBehavior: PlatformOptions.ConsoleBehavior
 
     init(
         processIdentifier: ProcessIdentifier,
-        processInformation: PROCESS_INFORMATION,
         consoleBehavior: PlatformOptions.ConsoleBehavior
     ) {
         self.processIdentifier = processIdentifier
-        self.processInformation = processInformation
         self.consoleBehavior = consoleBehavior
     }
     #else
@@ -54,17 +51,6 @@ public struct Execution: Sendable {
         self.processIdentifier = processIdentifier
     }
     #endif  // os(Windows)
-
-    internal func release() {
-        #if os(Windows)
-        guard CloseHandle(processInformation.hThread) else {
-            fatalError("Failed to close thread HANDLE: \(SubprocessError.UnderlyingError(rawValue: GetLastError()))")
-        }
-        guard CloseHandle(processInformation.hProcess) else {
-            fatalError("Failed to close process HANDLE: \(SubprocessError.UnderlyingError(rawValue: GetLastError()))")
-        }
-        #endif
-    }
 }
 
 // MARK: - Output Capture
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -477,26 +477,41 @@ extension Configuration {
     }
 }
 
-// Special keys used in Error's user dictionary
-extension String {
-    static let debugDescriptionErrorKey = "NSDebugDescription"
+// MARK: - ProcessIdentifier
+
+/// A platform independent identifier for a Subprocess.
+public struct ProcessIdentifier: Sendable, Hashable {
+    /// The platform specific process identifier value
+    public let value: pid_t
+
+    public init(value: pid_t) {
+        self.value = value
+    }
+
+    internal func close() { /* No-op on Darwin */ }
+}
+
+extension ProcessIdentifier: CustomStringConvertible, CustomDebugStringConvertible {
+    public var description: String { "\(self.value)" }
+
+    public var debugDescription: String { "\(self.value)" }
 }
 
 // MARK: - Process Monitoring
 @Sendable
 internal func monitorProcessTermination(
-    forExecution execution: Execution
+    for processIdentifier: ProcessIdentifier
 ) async throws -> TerminationStatus {
     return try await withCheckedThrowingContinuation { continuation in
         let source = DispatchSource.makeProcessSource(
-            identifier: execution.processIdentifier.value,
+            identifier: processIdentifier.value,
             eventMask: [.exit],
             queue: .global()
         )
         source.setEventHandler {
             source.cancel()
             var siginfo = siginfo_t()
-            let rc = waitid(P_PID, id_t(execution.processIdentifier.value), &siginfo, WEXITED)
+            let rc = waitid(P_PID, id_t(processIdentifier.value), &siginfo, WEXITED)
             guard rc == 0 else {
                 continuation.resume(
                     throwing: SubprocessError(
--- a/Sources/Subprocess/Platforms/Subprocess+Linux.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Linux.swift
@@ -304,7 +304,7 @@ internal func monitorProcessTermination(
                 // pidfd is only supported on Linux kernel 5.4 and above
                 // On older releases, use signalfd so we do not need
                 // to register anything with epoll
-                if _isLinuxKernelAtLeast(major: 5, minor: 4) {
+                if processIdentifier.processFileDescriptor > 0 {
                     // Register processFileDescriptor with epoll
                     var event = epoll_event(
                         events: EPOLLIN.rawValue,
@@ -418,8 +418,10 @@ private extension siginfo_t {
     }
 }
 
-// Okay to be unlocked global mutable because this thread is only created once
+// Okay to be unlocked global mutable because this value is only set once like dispatch_once
 private nonisolated(unsafe) var _signalPipe: (readEnd: CInt, writeEnd: CInt) = (readEnd: -1, writeEnd: -1)
+// Okay to be unlocked global mutable because this value is only set once like dispatch_once
+private nonisolated(unsafe) var _waitProcessFileDescriptorSupported = false
 private let _processMonitorState: Mutex<ProcessMonitorState> = .init(.notStarted)
 
 private func shutdown() {
@@ -518,7 +520,7 @@ private func monitorThreadFunc(args: UnsafeMutableRawPointer?) -> UnsafeMutableR
             }
 
             // P_PIDFD requires Linux Kernel 5.4 and above
-            if _isLinuxKernelAtLeast(major: 5, minor: 4) {
+            if _waitProcessFileDescriptorSupported {
                 _blockAndWaitForProcessFileDescriptor(targetFileDescriptor, context: context)
             } else {
                 _reapAllKnownChildProcesses(targetFileDescriptor, context: context)
@@ -569,9 +571,9 @@ private let setup: () = {
         return
     }
 
-    // On Linux kernel lower than 5.4 we have to rely on signal handler
+    // If the current kernel does not support pidfd, fallback to signal handler
     // Create the self-pipe that signal handler writes to
-    if !_isLinuxKernelAtLeast(major: 5, minor: 4) {
+    if !_isWaitProcessFileDescriptorSupported() {
         var pipeCreationError: SubprocessError? = nil
         do {
             let (readEnd, writeEnd) = try FileDescriptor.pipe()
@@ -608,6 +610,9 @@ private let setup: () = {
             _reportFailureWithErrno(errno)
             return
         }
+    } else {
+        // Mark waitid(P_PIDFD) as supported
+        _waitProcessFileDescriptorSupported = true
     }
     let monitorThreadContext = MonitorThreadContext(
         epollFileDescriptor: epollFileDescriptor,
@@ -755,36 +760,25 @@ private func _reapAllKnownChildProcesses(_ signalFd: CInt, context: MonitorThrea
     }
 }
 
-private func _isLinuxKernelAtLeast(major: Int, minor: Int) -> Bool {
-    var buffer = utsname()
-    guard uname(&buffer) == 0 else {
+internal func _isWaitProcessFileDescriptorSupported() -> Bool {
+    // waitid(P_PIDFD) is only supported on Linux kernel 5.4 and above
+    // Prob whether the current system supports it by calling it with self pidfd
+    // and checking for EINVAL (waitid sets errno to EINVAL if it does not
+    // recognize the id type).
+    var siginfo = siginfo_t()
+    let selfPidfd = _pidfd_open(getpid())
+    if selfPidfd < 0 {
+        // If we can not retrieve pidfd, the system does not support waitid(P_PIDFD)
         return false
     }
-    let releaseString = withUnsafeBytes(of: &buffer.release) { ptr in
-        let buffer = ptr.bindMemory(to: CChar.self)
-        return String(cString: buffer.baseAddress!)
-    }
-
-    // utsname release follows the format
-    // major.minor.patch-extra
-    guard let mainVersion = releaseString.split(separator: "-").first else {
-        return false
-    }
-
-    let versionComponents = mainVersion.split(separator: ".")
-
-    guard let currentMajor = Int(versionComponents[0]),
-          let currentMinor = Int(versionComponents[1]) else {
-        return false
-    }
-
-    if currentMajor > major {
-        return true
-    } else if currentMajor == major {
-        return currentMinor >= minor
-    } else {
-        return false
-    }
+    /// The following call will fail either with
+    /// - ECHILD: in this case we know P_PIDFD is supported and waitid correctly
+    ///     reported that we don't have a child with the same selfPidfd;
+    /// - EINVAL: in this case we know P_PIDFD is not supported because it does not
+    ///     recognize the `P_PIDFD` type
+    waitid(idtype_t(UInt32(P_PIDFD)), id_t(selfPidfd), &siginfo, WEXITED | WNOWAIT)
+    return errno == ECHILD
 }
+
 
 #endif  // canImport(Glibc) || canImport(Android) || canImport(Musl)
--- a/Sources/Subprocess/Platforms/Subprocess+Linux.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Linux.swift
@@ -142,7 +142,6 @@ extension Configuration {
                     errorRead: nil,
                     errorWrite: errorWriteFileDescriptor
                 )
-
                 let execution = Execution(
                     processIdentifier: .init(
                         value: pid,
@@ -186,7 +185,7 @@ extension Configuration {
     }
 }
 
-// MARK:  - ProcesIdentifier
+// MARK:  - ProcessIdentifier
 
 /// A platform independent identifier for a Subprocess.
 public struct ProcessIdentifier: Sendable, Hashable {
@@ -200,7 +199,9 @@ public struct ProcessIdentifier: Sendable, Hashable {
     }
 
     internal func close() {
-        _SubprocessCShims.close(self.processFileDescriptor)
+        if self.processFileDescriptor > 0 {
+            _SubprocessCShims.close(self.processFileDescriptor)
+        }
     }
 }
 
@@ -289,40 +290,85 @@ internal func monitorProcessTermination(
     for processIdentifier: ProcessIdentifier
 ) async throws -> TerminationStatus {
     return try await withCheckedThrowingContinuation { continuation in
-        _processMonitorState.withLock { state in
+        let status = _processMonitorState.withLock { state -> Result<TerminationStatus, SubprocessError>? in
             switch state {
             case .notStarted:
-                continuation.resume(throwing: SubprocessError(
+                let error = SubprocessError(
                     code: .init(.failedToMonitorProcess),
-                    underlyingError: nil)
+                    underlyingError: nil
                 )
+                return .failure(error)
             case .failed(let error):
-                continuation.resume(throwing: error)
+                return .failure(error)
             case .started(let storage):
-                // Register processFileDescriptor with epoll
-                var event = epoll_event(
-                    events: EPOLLIN.rawValue,
-                    data: epoll_data(fd: processIdentifier.processFileDescriptor)
-                )
-                let rc = epoll_ctl(
-                    storage.epollFileDescriptor,
-                    EPOLL_CTL_ADD,
-                    processIdentifier.processFileDescriptor,
-                    &event
-                )
-                if rc != 0 {
-                    let error = SubprocessError(
-                        code: .init(.failedToMonitorProcess),
-                        underlyingError: .init(rawValue: errno)
+                // pidfd is only supported on Linux kernel 5.4 and above
+                // On older releases, use signalfd so we do not need
+                // to register anything with epoll
+                if _isLinuxKernelAtLeast(major: 5, minor: 4) {
+                    // Register processFileDescriptor with epoll
+                    var event = epoll_event(
+                        events: EPOLLIN.rawValue,
+                        data: epoll_data(fd: processIdentifier.processFileDescriptor)
                     )
-                    continuation.resume(throwing: error)
-                    return
+                    let rc = epoll_ctl(
+                        storage.epollFileDescriptor,
+                        EPOLL_CTL_ADD,
+                        processIdentifier.processFileDescriptor,
+                        &event
+                    )
+                    if rc != 0 {
+                        let epollErrno = errno
+                        let error = SubprocessError(
+                            code: .init(.failedToMonitorProcess),
+                            underlyingError: .init(rawValue: epollErrno)
+                        )
+                        return .failure(error)
+                    }
+                    // Now save the registration
+                    var newState = storage
+                    newState.continuations[processIdentifier.processFileDescriptor] = continuation
+                    state = .started(newState)
+                    // No state to resume
+                    return nil
+                } else {
+                    // Fallback to using signal handler directly on older Linux kernels
+                    // Since Linux coalesce signals, it's possible by the time we request
+                    // monitoring the process has already exited. Check to make sure that
+                    // is not the case and only save continuation then.
+                    var siginfo = siginfo_t()
+                    // Use NOHANG here because the child process might still be running
+                    if 0 == waitid(P_PID, id_t(processIdentifier.value), &siginfo, WEXITED | WNOHANG) {
+                        // If si_pid and si_signo are both 0, the child is still running since we used WNOHANG
+                        if siginfo.si_pid == 0 && siginfo.si_signo == 0 {
+                            // Save this continuation to be called by signal hander
+                            var newState = storage
+                            newState.continuations[processIdentifier.processFileDescriptor] = continuation
+                            state = .started(newState)
+                            return nil
+                        }
+
+                        switch siginfo.si_code {
+                        case .init(CLD_EXITED):
+                            return .success(.exited(siginfo.si_status))
+                        case .init(CLD_KILLED), .init(CLD_DUMPED):
+                            return .success(.unhandledException(siginfo.si_status))
+                        default:
+                            fatalError("Unexpected exit status: \(siginfo.si_code)")
+                        }
+                    } else {
+                        let waitidError = errno
+                        let error = SubprocessError(
+                            code: .init(.failedToMonitorProcess),
+                            underlyingError: .init(rawValue: waitidError)
+                        )
+                        return .failure(error)
+                    }
                 }
-                // Now save the registration
-                var newState = storage
-                newState.continuations[processIdentifier.processFileDescriptor] = continuation
-                state = .started(newState)
             }
+        }
+
+        if let status {
+            continuation.resume(with: status)
         }
     }
 }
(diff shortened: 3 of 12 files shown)


iCharlesHu:
Resolves https://github.com/swiftlang/swift-subprocess/issues/111

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -189,6 +194,30 @@ extension Configuration {
     }
 }
 
+// MARK:  - ProcesIdentifier
+
+/// A platform independent identifier for a Subprocess.
+public struct ProcessIdentifier: Sendable, Hashable {
```
cthielen, line 191: Is it worthwhile to make this a protocol given the repetition?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -265,67 +294,70 @@ extension String {
...
-            let oldContinuation = continuations.updateValue(continuation, forKey: execution.processIdentifier.value)
-            precondition(oldContinuation == nil)
-
-            // Wake up the waiter thread if it is waiting for more child processes.
-            _ = pthread_cond_signal(_waitThreadNoChildrenCondition)
+    return try await withCheckedThrowingContinuation { continuation in
+        _processMonitorState.withLock { state in
+            switch state {
+            case .notStarted:
+                continuation.resume(throwing: SubprocessError(
```
cthielen: I've been told resuming a continuation from within a lock can be troublesome.

Could you return the continuation from `withLock` and then resume?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+        let error = SubprocessError(
+            code: .init(.failedToMonitorProcess),
+            underlyingError: .init(rawValue: errno)
+        )
+        _processMonitorState.withLock { state in
+            guard case .started(let storage) = state else {
+                return
+            }
+            for continuation in storage.continuations.values {
+                continuation.resume(throwing: error)
```
cthielen: Same note here about resuming a continuation from a lock

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+    let state = _processMonitorState.withLock { state -> (shutdownFileDescriptor: CInt, monitorThread: pthread_t)? in
+        switch state {
+        case .failed(_), .notStarted:
+            return nil
+        case .started(let storage):
+            return (storage.shutdownFileDescriptor, storage.monitorThread)
+        }
+    }
+
+    guard let state = state else {
```
cthielen:
```suggestion
    guard let state else {
```

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+        return
+    }
+
+    var one: UInt64 = 1
+    // Wake up the thread for shutdown
+    _ = _SubprocessCShims.write(state.shutdownFileDescriptor, &one, MemoryLayout<UInt64>.stride)
```
cthielen:
```suggestion
    _ = _SubprocessCShims.write(state.shutdownFileDescriptor, &one, MemoryLayout<UInt64>.size)
```

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
-                // state of a terminated (zombie) child process, allowing us to fetch the
-                // continuation (if available) before reaping.
+    rc = pthread_create(&thread, nil, { args in
+        func reportError(_ error: SubprocessError) {
+            _processMonitorState.withLock { state in
+                guard case .started(let storage) = state else {
+                    return
+                }
+                for continuation in storage.continuations.values {
+                    continuation.resume(throwing: error)
```
cthielen: Same comment about locks + continuations

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+                }
+            }
+        }
+
+        let unmanaged = Unmanaged<MonitorThreadContext>.fromOpaque(args!)
```
cthielen: Is `unmanaged` used again?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+
+                // Notify the continuation
+                _processMonitorState.withLock { state in
+                    guard case .started(let storage) = state,
+                          let continuation = storage.continuations[targetFileDescriptor] else {
+                        return
+                    }
+                    switch terminationStatus {
+                    case .success(let value):
+                        continuation.resume(returning: value)
```
cthielen: Lock + continuation

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+                    state = .started(newStorage)
                 }
             }
-        },
-        nil
-    )
+        }
+
+        return nil
+    },threadContext.toOpaque())
```
cthielen: Would it be helpful if the third argument to `pthread_create()` were its own function? The `threadContext` argument feels disconnected from the call site many lines up.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -681,12 +681,26 @@ extension Environment {
 // MARK: - ProcessIdentifier
 
 /// A platform independent identifier for a subprocess.
-public struct ProcessIdentifier: Sendable, Hashable, Codable {
+public struct ProcessIdentifier: Sendable, Hashable {
```
cthielen, line 677: Same comment here about whether a protocol would help with `ProcessIdentifier`

Sources/Subprocess/Platforms/Subprocess+Linux.swift
jakepetroules: This is definitely the case for `withTaskCancellationHandler` (see https://developer.apple.com/documentation/swift/withtaskcancellationhandler(operation:oncancel:isolation:)#Cancellation-handlers-and-locks); I'm not sure if that's true for continuations.

Sources/Subprocess/Execution.swift
```diff
@@ -35,16 +35,13 @@ public struct Execution: Sendable {
     public let processIdentifier: ProcessIdentifier
 
     #if os(Windows)
-    internal nonisolated(unsafe) let processInformation: PROCESS_INFORMATION
     internal let consoleBehavior: PlatformOptions.ConsoleBehavior
```
jakepetroules: nit: unrelated, but you could delete `consoleBehavior` as well as nothing actually uses it.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -189,6 +194,30 @@ extension Configuration {
     }
 }
 
+// MARK:  - ProcesIdentifier
```
jakepetroules: nit: spelling: "Proces" -> "Process"

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -189,6 +194,30 @@ extension Configuration {
...
+
+/// A platform independent identifier for a Subprocess.
+public struct ProcessIdentifier: Sendable, Hashable {
+    /// The platform specific process identifier value
+    public let value: pid_t
+    internal let processFileDescriptor: PlatformFileDescriptor
```
jakepetroules: > waitid with P_PIDFD was introduced in Linux kernel 5.4, which focal should have. I'm looking into what's missing

@iCharlesHu One thing I'm concerned about this with change is that unfortunately I don't think we can guarantee that the Linux kernel is version 5.4 or later on all Linux distributions officially supported by Swift.

_(Also note that PIDFD_NONBLOCK, if we want to use it, is Linux kernel 5.10 or later)_

Containers aren't particularly useful for testing this because they're not going to be using the same kernel as the actual OS distribution. Also, the Swift project is dropping support for Focal in Swift 6.2. Are you planning to keep support for Swift 6.1 in SwiftSubprocess?

Here's the minimum kernel versions associated with each Linux distribution currently officially supported by the Swift project and that I expect to be supported for the Swift 6.2 release:

- Amazon Linux 2 (kernel 4.14): https://docs.aws.amazon.com/linux/al2/ug/aml2-kernel.html
- Debian 12 (kernel 6.1): https://www.debian.org/News/2023/20230610.en.html
- Fedora 39 (kernel 6.5): https://en.wikipedia.org/wiki/Fedora_Linux_release_history, also that version is EoL anyways
- RHEL UBI9 (kernel 5.14): https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/9.0_release_notes/new-features#enhancement_kernel
- Ubuntu 20.04 (kernel 5.4): https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions
- Ubuntu 22.04 (kernel 5.15 or 5.17): https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions
- Ubuntu 24.04 (kernel 6.8): https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions

Thus, until the Swift project moves from Amazon Linux 2 to Amazon Linux 2023 (6.1 kernel), I think we have to retain the original process termination path for compatibility with that distribution, as a fallback path. And we'll need that implementation for OpenBSD and other platforms anyways.

There's also Android. As you may know, the Android ecosystem is often running quite old OS versions compared to iOS, and I think only as of Android 13 (2022) does the OS guarantee kernel 5.4 or later. I'm not sure which minimum Android version we're planning to target in the Swift project, but that would be a good question for @finagolfin or @marcprux or someone else from the Android working group.

Perhaps you could do this by making `processFileDescriptor` an optional property and falling back to the pid based paths when it is `nil`?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -265,67 +294,70 @@ extension String {
...
+                let rc = epoll_ctl(
+                    storage.epollFileDescriptor,
+                    EPOLL_CTL_ADD,
+                    processIdentifier.processFileDescriptor,
+                    &event
+                )
+                if rc != 0 {
+                    let error = SubprocessError(
+                        code: .init(.failedToMonitorProcess),
+                        underlyingError: .init(rawValue: errno)
```
jakepetroules: `errno` is not safe across Swift function calls, so technically you need to save it to a local variable immediately after calling `epoll_ctl` since the initializer you pass to `code` could change it. Worth an audit across the entire codebase as well.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,77 +379,225 @@ private extension siginfo_t {
...
+        // Enter the monitor loop
+        monitorLoop: while true {
+            let eventCount = epoll_wait(
+                context.epollFileDescriptor,
+                &events,
+                CInt(events.count),
+                -1
+            )
+            if eventCount < 0 {
+                if errno == EINTR || errno == EAGAIN {
```
jakepetroules: suggestion: it might be worth introducing a helper function to handle EINTR/EAGAIN since it's such a common pattern throughout this codebase; see https://github.com/apple/swift-system/blob/6ee9a58c36ad98f4bd917a64d153dd211512e65d/Sources/System/Util.swift#L27 for example.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -118,13 +100,35 @@ extension Execution {
...
+            }
+        }
         let pid = shouldSendToProcessGroup ? -(processIdentifier.value) : processIdentifier.value
-        guard kill(pid, signal.rawValue) == 0 else {
-            throw SubprocessError(
-                code: .init(.failedToSendSignal(signal.rawValue)),
-                underlyingError: .init(rawValue: errno)
-            )
+
+        #if os(Linux)
```
jakepetroules: I think you may want `|| os(Android)` here too, I'm not sure if os(Linux) applies there.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
FranzBusch: I general one shouldn’t do any arbitrary outcalls while holding a lock. We have already seen continuation resumptions under locks lead to deadlocks since the runtime itself is holding locks and doing outcalls such as calling cancellation handlers.

I also don’t see a reason why we should do it here. It doesn’t prevent any potential races from what I can see.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: @jakepetroules thanks for the analysis. I did not know about Ubuntu's kernel release schedule I thought since 20.04 has 5.4 the later ones must have newer ones...

I do have a backup implementation which is to use `signalfd`. I initially opted for `pidfd` because it's more modern and precise. I guess we'll have to keep both

jakepetroules: Nice, I wasn't aware of `signalfd` (just read up on it). Still, it's Linux specific, so OpenBSD and some other platforms would probably have to continue to rely on the waitid-loop implementation.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -664,6 +504,10 @@ int _subprocess_fork_exec(
         // If we reached this point, something went wrong
         write_error_and_exit;
     } else {
+        int _pidfd = _pidfd_open(childPid);
```
jakepetroules: Could we use `clone` + `CLONE_PIDFD` (Linux 5.2) instead of `fork` + `pidfd_open`? Like FreeBSD's `pdfork`, this avoids races since combining the latter two functions is not atomic.

jakepetroules:
> This is not a bug in the existing implementation. It is a bug in the POSIX specification (and a bug in the program.)

That may be true, but it's more defensive against other parts of the program misbehaving, which seems like a good thing.

Including scenarios where zombies are being reaped correctly throughout the entire program, but maybe the body of one particular `Subprocess.run` call is stuck or otherwise taking an incredibly long time -- this implementation prevents that one process from holding up everything else.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
finagolfin: I'll try natively compiling this repo and then this pull on Android and running the tests, will let you know what I find.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,157 +310,13 @@ int _subprocess_spawn(
...
-    return ENOTSUP;
-#elif defined(__GLIBC__) || TARGET_OS_DARWIN || defined(__FreeBSD__) || defined(__ANDROID__) || defined(__musl__)
-    // Pre-standard posix_spawn_file_actions_addchdir_np version available in:
-    //  - Solaris 11.3 (October 2015)
-    //  - Glibc 2.29 (February 2019)
-    //  - macOS 10.15 (October 2019)
-    //  - musl 1.1.24 (October 2019)
-    //  - FreeBSD 13.1 (May 2022)
-    //  - Android 14 (API level 34) (October 2023)
-    return posix_spawn_file_actions_addchdir_np(file_actions, path);
```
grynspan, line 358: This will emit a deprecation warning as of *OS 26 since the standardized version has been added.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -477,26 +477,41 @@ extension Configuration {
     }
 }
 
-// Special keys used in Error's user dictionary
-extension String {
-    static let debugDescriptionErrorKey = "NSDebugDescription"
+// MARK: - ProcessIdentifier
+
+/// A platform independent identifier for a Subprocess.
+public struct ProcessIdentifier: Sendable, Hashable {
```
grynspan, line 483: Could you make this type move-only and incorporate the `close()` operation into `deinit`?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
grynspan: Can you clarify that comment?

cthielen: It's a reference to this earlier comment: https://github.com/swiftlang/swift-subprocess/pull/125#discussion_r2211709154

There are a few places in this PR where a continuation is resumed from within a lock.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
FranzBusch: Closing might involve closing FDs right? Which might be an asynchronous and throwing operation.

grynspan: That `close()` can fail at all is an unfortunate weird corner of POSIX that I personally tend to ignore, because a failure in `close()` other than `EINTR`/`EAGAIN` is basically non-recoverable. What are you even supposed to do? What can a user do to fix the problem? Generally nothing.

So I just about always just drop a `close()` failure on the floor. &lt;/hottake>

(As for asynchronous, it's a blocking operation in userland but it can't fail to make forward progress in this case because there's no network I/O involved unless we're doing something _really_ wonky.)

itingliu: We've had this discussion in another place ([here?](https://github.com/swiftlang/swift-subprocess/blob/5715ed49b0a5493cb24f3904dc2d9736c180d949/Sources/Subprocess/Configuration.swift#L689)). We ended up not calling `close`, but asserting that `close` has already been called in `deinit`.

But it does seem like a design people are going to reach for repeatedly. I wonder if we can put our thought process down somewhere.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: Could you elaborate on how would we use this protocol? The reason it's repeated is because on different platforms we have different sets of fields. I don't think having a protocol would solve this problem because we'd still need to offer different concrete types.

cthielen: File this one under premature optimization but a protocol might help if the shared, non-platform-specific code started relying on the existence of methods or attributes of `ProcessIdentifier`. I took a look and don't see any at the moment beyond `description`.

If there are more expectations requiring the various `ProcessIdentifier` definitions to stay in sync, it might be helpful to introduce a protocol, not because any given platform needs more than one concrete type, but because the protocol will serve as a contract to keep the implementations in sync with the expectations of the shared code.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: I also renamed it so it's more clear.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
FranzBusch: > That `close()` can fail at all is an unfortunate weird corner of POSIX that I personally tend to ignore, because a failure in `close()` other than `EINTR`/`EAGAIN` is basically non-recoverable. What are you even supposed to do? What can a user do to fix the problem? Generally nothing.

I agree that it is a weirdness; nevertheless, we need to handle it and most likely surface it to the user. We shouldn't just swallow those errors.

> (As for asynchronous, it's a blocking operation in userland but it can't fail to make forward progress in this case because there's no network I/O involved unless we're doing something _really_ wonky.)

This is not entirely true. If you are using io_uring you can asynchronously listen for the subprocess termination with `pidfd` and `signalfd` via io_uring. We need to account for changes in the underlying I/O system where closing can become asynchronous otherwise we will lock ourselves in a corner API-wise.

The only pattern that keeps us flexible is a `with-style` based approach.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: Ahh I see your point. IMO that's not a great use case here because there's no place we want to write `some ProcessIdentifierProtocol` as opposed to the concrete type. A protocol would help ensure all variants of `ProcessIdentifier` to have the `value` field but that's about it. As part of execution we still want to have the concrete version.

Sources/Subprocess/Execution.swift
iCharlesHu: I originally left it there because I wasn't sure if it is useful... I agree with you that it seems not that useful. I'll remove it in a dedicate PR since it's an API change.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Good call. Unfortunately it still needs to be conditionalized due to kernel version requirements.

iCharlesHu:
Added fallback implementation that uses `signal handler` on Linux 5.4 and below.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
cthielen: What about combining that with a `typealias`, e.g.

```swift
protocol SomeSubprocessConstruct {
    func doSomething()
    ...
}

// SomeFile+Windows.swift
#if canImport(WinSDK)
struct WindowsSubprocessConstruct: SomeSubprocessConstruct {
    func doSomething()
    ...
}
typealias SubprocessConstruct = WindowsSubprocessConstruct
#endif

// SomeFile+Darwin.swift
#if canImport(Darwin)
struct DarwinSubprocessConstruct: SomeSubprocessConstruct {
    func doSomething()
    ...
}
typealias SubprocessConstruct = DarwinSubprocessConstruct
#endif

// Some shared code
struct Execution {
    private var construct: SubprocessConstruct

    ...
    if construct.doSomething() { ... }
}
```

You know this domain way better than I do but as a naive reader, it feels to me like there should be a clearer separation between core logic and platform-specific details. Right now there's a bit of a mix: there are platform-specific files with similarly named methods as well as shared code with `#if canImport()`.

Given there's so much platform-specific code, it'd be great to better separate it out and keep platform-specific details more sequestered away.

iCharlesHu: > Given there's so much platform-specific code, it'd be great to better separate it out and keep platform-specific details more sequestered away.

I 100% agree with this statement. However, could you elaborate on how `typealias` would help in this case? Because currently we have

```swift
// SomeFile+Windows.swift
#if canImport(WinSDK)
struct ProcessIdentifier {
    func doSomething()
    // Additional stuff
}
#endif

// SomeFile+Darwin.swift
#if canImport(Darwin)
struct ProcessIdentifier {
    func doSomething()
    // additional stuff
}
#endif

// Some shared code
struct Execution {
    private var construct: ProcessIdentifier

    ...
    if construct.doSomething() { ... }
}
```

To me having a dedicated type name (which unfortunately is also an API change) doesn't necessarily make the code clearer because 1) the platform-specific `ProcessIdentifier` is already separated to different files; and 2) because of the typealias, you still ended up using the generic name `SubprocessConstruct` at the call site which makes it the same as just using the type directly.  To me with the suggestion you basically get the same code structure as today, but now there are additional type-specific names only used once in typealias `typealias SubprocessConstruct = DarwinSubprocessConstruct`.

cthielen: The `protocol` + `typealias` approach would provide a contract that all platform-specific implementations follow when interfacing with shared code. This would make it easier to ensure all platform-specific code was being updated when that contract changes as you work on shared code. As it stands today, one has to remember to update any platform-specific types that are shared by virtue of being spelled the same way. A `protocol` solidifies this idea.

A pattern like this could help keep all `#if canImport` statements in platform-specific files (e.g. out of Input.swift, Output.swift, API.swift, AsyncBufferSequence.swift, etc.), which would be a good way to measure whether the platform-specific code has been abstracted enough.

jakepetroules: As an alternative idea, you could define a protocol and retroactively make the ProcessIdentifier type conform to that protocol in the SubprocessTests target. That way, it serves as a type system level "test" in the way @cthielen is advocating for, but without adding complexity to the production code / library interface.

Sources/Subprocess/Execution.swift
jakepetroules: To be clear, it's _not_ an API change since it has `internal` visibility.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
jakepetroules: ProcessIdentifier is only ever provided as a parameter to a closure called within a with-style function, except for runDetached. If we remove the latter from the API, there's no place where ProcessIdentifier needs to be used as a handle that needs to be closed and therefore no reason it needs to be responsible for lifetime management.

Sources/Subprocess/Execution.swift
iCharlesHu: Ah good point. I think I was planning on making it an API I'm glad I didn't.

iCharlesHu:
Also pin swift-system to `1.5.0` since `1.6.0` breaks on `focal`.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: @cthielen thanks for the explanation. I see your point now. I added the `PlatformConformance` file in the test target with the protocol and conformance. I decided to go with @jakepetroules's suggestion because since we can't use `ProcessIdentifierProtocol` directly in the API, there's no point in making it a public type anyway.

Tests/SubprocessTests/SubprocessTests+Windows.swift
```diff
@@ -456,7 +456,9 @@ extension SubprocessWindowsTests {
 
 // MARK: - PlatformOption Tests
 extension SubprocessWindowsTests {
-    @Test(.enabled(if: SubprocessWindowsTests.hasAdminPrivileges()))
+    // Disabled until we investigate CI specific failures
+    // https://github.com/swiftlang/swift-subprocess/issues/128
+    @Test(.enabled(if: false))
```
jakepetroules: nit: there's a `disabled` API which accepts a reason string, as well as a bugtracker URL API, which you could use like this:

```suggestion
    @Test(.disabled("Disabled until we investigate CI specific failures"), .bug("https://github.com/swiftlang/swift-subprocess/issues/128"))
```

Sources/_SubprocessCShims/include/process_shims.h
```diff
@@ -73,6 +81,15 @@ int _shims_snprintf(
     char * _Nonnull str1,
     char * _Nonnull str2
 );
+
+int _pidfd_send_signal(int pidfd, int signal);
+
+// P_PIDFD is only defined on Linux Kernel 5.4 and above
+// Define our dummy value if it's not available
```
jakepetroules: nit: not a dummy value, that's _the_ value

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -303,157 +311,53 @@ int _subprocess_spawn(
...
-#else
-    // Standardized posix_spawn_file_actions_addchdir version (POSIX.1-2024, June 2024) available in:
-    //  - Solaris 11.4 (August 2018)
-    //  - NetBSD 10.0 (March 2024)
-    return posix_spawn_file_actions_addchdir(file_actions, path);
+// SYS_clone3 is only defined on Linux Kernel 5.3 and above
+// Define our dummy value if it's not available
+#ifndef SYS_clone3
+#define SYS_clone3 0xFFFF
```
jakepetroules, line 321: It's 435 on all archs except alpha (where it is 545). and I don't think anything in Swift has to care about alpha right now

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -543,11 +421,21 @@ int _subprocess_fork_exec(
         return errno;
     }
 
-    // Finally, fork
+    // Finally, fork / clone
+    int _pidfd = -1;
+    pid_t childPid = -1;
+    if (_linux_kernel_version_at_least(5, 3)) {
```
jakepetroules: Instead of doing a version check, why not just call clone3 directly and fall back to fork if it returns ENOSYS? This seems safer in case some distributions mess with the version number

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -664,6 +555,15 @@ int _subprocess_fork_exec(
         // If we reached this point, something went wrong
         write_error_and_exit;
     } else {
+        // On Linux 5.3 and lower, we have to fetch pidfd separately
+        // Newer Linux supports clone3 which returns pidfd directly
+        if (_linux_kernel_version_at_least(5, 3) == 0) {
+            _pidfd = _pidfd_open(childPid);
```
jakepetroules, line 552: as above, suggest just calling it without the version check and reacting accordingly if it returns ENOSYS

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -265,67 +295,115 @@ extension String {
...
+                    underlyingError: nil
+                )
+                return .failure(error)
+            case .failed(let error):
+                return .failure(error)
+            case .started(let storage):
+                // pidfd is only supported on Linux kernel 5.4 and above
+                // On older releases, use signalfd so we do not need
+                // to register anything with epoll
+                if _isLinuxKernelAtLeast(major: 5, minor: 4) {
```
jakepetroules: could we omit the version check and check for EPERM?

> EPERM  The target file fd does not support epoll.  This error can
>              occur if fd refers to, for example, a regular file or a
>              directory.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -347,80 +425,374 @@ private extension siginfo_t {
...
+            // Breakout the monitor loop if we received shutdown
+            // from the shutdownFD
+            if targetFileDescriptor == context.shutdownFileDescriptor {
+                var buf: UInt64 = 0
+                _ = _SubprocessCShims.read(context.shutdownFileDescriptor, &buf, MemoryLayout<UInt64>.size)
+                break monitorLoop
+            }
+
+            // P_PIDFD requires Linux Kernel 5.4 and above
+            if _isLinuxKernelAtLeast(major: 5, minor: 4) {
```
jakepetroules: can we just check if the pidfd is > 0 instead of doing a version check?

Tests/SubprocessTests/SubprocessTests+Windows.swift
iCharlesHu: Ahh neat. Thanks!

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: I read somewhere it's arch dependent (like you said) so I decided to go with 0xFFFF to ensure it will always error out if we accidentally use it on a system that doesn't support it.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: I think here we can prob whether waitid with `P_PIDFD` is supported, similar to the `clone3` prob

Sources/_SubprocessCShims/process_shims.c
jakepetroules: Yeah, that's probably easiest.

jakepetroules (approved):

jakepetroules:
One last thing:

On Linux we have `processFileDescriptor` which is a `PlatformFileDescriptor` (`CInt`). On Windows we have `processHandle` which is also a `PlatformFileDescriptor` (`HANDLE`).

Since on both platforms the `PlatformIdentifier` contains a field typed as `PlatformFileDescriptor`, and Windows `HANDLE`s are largely the conceptual equivalent of Unix file descriptors, could we please give them the same name across platforms? Either rename Linux's `processFileDescriptor` to `processHandle` or rename Windows's `processHandle` to `processFileDescriptor` (and probably `threadHandle` to `threadFileDescriptor` for consistency).

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -664,6 +540,15 @@ int _subprocess_fork_exec(
         // If we reached this point, something went wrong
         write_error_and_exit;
     } else {
+        // On Linux 5.3 and lower, we have to fetch pidfd separately
+        // Newer Linux supports clone3 which returns pidfd directly
+        if (_pidfd < 0) {
+            _pidfd = _pidfd_open(childPid);
+            if (_pidfd < 0) {
+                return errno;
```
jakepetroules: We have a child pid though. I think we need to call waitid here before returning errno?

jakepetroules:
> Windows has file descriptors too though.

True, but not for processes, and I don't think they ever would, because a HANDLE already provides all the same benefits. But maybe you're right that it's best to leave them separate.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
finagolfin: Got this repo to build on Android, but the tests keep getting killed, perhaps because of Android's process killer. I will look into it more later and let you know.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Ahh good catch!

PR #130, opened 2025-07-29
Remove the default collected output buffer limit and throw an error when the limit is reached

iCharlesHu:
As discussed in https://github.com/swiftlang/swift-subprocess/issues/44 , remove the default output limit since we can't really specify one. Also throw an error if output exits limit.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/44

### DIFF
--- a/Sources/Subprocess/Error.swift
+++ b/Sources/Subprocess/Error.swift
@@ -42,6 +42,7 @@ extension SubprocessError {
             case failedToMonitorProcess
             case streamOutputExceedsLimit(Int)
             case asyncIOFailed(String)
+            case outputBufferLimitExceeded(Int)
             // Signal
             case failedToSendSignal(Int32)
             // Windows Only
@@ -70,18 +71,20 @@ extension SubprocessError {
                 return 6
             case .asyncIOFailed(_):
                 return 7
-            case .failedToSendSignal(_):
+            case .outputBufferLimitExceeded(_):
                 return 8
-            case .failedToTerminate:
+            case .failedToSendSignal(_):
                 return 9
-            case .failedToSuspend:
+            case .failedToTerminate:
                 return 10
-            case .failedToResume:
+            case .failedToSuspend:
                 return 11
-            case .failedToCreatePipe:
+            case .failedToResume:
                 return 12
-            case .invalidWindowsPath(_):
+            case .failedToCreatePipe:
                 return 13
+            case .invalidWindowsPath(_):
+                return 14
             }
         }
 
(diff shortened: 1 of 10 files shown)


Sources/Subprocess/Error.swift
```diff
@@ -70,18 +71,20 @@ extension SubprocessError {
                 return 6
             case .asyncIOFailed(_):
                 return 7
-            case .failedToSendSignal(_):
+            case .outputBufferLimitExceeded(_):
```
parkera, line 74: After we set the API we should probably consider the error numbers to be frozen, in case people have written documentation about them or compare them to raw values.

Sources/Subprocess/Error.swift
```diff
@@ -113,6 +116,8 @@ extension SubprocessError: CustomStringConvertible, CustomDebugStringConvertible
             return "Failed to create output from current buffer because the output limit (\(limit)) was reached."
         case .asyncIOFailed(let reason):
             return "An error occurred within the AsyncIO subsystem: \(reason). Underlying error: \(self.underlyingError!)"
+        case .outputBufferLimitExceeded(let limit):
+            return "Output exceeds the limit (\(limit))"
```
parkera:
```suggestion
            return "Output exceeds the limit of (\(limit))"
```

limit of `x` - bytes, I assume?

Sources/Subprocess/IO/Output.swift
```diff
@@ -158,13 +158,25 @@ public struct BytesOutput: OutputProtocol {
         var result: [UInt8]? = nil
         #endif
         do {
-            result = try await AsyncIO.shared.read(from: diskIO, upTo: self.maxSize)
+            var maxLength = self.maxSize
+            if maxLength != .max {
+                // If we actually have a max length, attempt to read one
+                // more byte to determine whether output exceeds the limit
+                maxLength += 1
```
parkera, line 165: Does this mean that we've drained 1 more byte from the I/O than the caller expected?

Sources/Subprocess/IO/Output.swift
```diff
@@ -213,10 +225,9 @@ extension OutputProtocol where Self == FileDescriptorOutput {
 }
 
 extension OutputProtocol where Self == StringOutput<UTF8> {
-    /// Create a `Subprocess` output that collects output as
-    /// UTF8 String with 128kb limit.
-    public static var string: Self {
-        .init(limit: 128 * 1024, encoding: UTF8.self)
+    /// Create a `Subprocess` output that collects output as UTF8 String
```
parkera, line 228: Additional documentation about what the limit means (bytes, characters, something else) would be useful here.

Sources/Subprocess/IO/Output.swift
iCharlesHu: Yes. Unfortunately this is the only way to detect whether there's more data coming. If the caller was only expected N bytes, then the child process should only process N bytes with EOF at the end. If not then we considered this an error.

Sources/Subprocess/Error.swift
iCharlesHu: I agree. I'm only doing this because we haven't tagged 0.0.1 yet.

jakepetroules (approved):

Issue #100, opened 2025-06-29
Handle inheritance should be constrained to specific handles on Windows?


jakepetroules:
Much like we have the file descriptor inheritance problems on Windows, it sounds like we have a similar issue on Windows. From CreateProcess docs:

> By default, passing TRUE as the value of the bInheritHandles parameter causes all inheritable handles to be inherited by the new process. This can be problematic for applications which create processes from multiple threads simultaneously yet desire each process to inherit different handles.

This implies that we should use STARTUPINFOEXW and PROC_THREAD_ATTRIBUTE_LIST to pass the specific list of handles, or it sounds like we may run into over-inheritance therefore and security issues.

I'm not certain if this is what we need to do or how it might interact with STARTF_USESTDHANDLES, so would be great to get some input from @compnerd about this.

PR #133, opened 2025-08-08
Unify tests on all supported platforms to ensure consistent behavior and add more tests.

iCharlesHu:
This PR refactors and consolidates tests across all platforms and introduces numerous additional tests.

Now, tests are categorized into four main types:
- `IntegrationTests`: This serves as the primary test file that evaluates the end-to-end functionality of Subprocess by executing child processes.
- `{Unix|Darwin|Linux|Windows}Tests`: These tests are platform-specific and cater to the unique characteristics of each operating system.
- `AsyncIOTests`: These tests are dedicated to unit testing the `AsyncIO` functionality specifically.
- `ProcessMonitoringTests`: These tests are dedicated to unit testing `func monitorProcessTermination` specifically

These newly established test categories have revealed a few bugs in the implementation, which are also addressed in this PR:
- Moved the `monitorProcessTermination` function inside the `withAsyncTaskCleanupHandler`. This adjustment ensures that the teardown sequence can still be executed even if the `body` closure terminates prematurely, but the child process remains stuck.
- On Windows, we’ve modified the code to use `STARTUPINFOEXW` to ensure that only the explicitly provided file descriptors are inheritable.
- Make sure `AsyncIO.shutdown` is idempotent.
- Implemented executable path searching for Windows.
- Implemented arg0 override for Windows

Resolves: https://github.com/swiftlang/swift-subprocess/issues/97
Resolves: https://github.com/swiftlang/swift-subprocess/issues/100

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Linux.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Linux.swift
@@ -357,59 +357,75 @@ extension AsyncIO {
         )
         var readLength: Int = 0
         let signalStream = self.registerFileDescriptor(fileDescriptor, for: .read)
-        /// Outer loop: every iteration signals we are ready to read more data
-        for try await _ in signalStream {
-            /// Inner loop: repeatedly call `.read()` and read more data until:
-            /// 1. We reached EOF (read length is 0), in which case return the result
-            /// 2. We read `maxLength` bytes, in which case return the result
-            /// 3. `read()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
-            ///     this case we `break` out of the inner loop and wait `.read()` to be
-            ///     ready by `await`ing the next signal in the outer loop.
-            while true {
-                let bytesRead = resultBuffer.withUnsafeMutableBufferPointer { bufferPointer in
-                    // Get a pointer to the memory at the specified offset
-                    let targetCount = bufferPointer.count - readLength
 
-                    let offsetAddress = bufferPointer.baseAddress!.advanced(by: readLength)
+        do {
+            /// Outer loop: every iteration signals we are ready to read more data
+            for try await _ in signalStream {
+                /// Inner loop: repeatedly call `.read()` and read more data until:
+                /// 1. We reached EOF (read length is 0), in which case return the result
+                /// 2. We read `maxLength` bytes, in which case return the result
+                /// 3. `read()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
+                ///     this case we `break` out of the inner loop and wait `.read()` to be
+                ///     ready by `await`ing the next signal in the outer loop.
+                while true {
+                    let bytesRead = resultBuffer.withUnsafeMutableBufferPointer { bufferPointer in
+                        // Get a pointer to the memory at the specified offset
+                        let targetCount = bufferPointer.count - readLength
 
-                    // Read directly into the buffer at the offset
-                    return _subprocess_read(fileDescriptor.rawValue, offsetAddress, targetCount)
-                }
-                if bytesRead > 0 {
-                    // Read some data
-                    readLength += bytesRead
-                    if maxLength == .max {
-                        // Grow resultBuffer if needed
-                        guard Double(readLength) > 0.8 * Double(resultBuffer.count) else {
-                            continue
+                        let offsetAddress = bufferPointer.baseAddress!.advanced(by: readLength)
+
+                        // Read directly into the buffer at the offset
+                        return _subprocess_read(fileDescriptor.rawValue, offsetAddress, targetCount)
+                    }
+                    let capturedErrno = errno
+                    if bytesRead > 0 {
+                        // Read some data
+                        readLength += bytesRead
+                        if maxLength == .max {
+                            // Grow resultBuffer if needed
+                            guard Double(readLength) > 0.8 * Double(resultBuffer.count) else {
+                                continue
+                            }
+                            resultBuffer.append(
+                                contentsOf: Array(repeating: 0, count: resultBuffer.count)
+                            )
+                        } else if readLength >= maxLength {
+                            // When we reached maxLength, return!
+                            try self.removeRegistration(for: fileDescriptor)
+                            return resultBuffer
                         }
-                        resultBuffer.append(
-                            contentsOf: Array(repeating: 0, count: resultBuffer.count)
-                        )
-                    } else if readLength >= maxLength {
-                        // When we reached maxLength, return!
+                    } else if bytesRead == 0 {
+                        // We reached EOF. Return whatever's left
                         try self.removeRegistration(for: fileDescriptor)
+                        guard readLength > 0 else {
+                            return nil
+                        }
+                        resultBuffer.removeLast(resultBuffer.count - readLength)
                         return resultBuffer
-                    }
-                } else if bytesRead == 0 {
-                    // We reached EOF. Return whatever's left
-                    try self.removeRegistration(for: fileDescriptor)
-                    guard readLength > 0 else {
-                        return nil
-                    }
-                    resultBuffer.removeLast(resultBuffer.count - readLength)
-                    return resultBuffer
-                } else {
-                    if self.shouldWaitForNextSignal(with: errno) {
-                        // No more data for now wait for the next signal
-                        break
                     } else {
-                        // Throw all other errors
-                        try self.removeRegistration(for: fileDescriptor)
-                        throw SubprocessError.UnderlyingError(rawValue: errno)
+                        if self.shouldWaitForNextSignal(with: capturedErrno) {
+                            // No more data for now wait for the next signal
+                            break
+                        } else {
+                            // Throw all other errors
+                            try self.removeRegistration(for: fileDescriptor)
+                            throw SubprocessError(
+                                code: .init(.failedToReadFromSubprocess),
+                                underlyingError: .init(rawValue: capturedErrno)
+                            )
+                        }
                     }
                 }
             }
+        } catch {
+            // Reset error code to .failedToRead to match other platforms
+            guard let originalError = error as? SubprocessError else {
+                throw error
+            }
+            throw SubprocessError(
+                code: .init(.failedToReadFromSubprocess),
+                underlyingError: originalError.underlyingError
+            )
         }
         resultBuffer.removeLast(resultBuffer.count - readLength)
         return resultBuffer
@@ -432,37 +448,52 @@ extension AsyncIO {
         let fileDescriptor = diskIO.channel
         let signalStream = self.registerFileDescriptor(fileDescriptor, for: .write)
         var writtenLength: Int = 0
-        /// Outer loop: every iteration signals we are ready to read more data
-        for try await _ in signalStream {
-            /// Inner loop: repeatedly call `.write()` and write more data until:
-            /// 1. We've written bytes.count bytes.
-            /// 3. `.write()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
-            ///     this case we `break` out of the inner loop and wait `.write()` to be
-            ///     ready by `await`ing the next signal in the outer loop.
-            while true {
-                let written = bytes.withUnsafeBytes { ptr in
-                    let remainingLength = ptr.count - writtenLength
-                    let startPtr = ptr.baseAddress!.advanced(by: writtenLength)
-                    return _subprocess_write(fileDescriptor.rawValue, startPtr, remainingLength)
-                }
-                if written > 0 {
-                    writtenLength += written
-                    if writtenLength >= bytes.count {
-                        // Wrote all data
-                        try self.removeRegistration(for: fileDescriptor)
-                        return writtenLength
+        do {
+            /// Outer loop: every iteration signals we are ready to read more data
+            for try await _ in signalStream {
+                /// Inner loop: repeatedly call `.write()` and write more data until:
+                /// 1. We've written bytes.count bytes.
+                /// 3. `.write()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
+                ///     this case we `break` out of the inner loop and wait `.write()` to be
+                ///     ready by `await`ing the next signal in the outer loop.
+                while true {
+                    let written = bytes.withUnsafeBytes { ptr in
+                        let remainingLength = ptr.count - writtenLength
+                        let startPtr = ptr.baseAddress!.advanced(by: writtenLength)
+                        return _subprocess_write(fileDescriptor.rawValue, startPtr, remainingLength)
                     }
-                } else {
-                    if self.shouldWaitForNextSignal(with: errno) {
-                        // No more data for now wait for the next signal
-                        break
+                    let capturedErrno = errno
+                    if written > 0 {
+                        writtenLength += written
+                        if writtenLength >= bytes.count {
+                            // Wrote all data
+                            try self.removeRegistration(for: fileDescriptor)
+                            return writtenLength
+                        }
                     } else {
-                        // Throw all other errors
-                        try self.removeRegistration(for: fileDescriptor)
-                        throw SubprocessError.UnderlyingError(rawValue: errno)
+                        if self.shouldWaitForNextSignal(with: capturedErrno) {
+                            // No more data for now wait for the next signal
+                            break
+                        } else {
+                            // Throw all other errors
+                            try self.removeRegistration(for: fileDescriptor)
+                            throw SubprocessError(
+                                code: .init(.failedToWriteToSubprocess),
+                                underlyingError: .init(rawValue: capturedErrno)
+                            )
+                        }
                     }
                 }
             }
+        } catch {
+            // Reset error code to .failedToWrite to match other platforms
+            guard let originalError = error as? SubprocessError else {
+                throw error
+            }
+            throw SubprocessError(
+                code: .init(.failedToWriteToSubprocess),
+                underlyingError: originalError.underlyingError
+            )
         }
         return 0
     }
@@ -478,37 +509,52 @@ extension AsyncIO {
         let fileDescriptor = diskIO.channel
         let signalStream = self.registerFileDescriptor(fileDescriptor, for: .write)
         var writtenLength: Int = 0
-        /// Outer loop: every iteration signals we are ready to read more data
-        for try await _ in signalStream {
-            /// Inner loop: repeatedly call `.write()` and write more data until:
-            /// 1. We've written bytes.count bytes.
-            /// 3. `.write()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
-            ///     this case we `break` out of the inner loop and wait `.write()` to be
-            ///     ready by `await`ing the next signal in the outer loop.
-            while true {
-                let written = span.withUnsafeBytes { ptr in
-                    let remainingLength = ptr.count - writtenLength
-                    let startPtr = ptr.baseAddress!.advanced(by: writtenLength)
-                    return _subprocess_write(fileDescriptor.rawValue, startPtr, remainingLength)
-                }
-                if written > 0 {
-                    writtenLength += written
-                    if writtenLength >= span.byteCount {
-                        // Wrote all data
-                        try self.removeRegistration(for: fileDescriptor)
-                        return writtenLength
+        do {
+            /// Outer loop: every iteration signals we are ready to read more data
+            for try await _ in signalStream {
+                /// Inner loop: repeatedly call `.write()` and write more data until:
+                /// 1. We've written bytes.count bytes.
+                /// 3. `.write()` returns -1 and sets `errno` to `EAGAIN` or `EWOULDBLOCK`. In
+                ///     this case we `break` out of the inner loop and wait `.write()` to be
+                ///     ready by `await`ing the next signal in the outer loop.
+                while true {
+                    let written = span.withUnsafeBytes { ptr in
+                        let remainingLength = ptr.count - writtenLength
+                        let startPtr = ptr.baseAddress!.advanced(by: writtenLength)
+                        return _subprocess_write(fileDescriptor.rawValue, startPtr, remainingLength)
                     }
-                } else {
-                    if self.shouldWaitForNextSignal(with: errno) {
-                        // No more data for now wait for the next signal
-                        break
+                    let capturedErrno = errno
+                    if written > 0 {
+                        writtenLength += written
+                        if writtenLength >= span.byteCount {
+                            // Wrote all data
+                            try self.removeRegistration(for: fileDescriptor)
+                            return writtenLength
+                        }
                     } else {
-                        // Throw all other errors
-                        try self.removeRegistration(for: fileDescriptor)
-                        throw SubprocessError.UnderlyingError(rawValue: errno)
+                        if self.shouldWaitForNextSignal(with: capturedErrno) {
+                            // No more data for now wait for the next signal
+                            break
+                        } else {
+                            // Throw all other errors
+                            try self.removeRegistration(for: fileDescriptor)
+                            throw SubprocessError(
+                                code: .init(.failedToWriteToSubprocess),
+                                underlyingError: .init(rawValue: capturedErrno)
+                            )
+                        }
                     }
                 }
             }
+        } catch {
+            // Reset error code to .failedToWrite to match other platforms
+            guard let originalError = error as? SubprocessError else {
+                throw error
+            }
+            throw SubprocessError(
+                code: .init(.failedToWriteToSubprocess),
+                underlyingError: originalError.underlyingError
+            )
         }
         return 0
     }
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -63,14 +63,14 @@ extension Configuration {
         // automatic PATH lookup so we only need to call `CreateProcessW` once. However, if
         // user wants to override executable path in arguments, we have to use `lpApplicationName`
         // to specify the executable path. In this case, manually loop over all possible paths.
-        let possibleExecutablePaths: Set<String>
+        let possibleExecutablePaths: _OrderedSet<String>
         if _fastPath(self.arguments.executablePathOverride == nil) {
             // Fast path: we can rely on `CreateProcessW`'s built in Path searching
             switch self.executable.storage {
             case .executable(let executable):
-                possibleExecutablePaths = Set([executable])
+                possibleExecutablePaths = _OrderedSet([executable])
             case .path(let path):
-                possibleExecutablePaths = Set([path.string])
+                possibleExecutablePaths = _OrderedSet([path.string])
             }
         } else {
             // Slow path: user requested arg0 override, therefore we must manually
@@ -255,14 +255,14 @@ extension Configuration {
         // automatic PATH lookup so we only need to call `CreateProcessW` once. However, if
         // user wants to override executable path in arguments, we have to use `lpApplicationName`
         // to specify the executable path. In this case, manually loop over all possible paths.
-        let possibleExecutablePaths: Set<String>
+        let possibleExecutablePaths: _OrderedSet<String>
         if _fastPath(self.arguments.executablePathOverride == nil) {
             // Fast path: we can rely on `CreateProcessW`'s built in Path searching
             switch self.executable.storage {
             case .executable(let executable):
-                possibleExecutablePaths = Set([executable])
+                possibleExecutablePaths = _OrderedSet([executable])
             case .path(let path):
-                possibleExecutablePaths = Set([path.string])
+                possibleExecutablePaths = _OrderedSet([path.string])
             }
         } else {
             // Slow path: user requested arg0 override, therefore we must manually
@@ -771,12 +771,12 @@ extension Executable {
     /// https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw
     internal func possibleExecutablePaths(
         withPathValue pathValue: String?
-    ) -> Set<String> {
+    ) -> _OrderedSet<String> {
         func insertExecutableAddingExtension(
             _ name: String,
             currentPath: String,
-            pathExtensions: Set<String>,
-            storage: inout Set<String>
+            pathExtensions: _OrderedSet<String>,
+            storage: inout _OrderedSet<String>
         ) {
             let fullPath = FilePath(currentPath).appending(name)
             if !name.hasExtension() {
@@ -792,10 +792,10 @@ extension Executable {
 
         switch self.storage {
         case .executable(let name):
-            var possiblePaths: Set<String> = []
+            var possiblePaths: _OrderedSet<String> = .init()
             let currentEnvironmentValues = Environment.currentEnvironmentValues()
             // If `name` does not include extensions, we need to try these extensions
-            var pathExtensions: Set<String> = Set(["com", "exe", "cmd", "bat"])
+            var pathExtensions: _OrderedSet<String> = _OrderedSet(["com", "exe", "bat", "cmd"])
             if let extensionList = currentEnvironmentValues["PATHEXT"] {
                 for var ext in extensionList.split(separator: ";") {
                     ext.removeFirst(1)
@@ -890,7 +890,7 @@ extension Executable {
             }
             return possiblePaths
         case .path(let path):
-            return Set([path.string])
+            return _OrderedSet([path.string])
         }
     }
 }
--- a/Tests/SubprocessTests/AsyncIOTests.swift
+++ b/Tests/SubprocessTests/AsyncIOTests.swift
@@ -146,20 +146,32 @@ extension SubprocessAsyncIOTests {
 
         try writeChannel.safelyClose()
 
-        do {
+        await #expect {
             _ = try await AsyncIO.shared.write([100], to: writeChannel)
-            Issue.record("Expected write to closed pipe to throw an error")
-        } catch {
+        } throws: { error in
             guard let subprocessError = error as? SubprocessError else {
-                Issue.record("Expecting SubprocessError, but got \(error)")
-                return
+                return false
             }
-            #if canImport(Darwin) || os(FreeBSD) || os(OpenBSD)
+            guard subprocessError.code == .init(.failedToWriteToSubprocess) else {
+                return false
+            }
+
+            #if os(Windows)
+            #expect(subprocessError.underlyingError == .init(rawValue: DWORD(ERROR_INVALID_HANDLE)))
+            #elseif canImport(Darwin) || os(FreeBSD) || os(OpenBSD)
             #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
-            #elseif os(Linux) || os(Android)
-            #expect(subprocessError.underlyingError == .init(rawValue: EBADF))
+            #else
+            // On Linux, depending on timing, either epoll_ctl or write
+            // could throw error first
+            #expect(
+                subprocessError.underlyingError == .init(rawValue: EBADF) ||
+                subprocessError.underlyingError == .init(rawValue: EINVAL) ||
+                subprocessError.underlyingError == .init(rawValue: EPERM)
+            )
             #endif
+            return true
         }
+
     }
 
     @Test func testReadFromClosedPipe() async throws {
@@ -172,19 +184,30 @@ extension SubprocessAsyncIOTests {
 
         try readChannel.safelyClose()
 
-        do {
+        await #expect {
             _ = try await AsyncIO.shared.read(from: readChannel, upTo: .max)
-            Issue.record("Expected write to closed pipe to throw an error")
-        } catch {
+        } throws: { error in
             guard let subprocessError = error as? SubprocessError else {
-                Issue.record("Expecting SubprocessError, but got \(error)")
-                return
+                return false
             }
-            #if canImport(Darwin) || os(FreeBSD) || os(OpenBSD)
+            guard subprocessError.code == .init(.failedToReadFromSubprocess) else {
+                return false
+            }
+
+            #if os(Windows)
+            #expect(subprocessError.underlyingError == .init(rawValue: DWORD(ERROR_INVALID_HANDLE)))
+            #elseif canImport(Darwin) || os(FreeBSD) || os(OpenBSD)
             #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
-            #elseif os(Linux) || os(Android)
-            #expect(subprocessError.underlyingError == .init(rawValue: EBADF))
+            #else
+            // On Linux, depending on timing, either epoll_ctl or read
+            // could throw error first
+            #expect(
+                subprocessError.underlyingError == .init(rawValue: EBADF) ||
+                subprocessError.underlyingError == .init(rawValue: EINVAL) ||
+                subprocessError.underlyingError == .init(rawValue: EPERM)
+            )
             #endif
+            return true
         }
     }
 
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -998,3 +998,39 @@ internal func withAsyncTaskCleanupHandler<Result>(
         }
     }
 }
+
+internal struct _OrderedSet<Element: Hashable & Sendable>: Hashable, Sendable {
+    private var elements: [Element]
+    private var hashValueSet: Set<Int>
+
+    internal init() {
+        self.elements = []
+        self.hashValueSet = Set()
+    }
+
+    internal init(_ arrayValue: [Element]) {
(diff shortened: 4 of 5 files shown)


Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -204,11 +205,15 @@ final class AsyncIO: Sendable {
         }
     }
 
-    private func shutdown() {
+    internal func shutdown() {
         guard case .success(let currentState) = self.state else {
             return
         }
 
+        guard self.shutdownFlag.add(1, ordering: .sequentiallyConsistent).newValue == 1 else {
```
jakepetroules, line 209: Should this be >=1 instead of ==1? I think as written this will run the shutdown procedure n-1 times for n calls.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -159,12 +162,17 @@ final class AsyncIO: @unchecked Sendable {
...
 
-    private func shutdown() {
-        // Post status to shutdown HANDLE
+    internal func shutdown() {
         guard case .success(let ioPort) = ioCompletionPort,
               case .success(let monitorThreadHandle) = monitorThread else {
             return
         }
+        // Make sure we don't shutdown the same instance twice
+        guard self.shutdownFlag.add(1, ordering: .relaxed).newValue == 1 else {
```
jakepetroules, line 171: needs to be >=1?

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -772,26 +803,33 @@ extension Configuration {
...
         errorRead errorReadFileDescriptor: borrowing IODescriptor?,
         errorWrite errorWriteFileDescriptor: borrowing IODescriptor?,
-    ) throws -> STARTUPINFOW {
-        var info: STARTUPINFOW = STARTUPINFOW()
-        info.cb = DWORD(MemoryLayout<STARTUPINFOW>.size)
-        info.dwFlags |= DWORD(STARTF_USESTDHANDLES)
+        _ body: (UnsafeMutablePointer<STARTUPINFOEXW>) throws -> Result
+    ) rethrows -> Result {
+        var info: STARTUPINFOEXW = STARTUPINFOEXW()
+        info.StartupInfo.cb = DWORD(MemoryLayout<STARTUPINFOEXW>.size)
```
jakepetroules: nit: can use `MemoryLayout.size(ofValue: info)` to avoid repeating the type name (seems safer)

Sources/Subprocess/Teardown.swift
```diff
@@ -194,6 +198,19 @@ extension Execution {
...
+    private func isStillAlive() -> Bool {
+        // Non-blockingly check whether the current execution has already exited
+        // Note here we do NOT want to reap the exit status because we are still
+        // running monitorProcessTermination()
+        #if os(Windows)
+        var exitCode: DWORD = 0
+        GetExitCodeProcess(self.processInformation.hProcess, &exitCode)
+        return exitCode == STILL_ACTIVE
+        #else
+        return kill(self.processIdentifier.value, 0) == 0
```
jakepetroules, line 209: Any particular reason to use kill 0 instead of waitpid with WNOHANG+WNOWAIT?

Sources/Subprocess/Teardown.swift
```diff
@@ -194,6 +198,19 @@ extension Execution {
             }
         }
     }
+
+    private func isStillAlive() -> Bool {
```
jakepetroules: Could already have exited by the time you return, perhaps a better name is "isPotentiallyStillAlive"? The only definitive case we can know is that it _did_ exit.

Sources/Subprocess/Teardown.swift
```diff
@@ -194,6 +198,19 @@ extension Execution {
...
+        // Non-blockingly check whether the current execution has already exited
+        // Note here we do NOT want to reap the exit status because we are still
+        // running monitorProcessTermination()
+        #if os(Windows)
+        var exitCode: DWORD = 0
+        GetExitCodeProcess(self.processInformation.hProcess, &exitCode)
+        return exitCode == STILL_ACTIVE
```
jakepetroules: I'm slightly skeptical of this approach because technically a user _could_ exit with this exit code if they aren't aware of the Windows conventions here, so this could return a false positive. It would be safest to wrap up the WaitForSingleObject call on the process handle in an abstraction which sets a thread-safe flag indicating when the process has exited, that you can check from here.

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+
+@Suite("Subprocess.AsyncIO Unit Tests", .serialized)
+struct SubprocessAsyncIOTests { }
+
+// MARK: - Basic Functionality Tests
+extension SubprocessAsyncIOTests {
+    @Test func testBasicReadWrite() async throws {
+        let testData = randomData(count: 1024)
+        try await runReadWriteTest { readIO, readTestBed in
+            let readData = try await readIO.read(from: readTestBed.ioChannel, upTo: .max)
```
jakepetroules: #require avoids the force unwrap:

```suggestion
            let readData = try await #require(readIO.read(from: readTestBed.ioChannel, upTo: .max))
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+        for _ in 0 ..< 10 {
+            // Generate some that are longer than buffer size
+            _chunks.append(randomData(count: Int.random(in: Subprocess.readBufferSize ..< Subprocess.readBufferSize * 3)))
+        }
+        _chunks.shuffle()
+        let chunks = _chunks
+        try await runReadWriteTest { readIO, readTestBed in
+            for expectedChunk in chunks {
+                let readData = try await readIO.read(from: readTestBed.ioChannel, upTo: expectedChunk.count)
```
jakepetroules: Avoids force unwraps:

```suggestion
                let readData = try await #require(readIO.read(from: readTestBed.ioChannel, upTo: expectedChunk.count))
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+            let written = try await writeIO.write([], to: writeTestBed.ioChannel)
+            #expect(written == 0)
+            try await writeTestBed.finish()
+        }
+    }
+
+    @Test func testLargeReadWrite() async throws {
+        let testData = randomData(count: 1024 * 1024)
+        try await runReadWriteTest { readIO, readTestBed in
+            let readData = try await readIO.read(from: readTestBed.ioChannel, upTo: .max)
```
jakepetroules: Avoid force unwrap:

```suggestion
            let readData = try await #require(readIO.read(from: readTestBed.ioChannel, upTo: .max))
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+
+// MARK: - Error Handling Tests
+extension SubprocessAsyncIOTests {
+    @Test func testWriteToClosedPipe() async throws {
+        var pipe = try CreatedPipe(closeWhenDone: true, purpose: .input)
+        var writeChannel = pipe.writeFileDescriptor()!.createIOChannel()
+        var readChannel = pipe.readFileDescriptor()!.createIOChannel()
```
jakepetroules, line 142:
```suggestion
        var readChannel = try #require(pipe.readFileDescriptor()).createIOChannel()
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+}
+
+
+// MARK: - Error Handling Tests
+extension SubprocessAsyncIOTests {
+    @Test func testWriteToClosedPipe() async throws {
+        var pipe = try CreatedPipe(closeWhenDone: true, purpose: .input)
+        var writeChannel = pipe.writeFileDescriptor()!.createIOChannel()
```
jakepetroules, line 141:
```suggestion
        var writeChannel = try #require(pipe.writeFileDescriptor()).createIOChannel()
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+        } catch {
+            guard let subprocessError = error as? SubprocessError else {
+                Issue.record("Expecting SubprocessError, but got \(error)")
+                return
+            }
+            #if canImport(Darwin)
+            #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
+            #elseif os(Linux)
+            #expect(subprocessError.underlyingError == .init(rawValue: EBADF))
+            #endif
```
jakepetroules, line 171: Can we add #else unknown platform so this doesn't silently fail to assert any expectations on new platforms?

Also, I bet you can add `os(FreeBSD) || os(OpenBSD)` to the Darwin clause since the BSDs should all be the same w.r.t. this behavior.

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+            _ = try await AsyncIO.shared.write([100], to: writeChannel)
+            Issue.record("Expected write to closed pipe to throw an error")
+        } catch {
+            guard let subprocessError = error as? SubprocessError else {
+                Issue.record("Expecting SubprocessError, but got \(error)")
+                return
+            }
+            #if canImport(Darwin)
+            #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
+            #elseif os(Linux)
```
jakepetroules: Needs Android check too:

```suggestion
            #elseif os(Linux) || os(Android)
```

(Only one OS condition is active at a time (with the exception of the xrOS alias for visionOS). See https://github.com/swiftlang/swift/blob/13c3ed900bc67950603f59a69c70c1c21404b3a4/lib/Basic/LangOptions.cpp#L513)

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+            #expect(subprocessError.underlyingError == .init(rawValue: EBADF))
+            #endif
```
jakepetroules, line 209: same Android/FreeBSD/OpenBSD comments as above

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    @Test func testBinaryDataWithNullBytes() async throws {
+        let binaryData: [UInt8] = [0x00, 0x01, 0x02, 0x00, 0xFF, 0x00, 0xFE, 0xFD]
+        try await runReadWriteTest { readIO, readTestBed in
+            let readData = try await readIO.read(from: readTestBed.ioChannel, upTo: .max)
```
jakepetroules:
```suggestion
            let readData = try await #require(readIO.read(from: readTestBed.ioChannel, upTo: .max))
```

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -0,0 +1,262 @@
...
+            let writeIO = AsyncIO()
+
+            group.addTask {
+                var readIOContainer: IOChannel? = readBox.take()
+                let readTestBed = TestBed(ioChannel: readIOContainer.take()!)
+                try await reader(readIO, readTestBed)
+            }
+            group.addTask {
+                var writeIOContainer: IOChannel? = writeBox.take()
+                let writeTestBed = TestBed(ioChannel: writeIOContainer.take()!)
```
jakepetroules, line 265:
```suggestion
                let writeTestBed = TestBed(ioChannel: try #require(writeIOContainer.take()))
```

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+
+        // Simple test to make sure we can find a common utility
+        let result = try await _run(
+            setup,
+            input: .none,
+            output: .string(limit: 32),
+            error: .discarded
+        )
+        #expect(result.terminationStatus.isSuccess)
+        // rdar://138670128
```
jakepetroules, line 64: Add the public reference as well? swiftlang/swift#77235

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+        #endif
+        let expected = FileManager.default.currentDirectoryPath
+        let result = try await _run(
+            setup,
+            input: .none,
+            output: .string(limit: .max),
+            error: .discarded
+        )
+        #expect(result.terminationStatus.isSuccess)
+        // rdar://138670128
```
jakepetroules, line 114: same comment about GitHub issue #

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+                return
+            }
+            #expect(subprocessError.code == .init(.executableNotFound("do-not-exist")))
+        }
+    }
+
+    @Test func testExecutableAtPath() async throws {
+        #if os(Windows)
+        let setup = TestSetup(
+            executable: .path(#"C:\Windows\System32\cmd.exe"#),
```
jakepetroules: Not guaranteed to be at that path. Suggest reading the `ComSpec` environment variable, which contains the cmd.exe path.

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+        // rdar://138670128
+        let output = result.standardOutput?
+            .trimmingNewLineAndQuotes()
+        #expect(
+            output == message
+        )
+    }
+
+    #if !os(Windows)
+    // Windows does not support argument 0 override
```
jakepetroules: I don't understand why it doesn't though; CreateProcessW has the lpApplicationName parameter, and you actually have most of it hooked up to `executablePathOverride` already. The API is just ifdef'ed out.

I actually have a use case in mind for this.

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+            output: .string(limit: .max),
+            error: .discarded
+        )
+        #expect(result.terminationStatus.isSuccess)
+        let pathValue = try #require(result.standardOutput)
+        #if os(Windows)
+        // As a sanity check, make sure there's
+        // `C:\Windows\system32` in PATH
+        // since we inherited the environment variables
+        #expect(pathValue.contains("C:\\Windows\\system32"))
```
jakepetroules: same comments as above about using ComSpec

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+        .enabled(if: ProcessInfo.processInfo.environment["SystemRoot"] != nil)
+    )
+    func testEnvironmentCustom() async throws {
+        #if os(Windows)
+        let setup = TestSetup(
+            executable: .path(#"C:\Windows\System32\cmd.exe"#),
+            arguments: ["/c", "set"],
+            environment: .custom([
+                "Path": "C:\\Windows\\system32;C:\\Windows",
+                "ComSpec": "C:\\Windows\\System32\\cmd.exe",
```
jakepetroules: Can use this to compute the system32 directory:

```swift
private func _url(for id: KNOWNFOLDERID) -> URL? {
    var pszPath: PWSTR?
    let hr: HRESULT = withUnsafePointer(to: id) { id in
        SHGetKnownFolderPath(id, DWORD(WinSDK.KF_FLAG_DEFAULT.rawValue), nil, &pszPath)
    }
    guard hr >= 0 else { return nil }
    defer { CoTaskMemFree(pszPath) }
    return URL(filePath: String(decodingCString: pszPath!, as: UTF16.self), directoryHint: .isDirectory)
}

_url(for: FOLDERID_System) // C:\\Windows\\System32
```

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+                group.addTask {
+                    _ = try await standardInputWriter.write(Array(expected))
+                    try await standardInputWriter.finish()
+                    return nil
+                }
+
+                var buffer: Data!
+                while let result = try await group.next() {
+                    if let result: Data = result {
+                        buffer = result
```
jakepetroules:
```suggestion
                        return result
```

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+                    }
+                }
+                return buffer
```
jakepetroules:
```suggestion
                preconditionFailure("unreachable")
```

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+        #endif
+
+        do {
+            _ = try await _run(
+                setup,
+                input: .none,
+                output: .string(limit: 16),
+                error: .discarded
+            )
+            Issue.record("Expected to throw")
```
jakepetroules: For this you can use:

```swift
#expect(throws: Error.self) {
    // code...
}
```

Tests/SubprocessTests/UnixTests.swift
```diff
@@ -0,0 +1,537 @@
...
+                        platformOptions: platformOptions,
+                        output: .string(limit: .max),
+                        error: .fileDescriptor(.standardError, closeAfterSpawningProcess: false)
+                    )
+                    #expect(result.terminationStatus.isSuccess)
+                    let output = try #require(result.standardOutput).trimmingNewLineAndQuotes()
+                    let grandChildPid = try #require(pid_t(output))
+                    // Make sure the grand child `/usr/bin/yes` actually exited
+                    // This is unfortunately racy because the pid isn't immediately invalided
+                    // once the exit exits. Allow a few failures and delay to counter this
```
jakepetroules:
```suggestion
                    // once `kill` returns. Allow a few failures and delay to counter this
```

?

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+        #else
+        let setup = TestSetup(
+            executable: .path("/usr/bin/tee"),
+            arguments: ["/dev/stderr"]
+        )
+        #endif
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            for _ in 0 ..< 8 {
+                group.addTask {
+                    // This invocation specifically requires bash semantics; sh (on FreeBSD at least) does not consistently support -s in this way
```
jakepetroules: Leftover comment? nothing here is using bash

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -226,7 +231,6 @@ final class AsyncIO: Sendable {
         }
     }
 
-
     private func registerFileDescriptor(
```
jakepetroules, line 230: Should registerFileDescriptor precondition that the fd is not already in the `_registration` map?

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2279 @@
...
+                }
+                return buffer
+            }
+
+        }
+        #expect(result.terminationStatus.isSuccess)
+        #expect(result.value == expected)
+    }
+
+    @Test func stressTestWithLittleOutput() async throws {
```
jakepetroules, line 1457: ✘ Test stressTestWithLittleOutput() recorded an issue at IntegrationTests.swift:1460:6: Caught error: An error occurred within the AsyncIO subsystem: failed to add 27 to epoll list. Underlying error: UnderlyingError(rawValue: 17)

Could this be because `_safelyClose` does not call AsyncIO.shared.removeRegistration for the fd before closing it? Reading I/O fails because of low limit, fails, calls _safelyClose, fd is closed, fd # gets reused, tries to re-add, already in the epoll list, boom.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: No This should be 1. This value starts with 0 and every time we run `shutdown` we increment by one. I only want to run shutdown exactly 1 times. If it's `>=1` then it will run it `n - 1` times instead of 1 time.

Sources/Subprocess/Teardown.swift
iCharlesHu: No particular reason besides this seems simpler and more concise and guarantees that we don't accidentally reap the child.

Tests/SubprocessTests/AsyncIOTests.swift
iCharlesHu: Unfortunately this one doesn't work because `#require` requires the type to be `Copyable`.

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: IMO since this test is specifically testing the literal path, I'd rather just wait until we hit an edge case and then fix the test. I think it's defeats the point of test if we ended up replicating the logic for `.name()` so we can find the right path.

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: I think initially I misunderstood the `CreateProcessW` API thought Windows does not support this feature. Now I'm going back and read it again it does seem like it would work. I'll give it a try.

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: This is intentional. We want to wait for both tasks to finish and `let result = result` only check for the one that actually returns result. We also need to wait for the writer task which doesn't return anything.

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: Ditto. This is intentionally to wait for both tasks.

Sources/Subprocess/IO/AsyncIO+Linux.swift
jakepetroules: Ah, of course. Sorry, I misread this before.

Tests/SubprocessTests/IntegrationTests.swift
jakepetroules: Oh, right. In that case, how about using async tasks instead of a task group? That would give you a cleaner way to wait for both with no force unwraps. Both things continue to run in parallel and guaranteed to wait for everything to complete.

```swift
async let buffer = {
    var buffer = Data()
    for try await chunk in standardOutput {
    let currentChunk = chunk.withUnsafeBytes { Data($0) }
        buffer += currentChunk
    }
    return buffer
}()

_ = try await standardInputWriter.write(Array(expected))
try await standardInputWriter.finish()

return try await buffer
```

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: We can, but I don't think it's necessary because `epoll_ctl` will fail if it's already registered so you get the same error.

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: Ahh good call.

iCharlesHu:
Introduced `ProcessMonitoringTests` to unit test `monitorProcessTermination()`.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+                    currentPath: systemDirectory,
+                    pathExtensions: pathExtensions,
+                    storage: &possiblePaths
+                )
+            }
+            // 4. 16 bit Systen Directory
+            // Windows documentation stats that
+            // "No such standard function (similar to GetSystemDirectory)
+            // exists for the 16-bit system folder". Use C:\Windows\System instead
+            let systemDirectory16 = FilePath(#"C:\Windows\System"#)
```
jakepetroules: There's no guarantee Windows is installed on the C:\ drive.

I think to get the 16-bit Windows directory you need to call GetWindowsDirectoryW (which will give you C:\Windows in most cases), then append a "System" component.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+            insertExecutableAddingExtension(
+                name,
+                currentPath: systemDirectory16.string,
+                pathExtensions: pathExtensions,
+                storage: &possiblePaths
+            )
+            // 5. The Windows directory
+            let windowsDirectorySize = GetSystemWindowsDirectoryW(nil, 0)
```
jakepetroules: Based on the documentation I think this should be GetWindowsDirectoryW instead of GetSystemWindowsDirectoryW -- the former will give you the user-specific Windows directory for terminal services aware apps.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+                    currentPath: applicationDirectory,
+                    pathExtensions: pathExtensions,
+                    storage: &possiblePaths
+                )
+            }
+            // 2. Current directory
+            let directorySize = GetCurrentDirectoryW(0, nil)
+            let currentDirectory = try? fillNullTerminatedWideStringBuffer(
+                initialSize: directorySize >= 0 ? directorySize : DWORD(MAX_PATH),
+                maxSize: DWORD(MAX_PATH)
```
jakepetroules: maxSize needs to be `DWORD(Int16.max)` to correctly handle extended length paths. `MAX_PATH` is only 260.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+            var pathExtensions: Set<String> = Set(["com", "exe", "cmd", "bat"])
+            if let extensionList = currentEnvironmentValues["PATHEXT"] {
+                for var ext in extensionList.split(separator: ";") {
+                    ext.removeFirst(1)
+                    pathExtensions.insert(String(ext).lowercased())
+                }
+            }
+            // 1. The directory from which the application loaded.
+            let applicationDirectory = try? fillNullTerminatedWideStringBuffer(
+                initialSize: DWORD(MAX_PATH), maxSize: DWORD(MAX_PATH)
```
jakepetroules: maxSize needs to be `DWORD(Int16.max)` to correctly handle extended length paths. `MAX_PATH` is only 260.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+                    currentPath: currentDirectory,
+                    pathExtensions: pathExtensions,
+                    storage: &possiblePaths
+                )
+            }
+            // 3. System directory (System32)
+            let systemDirectorySize = GetSystemDirectoryW(nil, 0)
+            let systemDirectory = try? fillNullTerminatedWideStringBuffer(
+                initialSize: systemDirectorySize >= 0 ? systemDirectorySize : DWORD(MAX_PATH),
+                maxSize: DWORD(MAX_PATH)
```
jakepetroules: maxSize needs to be `DWORD(Int16.max)` to correctly handle extended length paths. `MAX_PATH` is only 260.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+                name,
+                currentPath: systemDirectory16.string,
+                pathExtensions: pathExtensions,
+                storage: &possiblePaths
+            )
+            // 5. The Windows directory
+            let windowsDirectorySize = GetSystemWindowsDirectoryW(nil, 0)
+            let windowsDirectory = try? fillNullTerminatedWideStringBuffer(
+                initialSize: windowsDirectorySize >= 0 ? windowsDirectorySize : DWORD(MAX_PATH),
+                maxSize: DWORD(MAX_PATH)
```
jakepetroules: maxSize needs to be `DWORD(Int16.max)` to correctly handle extended length paths. `MAX_PATH` is only 260.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+                storage.insert(fullPath.string)
+            }
+        }
+
+        switch self.storage {
+        case .executable(let name):
+            var possiblePaths: Set<String> = []
+            let currentEnvironmentValues = Environment.currentEnvironmentValues()
+            // If `name` does not include extensions, we need to try these extensions
+            var pathExtensions: Set<String> = Set(["com", "exe", "cmd", "bat"])
```
jakepetroules: why hardcode these extensions at all? why not rely solely on PATHEXT? also, the default order in PATHEXT is com,exe,bat,cmd, rather than cmd appearing before bat

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2296 @@
...
+        let result = try await _run(
+            setup,
+            input: .none,
+            output: .string(limit: .max),
+            error: .discarded
+        )
+        #expect(result.terminationStatus.isSuccess)
+        let pathValue = try #require(result.standardOutput)
+        #if os(Windows)
+        let expected = try #require(ProcessInfo.processInfo.environment["Path"])
```
jakepetroules: Path/PATH must be treated case-insensitively as there's no guarantee of its capitalization form.

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -0,0 +1,2296 @@
...
+        #endif
+
+        await #expect(throws: expectedError) {
+            _ = try await Subprocess.run(.name("do-not-exist"), output: .discarded)
+        }
+    }
+
+    @Test func testExecutableAtPath() async throws {
+        #if os(Windows)
+        let cmdExe = ProcessInfo.processInfo.environment["COMSPEC"] ??
```
jakepetroules, line 92: needs to be treated case insensitively (it's ComSpec by default), and you might as well make this #require without falling back to a hardcoded path, or just call GetSystemDirectoryW and join cmd.exe to that.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -629,6 +750,148 @@ extension Executable {
...
+            // 1. The directory from which the application loaded.
+            let applicationDirectory = try? fillNullTerminatedWideStringBuffer(
+                initialSize: DWORD(MAX_PATH), maxSize: DWORD(MAX_PATH)
+            ) {
+                return GetModuleFileNameW(nil, $0.baseAddress, DWORD($0.count))
+            }
+            if let applicationDirectory {
+                insertExecutableAddingExtension(
+                    name,
+                    currentPath: applicationDirectory,
```
jakepetroules: Don't you need to remove the last path component first?

Sources/Subprocess/Error.swift
```diff
@@ -111,7 +111,7 @@ extension SubprocessError: CustomStringConvertible, CustomDebugStringConvertible
         case .failedToWriteToSubprocess:
             return "Failed to write bytes to the child process."
         case .failedToMonitorProcess:
-            return "Failed to monitor the state of child process with underlying error: \(self.underlyingError!)"
+            return "Failed to monitor the state of child process with underlying error: \(self.underlyingError == nil ? "nil" : "\(self.underlyingError!)")"
```
jakepetroules: to avoid the force unwrap:

```suggestion
            return "Failed to monitor the state of child process with underlying error: \(self.underlyingError.map { "\($0)" } ?? "nil")"
```

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+import Foundation
+import TestResources
+import _SubprocessCShims
+@testable import Subprocess
+
+@Suite("Subprocess Process Monitoring Unit Tests", .serialized)
+struct SubprocessProcessMonitoringTests {
+
+    init() {
+        #if os(Linux)
```
jakepetroules: Also needs to check Android:

```suggestion
        #if os(Linux) || os(Android)
```

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+
+    private func longRunningProcess(withTimeOutSeconds timeout: Double? = nil) -> Configuration {
+        #if os(Windows)
+        let waitTime = timeout ?? 99999
+        return Configuration(
+            executable: .name("powershell.exe"),
+            arguments: ["-Command", "Start-Sleep -Seconds \(waitTime)"]
+        )
+        #else
+        let waitTime = timeout != nil ? "\(timeout!)" : "infinite"
```
jakepetroules: to avoid the force unwrap:

```suggestion
        let waitTime = timeout.map { "\($0)" } ?? "infinite"
```

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+
+    @Test func testExitCode() async throws {
+        let config = self.immediateExitProcess(withExitCode: 42)
+        let spawnResult = try config.spawn(
+            withInput: self.devNullInputPipe(),
+            outputPipe: self.devNullOutputPipe(),
+            errorPipe: self.devNullOutputPipe()
+        )
+        defer {
+            spawnResult.execution.processIdentifier.close()
```
jakepetroules: Consider a withSpawnResult helper that guarantees closing the process descriptor? The pattern is repeated many times throughout the file and a future test author could forget to call close.

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        async let monitorResult = try monitorProcessTermination(
+            for: spawnResult.execution.processIdentifier
+        )
+        // Send signal to process
+        try spawnResult.execution.send(signal: .terminate)
```
jakepetroules: Theoretically racy on Darwin and OpenBSD (or any other unix without process descriptors).

By the time send() is called, monitorProcessTermination might've reaped the PID. It looks like there's no reason you can't just call send before monitorProcessTermination instead.

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+            withInput: self.devNullInputPipe(),
+            outputPipe: self.devNullOutputPipe(),
+            errorPipe: self.devNullOutputPipe()
+        )
+        defer {
+            spawnResult.execution.processIdentifier.close()
+        }
+        // Manually wait for the process to make sure it exits
+        #if os(Windows)
+        WaitForSingleObject(
```
jakepetroules: We still shouldn't block the concurrency pool in tests. This should use WaitForSingleObjectEx with a continuation, and the waitid below should spawn a pthread for the blocking work e.g. with something like this (the other pthread_create overload is already in Subprocess+Linux.swift)

```swift
internal func withPlatformThread<T: Sendable>(_ body: @Sendable @escaping () throws -> T) async throws -> T {
    return try await withCheckedThrowingContinuation { continuation in
        do {
            _ = try pthread_create {
                continuation.resume(with: Result { try body() })
            }
        } catch {
            continuation.resume(throwing: error)
        }
    }
}
```

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+    @Test func testInvalidProcessIdentifier() async throws {
+        #if os(Windows)
+        let expectedError = SubprocessError(
+            code: .init(.failedToMonitorProcess),
+            underlyingError: .init(rawValue: DWORD(ERROR_INVALID_PARAMETER))
+        )
+        let processIdentifier = ProcessIdentifier(
+            value: .max, processDescriptor: INVALID_HANDLE_VALUE, threadHandle: INVALID_HANDLE_VALUE
+        )
+        #elseif os(Linux)
```
jakepetroules:
```suggestion
        #elseif os(Linux) || os (FreeBSD)
```

I have the code to restore FreeBSD support on a branch; one less thing to update.

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+        defer {
+            child2Result.execution.processIdentifier.close()
+        }
+        // Monitor child2, but make sure we don't reap child1's status
+        let status = try await monitorProcessTermination(for: child2Result.execution.processIdentifier
+        )
+        #expect(status.isSuccess)
+        // Make sure we can still fetch child 1
+        #if os(Windows)
+        let rc = WaitForSingleObject(
```
jakepetroules: Same comments about blocking functions

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -0,0 +1,361 @@
...
+                    let status = try await monitorProcessTermination(for: pid)
+                    #expect(status.isSuccess)
+                }
+            }
+
+            try await group.waitForAll()
+        }
+
+        for pid in spawnedProcesses {
+            pid.close()
```
jakepetroules: These leak if any errors are thrown; should go in a defer block at the top

Sources/Subprocess/Platforms/Subprocess+Windows.swift
iCharlesHu: We can't solo rely on `PATHEXT` because it could not exit. I don't think this is the sort of value that warrants throwing an error, so let's just use some fallback value to try our best. Worst case (if the default value is "wrong“), we will get the same error here, which is "executable not found".

iCharlesHu: > he default order in PATHEXT is com,exe,bat,cmd, rather than cmd appearing before bat

Does the order matter here?  I'm currently using a `Set` so it's orderless anyway.

jakepetroules: Absolutely -- if a directory contains both foo.bat and foo.exe you'll want to search in deterministic order.

Tests/SubprocessTests/ProcessMonitoringTests.swift
iCharlesHu: I'm only calling `WaitForSingleObject` directly here because I control the process being spawned, and in this case, we are deliberately spawning processes that exit immediately. I don't think `WaitForSingleObject` and `waitid` are technically blocking here, and it seems more wasteful to spawn a new thread.

jakepetroules: To be fair, there is a very small race condition where either of these could be blocking. By the time you call this, the process _might_ not have technically exited yet.

iCharlesHu: > By the time you call this, the process might not have technically exited yet.

I agree. I just think that that's no more "blocking" than spawning another thread since the purpose is to test something that exit immediately.

jakepetroules: I suppose that's a fair point, it probably can't block the concurrency pool in a way that can deadlock.

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -349,60 +357,79 @@ extension AsyncIO {
...
-                        // Throw all other errors
-                        try self.removeRegistration(for: fileDescriptor)
-                        throw SubprocessError.UnderlyingError(rawValue: errno)
+                        if self.shouldWaitForNextSignal(with: capturedErrno) {
+                            // No more data for now wait for the next signal
+                            break
+                        } else {
+                            // Throw all other errors
+                            try self.removeRegistration(for: fileDescriptor)
+                            perror("Throw error directly")
```
jakepetroules: Left over debugging code - libraries should not write to standard streams

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -349,60 +357,79 @@ extension AsyncIO {
...
+                                code: .init(.failedToReadFromSubprocess),
+                                underlyingError: .init(rawValue: capturedErrno)
+                            )
+                        }
                     }
                 }
             }
+        } catch {
+            // Reset error code to .failedToRead to match other platforms
+            perror("Re threw the signal error: \(error)")
```
jakepetroules: Left over debugging code - libraries should not write to standard streams

jakepetroules (approved):
Looks good to me!

PR #136, opened 2025-08-11
Fix the build with Musl libc

### LINKED ISSUE #118: Subprocess unable to build using musl on linux host
**Describe the bug**
Hello everyone 👋 

I found this curious compilation error when building an SPM package that has a dependency on subprocess when building using the musl sdk on a linux host:

```swift
Swift.OpaquePointer.init:2:8: note: 'init(bitPattern:)' declared here
1 | struct OpaquePointer {
2 | public init?(bitPattern: Int)}
  |        `- note: 'init(bitPattern:)' declared here
3 | 
/app_src/Backend/.build/checkouts/swift-subprocess/Sources/Subprocess/Platforms/Subprocess+Linux.swift:358:28: error: missing argument for parameter 'b
itPattern' in call
356 |     var thread: pthread_t?
357 |     #else
358 |     var thread = pthread_t()
    |                            `- error: missing argument for parameter 'bitPattern' in call
359 |     #endif
360 |     _ = pthread_create(
```

**To Reproduce**
1. Get a linux host (preferably RHEL)
2. Install swift
3. Install a matching static sdk (MUSL)
4. Compile a package that depends on subprocess (preferably in a different processor architecture than the host)

**Expected behavior**
It should compile without errors

 - OS RHEL UBI 9
 - Swift version: Static SDK (MUSL) (6.1.2, 6.2 and main)

**Additional context**
I tracked the issue and it seems that it was introduced by this commit: https://github.com/swiftlang/swift-subprocess/commit/7faeeb898fd823a7b945e20fec9d9e59f980bcca

jakepetroules:
This also adds CI to prevent it from regressing. Also fixed the Android build.

Closes #118


Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -40,7 +41,7 @@ final class AsyncIO: Sendable {
 
     typealias OutputStream = AsyncThrowingStream<AsyncBufferSequence.Buffer, any Error>
 
-    private final class MonitorThreadContext {
+    private final class MonitorThreadContext: Sendable {
```
iCharlesHu: Can this be `struct` now since we are no longer passing it as a pointer?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -31,6 +40,42 @@ import Synchronization
...
+extension UInt32 {
+    // for EPOLLIN/EPOLLOUT
+    var rawValue: UInt32 {
+        self
+    }
+}
+
+extension Int32 {
+    var rawValue: UInt32 {
+        UInt32(self)
```
iCharlesHu: This will `fatalError` if `self` is negative right? Did you mean to use `UInt32(bitPattern:)`?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -31,6 +40,42 @@ import Synchronization
...
+    }
+}
+#elseif canImport(Musl)
+extension EPOLL_EVENTS {
+    init(_ rawValue: Int32) {
+        self.init(UInt32(rawValue))
```
iCharlesHu: Same here: UInt32() will fatalError if rawValue is actually negative.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -31,6 +40,42 @@ import Synchronization
...
+    }
+}
+
+extension Int32 {
+    // for EPOLLIN/EPOLLOUT
+    var rawValue: UInt32 {
+        UInt32(self)
```
iCharlesHu: ditto

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -386,7 +435,7 @@ private enum ProcessMonitorState {
     case failed(SubprocessError)
 }
 
-private final class MonitorThreadContext {
+private final class MonitorThreadContext: Sendable {
```
iCharlesHu: Ditto: can this be a `struct` now?

Sources/Subprocess/IO/AsyncIO+Linux.swift
jakepetroules: Done

Sources/Subprocess/Platforms/Subprocess+Linux.swift
jakepetroules: Done

Sources/Subprocess/Platforms/Subprocess+Linux.swift
jakepetroules: Done

Sources/Subprocess/Platforms/Subprocess+Linux.swift
jakepetroules: Done

Sources/Subprocess/Platforms/Subprocess+Linux.swift
jakepetroules: Done

PR #139, opened 2025-08-12
Add Debian 12 (Bookworm) support

cmcgee1024:


iCharlesHu:
The test failure looks relevant but infrastructure?

README.md
```diff
@@ -232,6 +232,7 @@ The table below describes the current level of support that Subprocess has for v
 | **Ubuntu 22.04** | Supported |
 | **Ubuntu 24.04** | Supported |
 | **Red Hat Universal Base Image 9** | Supported |
+| **Debian 12** | Supported |
```
jakepetroules, line 235: Is there much point listing individual Linux distributions versus just saying "Linux"? It's implied to be "whatever Swift itself supports".

Besides, it's really "Glibc" and "Musl" that we support; there is nothing at the distribution level that swift-subprocess is aware of. Just the kernel and the libc.

cmcgee1024: I wondered about this myself. Because subprocess is a package, I suppose that it must be built on host where a supported toolchain can exist, and therefore can probably claim that it supports every Linux that has a supported toolchain. This can be different than the supported deployment platforms, which might be based on Glibc/Musl versions?

cmcgee1024:
> because we'd have to replicate a bunch of logic in the GitHub workflow to set up and install the toolchain versus letting the shared one do it for us

Swiftly can be used for most of this part so that it doesn't have to be duplicated, and can be shared between the different distributions. It works quite well in practice.

PR #140, opened 2025-08-14
Introduce preExecProcessAction on Darwin, rename preSpawnProcessConfigurator to preExecProcessAction on Linux

### LINKED ISSUE #25: Could preSpawnProcessConfigurator be made available for posix_spawn OR fork/exec?
Currently, setting `preSpawnProcessConfigurator` requires pre-fork on Linux, meaning `preSpawnProcessConfigurator` can't be used to control the posix_spawn path.

On Linux, the signature is:

```swift
public var preSpawnProcessConfigurator: (@convention(c) @Sendable () -> Void)? = nil
```

Could it instead become:

```swift
public enum SpawnStrategy {
    case posixSpawn(_ attr: posix_spawnattr_t?, _ fileactions: posix_spawn_file_actions_t?)
    case forkExec
}

public var preSpawnProcessConfigurator: (@convention(c) @Sendable (inout SpawnStrategy) -> Void)? = nil
```

...and then the configurator could be called for both the posix_spawn and fork/exec paths, without influencing which one is taken.

Bonus points if this could be applied to Darwin as well (removing the forkExec option), to make the PlatformOptions API interface a little more similar.

jakepetroules:
This patch introduces a preExecProcessAction on Darwin, which runs immediately before posix_spawn-as-exec, just as preSpawnProcessConfigurator (now called preExecProcessAction) runs immediately before exe on Linux.

The reason to introduce this API is because there is presently no way to run code _after_ fork on Darwin in the case where pre-forking occurs (which is true when setting a uid, gid, sgroups, or creating a session). This can be useful when performing operations that modify process global state and wanting to do so without affecting the parent.

The reason to rename the Linux API is that these are fundamentally two different logical operations: one is modifying flags to the process spawning mechanism (CreateProcessW, posix_spawn), while the other is performing an action at a specific point in the process lifecycle (before exec). Only Darwin uses posix_spawn, so there is no opportunity to modify any flags, and preSpawnProcessConfigurator is therefore not available on Linux/Android/FreeBSD/etc. On Windows, there is no fork/exec concept, so preExecProcessAction is not available on Windows.

This brings the set of APIs to:

- preSpawnProcessConfigurator (Windows, Darwin) - modifies flags bitset / STARTUPINFOW, or posix_spawnattr_t / posix_spawn_file_actions_t
- preExecProcessAction (Darwin, Linux/Android/FreeBSD/etc.) - runs before exec or posix_spawn-as-exec

or

- Windows: preSpawnProcessConfigurator
- Darwin: preSpawnProcessConfigurator + preExecProcessAction
- Linux/Android/FreeBSD/etc.: preExecProcessAction

This means that Darwin has both preSpawnProcessConfigurator and preExecProcessAction, because the pre-forking case uses posix_spawn _and_ fork. It also give us the opportunity to introduce preSpawnProcessConfigurator on Linux and other platforms in the future if they introduce a similar POSIX_SPAWN_CLOEXEC_DEFAULT flag or POSIX otherwise standardizes a mechanism to disable file descriptor inheritance by default, since that would allow us to use posix_spawn on those platforms.

Note that merely setting preExecProcessAction will also cause pre-forking to occur.

Closes #25

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -81,6 +81,29 @@ public struct PlatformOptions: Sendable {
                 inout posix_spawn_file_actions_t?
             ) throws -> Void
         )? = nil
+    /// A closure to configure platform-specific
+    /// spawning constructs. This closure enables direct
+    /// configuration or override of underlying platform-specific
+    /// spawn settings that `Subprocess` utilizes internally,
+    /// in cases where Subprocess does not provide higher-level
+    /// APIs for such modifications.
+    ///
+    /// On Darwin, Subprocess uses `posix_spawn()` as the
+    /// underlying spawning mechanism, but may require an initial `fork()`
+    /// depending on the configured `PlatformOptions`.
+    /// This closure is called after `fork()` but before `posix_spawn()`
+    /// (with the `POSIX_SPAWN_SETEXEC` flag set).
+    /// You may use it to call any necessary process setup functions.
+    ///
+    /// - note: You can set both `preExecProcessAction` and
+    /// `preSpawnProcessConfigurator` and both will be called.
+    /// Setting `preExecProcessAction` will always cause Subprocess
+    /// to pre-`fork()` before calling `posix_spawn()` (with the
+    /// `POSIX_SPAWN_SETEXEC` flag set) even if it would not have otherwise
+    /// done so based on the configured `PlatformOptions`.
+    ///
+    /// - warning: You may ONLY call [async-signal-safe functions](https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html) within this closure (note _"The following table defines a set of functions and function-like macros that shall be async-signal-safe."_).
+    public var preExecProcessAction: (@convention(c) @Sendable () -> Void)? = nil
 
     public init() {}
 }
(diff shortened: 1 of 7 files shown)


Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -81,6 +81,22 @@ public struct PlatformOptions: Sendable {
...
         )? = nil
+    /// A closure to configure platform-specific
+    /// spawning constructs. This closure enables direct
+    /// configuration or override of underlying platform-specific
+    /// spawn settings that `Subprocess` utilizes internally,
+    /// in cases where Subprocess does not provide higher-level
+    /// APIs for such modifications.
+    ///
+    /// On Darwin, Subprocess uses `posix_spawn()` as the
+    /// underlying spawning mechanism, but may require an initial `fork()`
```
iCharlesHu, line 92: Could you explicitly mention that

1) On Darwin you can set both `preExecProcessAction` and `preSpawnProcessConfigurator`, and
2) Setting `preExecProcessAction` will cause Subprocess to use the "pre-fork" and use posix_spawn solely as `exec` replacement with `POSIX_SPAWN_SETEXEC` flag.

Tests/SubprocessTests/SubprocessTests+Darwin.swift
```diff
@@ -26,6 +26,19 @@ import Testing
 // MARK: PlatformOptions Tests
 @Suite(.serialized)
 struct SubprocessDarwinTests {
+    @Test func testSubprocessPlatformOptionsPreExecProcessAction() async throws {
```
iCharlesHu, line 29: Could you also add a test to test setting both to make sure both works at the same time?

iCharlesHu (approved):

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
jakepetroules: Done

Tests/SubprocessTests/SubprocessTests+Darwin.swift
jakepetroules: Done

PR #147, opened 2025-08-15
Create a runPipe that sets up a process pipeline from multiple Configurations

cmcgee1024:
The runPipe() function helps to create easy multi-process pipelines in a
similar method as the run() function that accepts a Configuration and
returns a CollectedResult.

```
let finalJson = try await run(
    Configuration(executable: .path("/bin/echo"), arguments: [#"{"abc":"def"}"#]),
    Configuration(executable: .name("jq")),
    output: .string(limit: 1024*1024)
)
print("\(finalJson.standardOutput)")
```

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -557,6 +557,100 @@ public func run<
     )
 }
 
+public enum PipeConfigurationError: Error {
+    case emptyConfiguration
+}
+
+/// Run a series of `Configuration` asynchronously and returns
+/// a `CollectedResult` collecting the result of the final process at the
+/// end of a pipeline of all processes joined together with pipes from standard input
+/// to standard output.
+///
+/// - Parameters:
+///   - configurations: The `Subprocess` configuration to run in a pipeline.
+///   - input: The input to send to the first executable.
+///   - output: The method to use for redirecting the standard output of the last executable.
+///   - error: The method to use for redirecting the standard error of all executables.
+/// - Returns a CollectedResult containing the result of the pipeline and the termination status of the final executable.
+public func runPipe<
+    Input: InputProtocol,
+    Output: OutputProtocol,
+    Error: OutputProtocol
+>(
+    _ pipeline: Configuration...,
+    input: Input = .none,
+    output: Output,
+    error: Error = .discarded
+) async throws -> CollectedResult<Output, Error> {
+    try await runPipe(pipeline, input: input, output: output, error: error)
+}
+
+/// Run a series of `Configuration` asynchronously and returns
+/// a `CollectedResult` collecting the result of the final process at the
+/// end of a pipeline of all processes joined together with pipes from standard input
+/// to standard output.
+///
+/// - Parameters:
+///   - configurations: The `Subprocess` configuration to run in a pipeline.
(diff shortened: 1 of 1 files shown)


Sources/Subprocess/API.swift
```diff
@@ -557,6 +557,100 @@ public func run<
...
+        }
+
+        if idx == pipeline.count - 1 {
+            let result = await Task {
+                try await run(cmd, input: currentInput, output: output, error: error)
+            }.result
+
+            return try result.get()
+        } else {
+            Task {
```
jakepetroules: It would be best to use a task group here, rather than unstructured tasks. That way the subtasks will inherit cancellation & priority properly.

Sources/Subprocess/API.swift
```diff
@@ -557,6 +557,100 @@ public func run<
...
+        } else if idx != pipeline.count - 1 {
+            let (reader, writer) = try FileDescriptor.pipe()
+            prevInput = .fileDescriptor(reader, closeAfterSpawningProcess: true)
+            currentOutput = .fileDescriptor(writer, closeAfterSpawningProcess: true)
+        } else {
+            currentOutput = output
+        }
+
+        if idx == pipeline.count - 1 {
+            let result = await Task {
```
jakepetroules: The task here looks redundant; this:

```swift
let result = await Task { try await (...) }.result
return try result.get()
```

is equivalent to:
```swift
return try await (...)
```

Sources/Subprocess/API.swift
```diff
@@ -557,6 +557,100 @@ public func run<
...
+public func runPipe<
+    Input: InputProtocol,
+    Output: OutputProtocol,
+    Error: OutputProtocol
+>(
+    _ pipeline: [Configuration],
+    input: Input = .none,
+    output: Output,
+    error: Error = .discarded
+) async throws -> CollectedResult<Output, Error> {
```
jakepetroules: Do shells like bash start every process in a pipeline in parallel regardless of how many there are? Should there be a maxConcurrency parameter to limit it? People probably won't set up 1000+-process pipelines in practice, but maybe there are actually some use cases for that?

Another approach could be an optional parameter like an execution pool, defined like:

```swift
protocol ExecutionPool {
    func execute<ReturnValue>(_ work: () async throws -> sending ReturnValue) async throws -> ReturnValue
}
```

cmcgee1024: I've changed this to a variadic function, which should help to reduce the probability of very large process bombs.

Sources/Subprocess/API.swift
cmcgee1024: I've converted this to use a task group, which collects the results, and failures.

cmcgee1024:
Pipefail is a good point, and something that I can add as a parameter.

In terms of individually controlling, or redirecting elements in the pipeline I think that would be difficult to achieve with a single function call. I think that the trouble is that the existing Configuration structure isn't really a configuration in the sense that we're talking about here because it doesn't do anything to specify input/output/error. If it did, then this could become multiple run function calls, and the type checker could verify that the next in the pipe has a standard out fd (redirected, or not) that can be the next function's input.

I'll see if I can get an experiment going that yields some ergonomic results.

Issue #129, opened 2025-07-29
Assertion failure in TrackedDispatchIO


davedelong:
**Describe the bug**
Running a process that attempts to operate on the spawned process's output or error streams will crash in a `fatalError()` inside `TrackedDispatchIO`

**To Reproduce**
```swift
let _ = try await Subprocess.run(.name("ls"), body: { execution, stdin, stdout, stderr in
    for try await line in stdout.lines(encoding: UTF8.self) {
        print(line)
    }
})
```

**Expected behavior**
This should print the output of `ls` line by line.

**Environment (please complete the following information):**
- swift-subprocess @ 5715ed4
- macOS 15.5 (24F74)
- Apple Swift version 6.1.2 (swift-6.1.2-RELEASE)

**Additional context**
A glance at the code suggests that `TrackedDispatchIO` (and `TrackedFileDescriptor`) might be missing their attempts to close their respective handles inside their `deinit` methods.

davedelong:
@iCharlesHu I checked out that commit and the branch does not compile:

```'Region' is not a member type of struct 'Dispatch.DispatchData'```

omrd:
I am having the same DispatchData.Region error.

davedelong:
I forked this and added `import Foundation` to `Buffer.swift`, and things compile just fine now. However, I don't know if that's the "right" fix.

PR #152, opened 2025-08-19
Fix the build on FreeBSD (and OpenBSD?)

### LINKED ISSUE #115: OpenBSD / FreeBSD Support for Subprocess
**Describe the bug**
Building swift-subprocess on OpenBSD fails in `Subprocess+Linux.swift` (and there is no other `spawn` implementation in `Subprocess+Unix.swift`).

**To Reproduce**
1. `swift build` on OpenBSD.

**Expected behavior**
Package builds.

 - OpenBSD
 - Swift 6.2/HEAD-ish.

**Additional context**

```
/tmp/swift-subprocess/Sources/Subprocess/Platforms/Subprocess+Linux.swift:373:39: error: value of type 'siginfo_t' has no member 'si_pid'
371 |                 errno = 0
372 |                 if waitid(P_ALL, id_t(0), &siginfo, WEXITED | WNOWAIT) == 0 {
373 |                     let pid = siginfo.si_pid
    |                                       `- error: value of type 'siginfo_t' has no member 'si_pid'
```

This is because for whatever reason, `siginfo_t` on this platform has some internal fields that may be undergoing change and reference to `si_pid` is `#defined` as the internal field path, i.e., `_data._proc._pid`. Of course, Swift doesn't know how to paste those tokens together when importing macros from C.

We probably want some C shims for the platform or something.

As to not lose context, I'm just making note of this right now since it's not in the critical path, but might be in the future, otherwise I'd just dabble with this and fix it myself. (Feel free to assign it to me if you'd like.)

jakepetroules:
For now, we use Dispatch for AsyncIO and process termination monitoring, since BSDs (including macOS) all use kqueue as the Dispatch backend anyways.

This likely also fixes the build on OpenBSD, but I haven't been able to verify that.

Closes #115

### DIFF
--- a/Sources/Subprocess/Buffer.swift
+++ b/Sources/Subprocess/Buffer.swift
@@ -17,7 +17,7 @@
 extension AsyncBufferSequence {
     /// A immutable collection of bytes
     public struct Buffer: Sendable {
-        #if canImport(Darwin)
+        #if SUBPROCESS_ASYNCIO_DISPATCH
         // We need to keep the backingData alive while Slice is alive
         internal let backingData: DispatchData
         internal let data: DispatchData.Region
@@ -45,7 +45,7 @@ extension AsyncBufferSequence {
         internal static func createFrom(_ data: [UInt8]) -> [Buffer] {
             return [.init(data: data)]
         }
-        #endif // canImport(Darwin)
+        #endif // SUBPROCESS_ASYNCIO_DISPATCH
     }
 }
 
@@ -92,7 +92,7 @@ extension AsyncBufferSequence.Buffer {
 
 // MARK: - Hashable, Equatable
 extension AsyncBufferSequence.Buffer: Equatable, Hashable {
-    #if canImport(Darwin)
+    #if SUBPROCESS_ASYNCIO_DISPATCH
     public static func == (lhs: AsyncBufferSequence.Buffer, rhs: AsyncBufferSequence.Buffer) -> Bool {
         return lhs.data == rhs.data
     }
@@ -104,7 +104,7 @@ extension AsyncBufferSequence.Buffer: Equatable, Hashable {
     // else Compiler generated conformances
 }
 
-#if canImport(Darwin)
+#if SUBPROCESS_ASYNCIO_DISPATCH
 extension DispatchData.Region {
     static func == (lhs: DispatchData.Region, rhs: DispatchData.Region) -> Bool {
         return lhs.withUnsafeBytes { lhsBytes in
@@ -120,7 +120,7 @@ extension DispatchData.Region {
         }
     }
 }
-#if !SubprocessFoundation
+#if !canImport(Darwin) || !SubprocessFoundation
 /// `DispatchData.Region` is defined in Foundation, but we can't depend on Foundation when the SubprocessFoundation trait is disabled.
 extension DispatchData {
     typealias Region = _ContiguousBufferView
--- a/Sources/Subprocess/IO/AsyncIO+Linux.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Linux.swift
@@ -11,7 +11,7 @@
 
 /// Linux AsyncIO implementation based on epoll
 
-#if canImport(Glibc) || canImport(Android) || canImport(Musl)
+#if os(Linux) || os(Android)
 
 #if canImport(System)
 @preconcurrency import System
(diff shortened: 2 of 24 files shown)


jakepetroules:
@3405691582 does this build on OpenBSD?

Sources/Subprocess/Buffer.swift
```diff
@@ -120,5 +120,45 @@ extension DispatchData.Region {
         }
     }
 }
-#endif
+#if !canImport(Darwin)
```
iCharlesHu: Actually, can you make this available on Darwin and BSD? As it turns out, `DispatchData.Region` is defined by `Foundation`, not `Dispatch`, which means we can't use it directly because this package sits below `Foundation`.

This would also close https://github.com/swiftlang/swift-subprocess/issues/129

3405691582:
Some of the changes discussed on #115 are still missing, so with just the patch atop HEAD it doesn't build. If I apply those changes, namely those below, it seems to build:
* Sources/Subprocess/Platforms/Subprocess+Linux.swift needs to expose the extension and [these changes](https://github.com/swiftlang/swift-subprocess/issues/115#issuecomment-3036878054)
* Sources/Subprocess/Platforms/Subprocess+Unix.swift: the snprintf shim change
* Sources/_SubprocessCShims/include/process_shims.h: snprintf shim needs TARGET_OS_UNIX
* Sources/_SubprocessCShims/process_shims.c: snprintf shim needs TARGET_OS_UNIX
* Tests/SubprocessTests/ProcessMonitoringTests.swift: `testInvalidProcessIdentifier` needs an `os(OpenBSD)`; in `testDoesNotReapUnrelatedChildProcess` the `si_status` extension doesn't seem to be active and I don't know why at the moment.

I tried to run the tests but ran into some other unrelated issues.

Sources/Subprocess/Buffer.swift
jakepetroules: Sure, although I'm a little confused why the CI is not catching that.

iCharlesHu: We have `SubprocessFoundation` trait enabled by default which will get us `DispatchData.Region`. This is only an issue if you specifically turn off `SubprocessFoundation`.

jakepetroules: I split that out into #154

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -266,6 +266,11 @@ final class AsyncIO: Sendable {
                     targetEvent = EPOLL_EVENTS(EPOLLOUT)
                 }
 
+                // Save the continuation (before calling epoll_ctl, so we don't miss any data)
+                _registration.withLock { storage in
```
iCharlesHu, line 270: Can you elaborate a bit more on why this change is needed?

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -411,4 +411,373 @@ extension FileDescriptor {
...
+        try _blockingReap(pid: value)
+    }
+
+    /// Reaps the zombie for the exited process, or returns `nil` if the process is still running. This function will not block.
+    internal func reap() throws(SubprocessError.UnderlyingError) -> TerminationStatus? {
+        try _reap(pid: value)
+    }
+}
+
+@available(*, noasync)
```
iCharlesHu, line 717: Just curious, what's this for?

iCharlesHu (approved):

Sources/Subprocess/Platforms/Subprocess+Unix.swift
jakepetroules: Semantically labels it as blocking -- you'll get an error if you accidentally try to call it in an `async` context.

Sources/Subprocess/IO/AsyncIO+Linux.swift
jakepetroules: There's two threads:
- thread A: thread that calls `registerFileDescriptor`
- thread B: monitor loop

With the original code, the following ordering could occur:
- (thread A) add file descriptor N to epoll
- (thread B) process event for N
- (thread B) look up continuation <-- continuation is nil!
- (thread A) save continuation <-- too late

This change forces the ordering to become:
- (thread A) save continuation
- (thread A) add file descriptor N to epoll
- (thread B) process event for N
- (thread B) look up continuation <-- continuation cannot be nil

PR #154, opened 2025-08-21
Fix the build when the SubprocessFoundation trait is disabled

jakepetroules:


Sources/Subprocess/Buffer.swift
```diff
@@ -120,5 +120,45 @@ extension DispatchData.Region {
...
-#endif
+#if !SubprocessFoundation
+extension DispatchData {
+    typealias Region = _ContiguousBufferView
+
+    var regions: [Region] {
+        contiguousBufferViews
+    }
+
+    /// Unfortunately `DispatchData.Region` is not available on Linux, hence our own wrapper
```
iCharlesHu: Stale comment. It's more like we need `Foundation` in order to use it but we can't depend on Foundation unconditionally.

jakepetroules: Updated

PR #158, opened 2025-08-23
Call spawn() on a background thread because it might block

iCharlesHu:
Introduce `runOnBackgroundThread()` to run closures on a background thread without blocking the Swift Concurrency thread pool.

On Darwin, `runOnBackgroundThread()` uses `DispatchQueue`; on Linux and Windows, it uses a single thread in the work loop.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/85

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Windows.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Windows.swift
@@ -60,9 +60,11 @@ final class AsyncIO: @unchecked Sendable {
     internal init() {
         var maybeSetupError: SubprocessError? = nil
         // Create the the completion port
-        guard let ioCompletionPort = CreateIoCompletionPort(
-            INVALID_HANDLE_VALUE, nil, 0, 0
-        ), ioCompletionPort != INVALID_HANDLE_VALUE else {
+        guard
+            let ioCompletionPort = CreateIoCompletionPort(
+                INVALID_HANDLE_VALUE, nil, 0, 0
+            ), ioCompletionPort != INVALID_HANDLE_VALUE
+        else {
             let error = SubprocessError(
                 code: .init(.asyncIOFailed("CreateIoCompletionPort failed")),
                 underlyingError: .init(rawValue: GetLastError())
@@ -91,45 +93,7 @@ final class AsyncIO: @unchecked Sendable {
                     var bytesTransferred: DWORD = 0
                     var targetFileDescriptor: UInt64 = 0
                     var overlapped: LPOVERLAPPED? = nil
-                // Monitor loop
-                while true {
-                    var bytesTransferred: DWORD = 0
-                    var targetFileDescriptor: UInt64 = 0
-                    var overlapped: LPOVERLAPPED? = nil
 
-                    let monitorResult = GetQueuedCompletionStatus(
-                        context.ioCompletionPort,
-                        &bytesTransferred,
-                        &targetFileDescriptor,
-                        &overlapped,
-                        INFINITE
-                    )
-                    if !monitorResult {
-                        let lastError = GetLastError()
-                        if lastError == ERROR_BROKEN_PIPE {
-                            // We finished reading the handle. Signal EOF by
-                            // finishing the stream.
-                            // NOTE: here we deliberately leave now unused continuation
-                            // in the store. Windows does not offer an API to remove a
-                            // HANDLE from an IOCP port, therefore we leave the registration
-                            // to signify the HANDLE has already been resisted.
-                            let continuation = _registration.withLock { store -> SignalStream.Continuation? in
-                                if let continuation = store[targetFileDescriptor] {
-                                    return continuation
-                                }
-                                return nil
-                            }
-                            continuation?.finish()
-                            continue
-                        } else {
-                            let error = SubprocessError(
-                                code: .init(.asyncIOFailed("GetQueuedCompletionStatus failed")),
-                                underlyingError: .init(rawValue: lastError)
-                            )
-                            reportError(error)
-                            break
-                        }
-                    }
                     let monitorResult = GetQueuedCompletionStatus(
                         context.ioCompletionPort,
                         &bytesTransferred,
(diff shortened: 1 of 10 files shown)


Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,325 @@
...
+#if canImport(Synchronization)
+import Synchronization
+#endif
+
+#if canImport(Darwin)
+internal func runOnBackgroundThread<Result>(
+    _ body: @Sendable @escaping () throws -> Result
+) async throws -> Result {
+    let result = try await withCheckedThrowingContinuation { continuation in
+        // On Darwin, use DispatchQueue directly
```
jakepetroules: Why use a DispatchQueue? I'm not sure if this is better or desirable; would be good to get some input from @weissi or @FranzBusch on this.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -68,89 +68,81 @@ final class AsyncIO: @unchecked Sendable {
...
-            func reportError(_ error: SubprocessError) {
-                let continuations = _registration.withLock { store in
-                    return store.values
-                }
-                for continuation in continuations {
-                    continuation.finish(throwing: error)
+        let context = MonitorThreadContext(ioCompletionPort: ioCompletionPort)
+        let threadHandle: HANDLE
+        do {
+            threadHandle = try begin_thread_x {
```
jakepetroules, line 81: Why not using runOnBackgroundThread here?

Sources/Subprocess/Thread.swift
iCharlesHu: @jakepetroules can you elaborate on what's the concern with using `DispatchQueue`? I know we prefer not to use it on other platforms, but from my understanding on Darwin, it's very much supported/maintained and the de facto threading library. I also chose it because it _should_ have better performance on Darwin than a fixed thread because it's an actual managed thread pool.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Because I think we should now create 3 threads during `Subprocess`’s lifetime:
- `AsyncIO` monitor thread specially for any IO activities
- `monitorProcessTermination` monitor thread specifically for termination monitoring
- `runOnBackgroundThread` for everything else that's blocking

I think this is a nice balance between needing to spawn a new thread on every `run()`.

Sources/Subprocess/Thread.swift
jakepetroules: From what I understand, Dispatch's thread pool is shared with the Swift concurrency pool and therefore you're not actually offloading the work and still subject to the same concerns around deadlocking and thread starvation, especially in low core count environments, like a VM or container that just has a couple cores.

I recall a scenario where this happened with swift-build in a 3-core-count environment; we hit a deadlock even though we offloaded blocking work to a DispatchQueue. Offloading to a dedicated thread resolved it.

I could be wrong about some of the specifics though, so having Johannes or Franz weigh in would be ideal.

Sources/Subprocess/IO/AsyncIO+Windows.swift
jakepetroules: I see, this makes sense now, thanks.

Sources/Subprocess/Thread.swift
FranzBusch: AFAIK this should be fine. Dispatch is automatically growing its thread pool if there is no forward progress. This is also the reason why Dispatch can run into thread explosion issues.

iCharlesHu: Just to be consistent with other platforms I updated it to use `pthread_create` directly.

Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+
+    private func withUnsafeUnderlyingLock<R>(
+        _ body: (UnsafeMutablePointer<MutexType>, inout [BackgroundWorkItem]) throws -> R
+    ) rethrows -> R {
+        #if canImport(WinSDK)
+        EnterCriticalSection(self.mutex)
+        defer {
+            LeaveCriticalSection(self.mutex);
```
jakepetroules: nit: stray semicolon

Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,345 @@
...
+            return queue.removeFirst()
+        }
+    }
+
+    // Only called in parent thread. Signals wait condition to wake up worker thread
+    func enqueue(_ workItem: BackgroundWorkItem) {
+        self.withLock { queue in
+            queue.append(workItem)
+            #if canImport(WinSDK)
+            WakeConditionVariable(self.waitCondition);
```
jakepetroules: nit: stray semicolon

Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,345 @@
...
+            pthread_cond_signal(self.waitCondition)
+            #endif
+        }
+    }
+
+    func shutdown() {
+        self.withLock { queue in
+            queue.removeAll()
+            #if canImport(WinSDK)
+            WakeConditionVariable(self.waitCondition);
```
jakepetroules: nit: stray semicolon

Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,345 @@
...
+    guard case .success(let thread) = _workerThread else {
+        return
+    }
+    guard _workQueueShutdownFlag.addOne() == 1 else {
+        // We already shutdown this thread
+        return
+    }
+    _workQueue.shutdown()
+    #if canImport(WinSDK)
+    WaitForSingleObject(thread, INFINITE);
```
jakepetroules: nit: stray semicolon

Sources/Subprocess/Thread.swift
```diff
@@ -0,0 +1,345 @@
...
+    pthread_cond_destroy(_workQueue.waitCondition)
+    #endif
+    _workQueue.mutex.deallocate()
+    _workQueue.waitCondition.deallocate()
+}
+
+// MARK: - AtomicCounter
+
+#if canImport(Darwin)
+// Unfortunately on Darwin we can unconditionally use Atomic since it requires macOS 15
```
jakepetroules: can => cannot

PR #162, opened 2025-08-28
Introduce .swift-format and enable linting requirement on PR requests

iCharlesHu:
This PR introduces a baseline `.swift-format` and enforce linting when testing.

It is **NOT** a goal of this PR to introduce a finalized styling. Instead, most rules in `.swift-format` are turned off to ensure the diff in this PR is minimal. We will introduce each rule individually with future PRs.


jakepetroules (approved):

Sources/Subprocess/Configuration.swift
```diff
@@ -404,15 +404,15 @@ extension Environment: CustomStringConvertible, CustomDebugStringConvertible {
...
                 }
-                #endif  // os(Windows)
+                #endif // os(Windows)
 
                 guard let delimiter = environmentString.firstIndex(of: "=") else {
                     continue
                 }
 
-                let key = String(environmentString[environmentString.startIndex..<delimiter])
+                let key = String(environmentString[environmentString.startIndex ..< delimiter])
```
parkera: I don't think I like this particular rule - it's pretty common to use ranges without spaces.

Sources/Subprocess/Configuration.swift
```diff
@@ -509,17 +509,17 @@ internal enum StringOrRawBytes: Sendable, Hashable {
     func createRawBytes() -> UnsafeMutablePointer<CChar> {
         switch self {
         case .string(let string):
-#if os(Windows)
+            #if os(Windows)
```
parkera, line 512: I'm sure we're very inconsistent about this one. I sort of like that `#if` is aligned differently - it calls out something that changes at a different time (compile time vs run time). Indenting them to me seems a little weird too, because we don't indent the lines under the 'else' like we would have if it were a traditional `if { }` block.

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -174,10 +183,10 @@ final class AsyncIO: @unchecked Sendable {
...
         // Post status to shutdown HANDLE
         PostQueuedCompletionStatus(
-            ioPort,         // CompletionPort
-            0,              // Number of bytes transferred.
-            shutdownPort,   // Completion key to post status
-            nil             // Overlapped
+            ioPort, // CompletionPort
+            0, // Number of bytes transferred.
+            shutdownPort, // Completion key to post status
+            nil // Overlapped
```
parkera: Personally I find the previous version more readable.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -488,7 +488,7 @@ public struct ProcessIdentifier: Sendable, Hashable {
         self.value = value
     }
 
-    internal func close() { /* No-op on Darwin */ }
+    internal func close() { /* No-op on Darwin */  }
```
parkera, line 491: Curious why this inserted an additional space.

Sources/Subprocess/Configuration.swift
iCharlesHu: Yeah fortunately the only settings are either
```
#if os(macOS)
var label = ""
#endif
```
or
```
#if os(macOS)
    var label = ""
#endif
```
I seem to see we use former more so I decided to go with that

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Me too, but there's no setting for swift-format to recognize that... There's also no "disable linting for the next n lines" either so we are out of options for now. I think `swift-format` desperately  need a `ignore next N lines` in addition to `ignore this file`.

iCharlesHu: Ignored this line form formatting.

Sources/Subprocess/Configuration.swift
iCharlesHu: Updated setting.

Issue #150, opened 2025-08-17
Process descriptor leak on error


jakepetroules:
In configuration.run, this is called:

```swift
// Close process file descriptor now we finished monitoring
execution.processIdentifier.close()
```

This needs to be moved up into a defer block immediately after `execution` is retrieved, or the process descriptors will leak when `monitorProcessTermination` throws.

Issue #145, opened 2025-08-15
Builtin standard inputs/outputs to share the current process' descriptors


cmcgee1024:
When using subprocess in the context of scripting it's very common to share the current process standard output, standard error, and sometimes even standard input. this is often what happens when running in a shell, and it allows interactive tools like vim to be run in this context if that's the goal the script.

It's possible to do this today with something like the following in a collected result invocation:

```
try await run(..., output: .fileDescriptor(FileDescriptor.standardInput, closeAfterSpawningProcess: false))
```

A similar pattern can be applied for standard error, and standard input.

This isn't very discoverable, it's verbose, and it's important that the file descriptor not be closed after spawning in most cases like this.

**Describe the solution you'd like**
Instead of this, there should be much simpler helper functions on the Input/Output protocols to hook into the standard input/output/error.

```
try await run(..., input: .standardInput, output: .standardOutput, error: .standardError)
```

This is shorter, clearer, and the file descriptors are left open until the calling process finishes instead of being accidentally closed.

cmcgee1024:
Here's a short mockup of these new helper computed properties:

```
extension OutputProtocol where Self == FileDescriptorOutput {
    /// Create a Subprocess output that writes output to the standard output of
    /// this process. The file descriptor should no be closed afterwards.
    public static var standardOutput: Self {
        Self.fileDescriptor(FileDescriptor.standardOutput, closeAfterSpawningProcess: false)
    }

    // Create a Subprocess output that write output to the standard error of
    /// this process. The file descriptor should no be closed afterwards.
    public static var standardError: Self {
        Self.fileDescriptor(FileDescriptor.standardError, closeAfterSpawningProcess: false)
    }
}

extension InputProtocol where Self == FileDescriptorInput {
    /// Create a Subprocess output that writes output to the standard output of
    /// this process. The file descriptor should no be closed afterwards.
    public static var standardInput: Self {
        Self.fileDescriptor(FileDescriptor.standardInput, closeAfterSpawningProcess: false)
    }
}
```

jakepetroules:
That feels like a nice, elegant solution.

PR #167, opened 2025-09-02
Introduce FileDescriptorInput.standardInput, FileDescriptorOutput.standardOutput, and FileDescriptorOutput.standardError to easily bind current process' descriptors to child processes

iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/145

While I was here:
Also resolves: https://github.com/swiftlang/swift-subprocess/issues/150


Sources/Subprocess/IO/Input.swift
```diff
@@ -168,6 +168,15 @@ extension InputProtocol where Self == FileDescriptorInput {
             closeAfterSpawningProcess: closeAfterSpawningProcess
         )
     }
+
+    /// Create a Subprocess input that reads from the standard input of
+    /// current process. The file descriptor should not be closed afterwards.
```
jakepetroules: Should this say "will not be closed" rather than "should not be closed"?

jakepetroules (approved):
Nice, I will use these in swiftlang/swift-build#638

Sources/Subprocess/IO/Input.swift
heckj: Preference for not using the future tense, and keeping the abstract to a single sentence, with the detail broken out with "a blank line" so that it shows up in the discussion section.
```suggestion
    /// current process. 
    /// 
    /// The file descriptor isn't closed afterwards.
```

iCharlesHu: Thanks @heckj !!

PR #168, opened 2025-09-02
Introduce preferredBufferSize parameter to allow custom buffer size when streaming subprocess output

iCharlesHu:
Currently, `Subprocess` always uses the same buffer size (system page size). This is an issue for executables with small but frequent outputs because `Subprocess` will always wait until it gathers buffer-size-bytes before it returns the read byte., causing streaming delays. Introduce the `preferredBufferSize` parameter to allow buffer size customization. Developers could choose a smaller buffer size if they anticipate that the output from the executable will be small.

resolves: https://github.com/swiftlang/swift-subprocess/issues/39

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -156,6 +156,11 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol, Error: Out
 ///   - platformOptions: The platform-specific options to use when running the executable.
 ///   - input: The input to send to the executable.
 ///   - error: How to manage executable standard error.
+///   - preferredBufferSize: The preferred size in bytes for the buffer used when reading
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/API.swift
```diff
@@ -160,6 +160,11 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol, Error: Out
 ///     when running the executable.
 ///   - input: The input to send to the executable.
 ///   - error: How to manager executable standard error.
+///   - preferredBufferSize: The preferred size in bytes for the buffer used when reading
```
jakepetroules, line 159: The parameter name (preferredOutputSize) and documentation (preferredBufferSize) disagree on the name.

jakepetroules (approved):
LGTM other than the docs typo (or parameter name type).

Issue #165, opened 2025-09-01
LineSequence.AsyncIterator should use a slice type for its underlyingBuffer to avoid unnecessary copies


jlukas:
**Describe the bug**
When reading the output of a subprocess by lines, the underlying buffer returns a single character at a time from the front of its buffer. This requires copying the buffer each time. Instead, a slice type could be used to avoid this copying.

**To Reproduce**
My call is something like:

```swift
Subprocess.run(…) { execution, stdin, stdout, stderr in
    …
    for try await line in stdout.lines() {
        …
    }
}
```

Applying [this diff](https://github.com/user-attachments/files/22081730/slice.txt) reduces the time taken by subprocess from ~2 seconds to ~1 second (for calls with lots of output).

**Environment (please complete the following information):**
 - macOS 15
 - swift from various betas of Xcode 26, including: `swift-driver version: 1.127.14.1 Apple Swift version 6.2 (swiftlang-6.2.0.19.9 clang-1700.3.19.1)`

Issue #160, opened 2025-08-24
testSubprocessDoesNotInheritVeryHighFileDescriptors and testLineSequence are extremely slow


jakepetroules:
Following #79, testSubprocessDoesNotInheritVeryHighFileDescriptors and testLineSequence appear to be extremely slow on amazonlinux2 and focal (which notably contain older glibc missing modern features related to process descriptors and close_range). These CI jobs are taking multiple hours to complete.

jakepetroules:
Note: this might reproduce only in containers by default.

Amazon Linux 2 container:

```sh
$ ulimit -Sn
2147483584
```

Real Debian VM:

```sh
$ ulimit -Sn
1024
```

Bumping up the limit and disabling use of close_range (simulating older glibc and/or kernel) demonstrates the problem.

iCharlesHu:
@weissi can you elaborate on why do we need to calculate `_highest_possibly_open_fd_dir_linux`? Can we simply iterate through `/dev/fd` like we already do in `_highest_possibly_open_fd_dir_linux` and simply close the ones listed?

Issue #159, opened 2025-08-23
closeAfterSpawningProcess can double-close file descriptors


jakepetroules:
The safelyCloseMultiple function assumes all 6 fds are unique, but the same underlying FileDescriptor could be wrapped in multiple IODescriptor instances.

Subprocess should detect such invalid input and produce an error if the same fd is given multiple times with closeAfterSpawningProcess set to true.

PR #171, opened 2025-09-03
Various bug fixes throughout the code base

iCharlesHu:
- Instead of calling `[UInt8].removeFirst` on `LineSequence`’s `underlyingBuffer`, we now use an index-based approach. This change improves overall streaming performance by eliminating the `O(n)` `.removeFirst()` operation.
- The `safelyCloseMultiple` function has been updated to handle cases where the same `IODescriptor` is passed to multiple input/outputs.
- On Linux, `_highest_possibly_open_fd_dir_linux()` has been updated to `_close_open_fd_linux()` to more efficiently emulate `POSIX_SPAWN_CLOEXEC_DEFAULT`.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/165, https://github.com/swiftlang/swift-subprocess/issues/160, https://github.com/swiftlang/swift-subprocess/issues/159

### DIFF
--- a/Sources/_SubprocessCShims/process_shims.c
+++ b/Sources/_SubprocessCShims/process_shims.c
@@ -431,14 +431,13 @@ static int _positive_int_parse(const char *str) {
 }
 
 #if defined(__linux__)
-// Linux-specific version that uses syscalls directly and doesn't allocate heap memory.
-// Safe to use after vfork() and before execve()
-static int _highest_possibly_open_fd_dir_linux(const char *fd_dir) {
-    int highest_fd_so_far = 0;
+/// Set `FD_CLOEXEC` on all open file descriptors listed under `fd_dir` so
+/// they are automatically closed upon `execve()`.
+/// Safe to use after `vfork()` and before `execve()`
+static void _set_cloexec_to_open_fds(const char *fd_dir) {
     int dir_fd = open(fd_dir, O_RDONLY);
     if (dir_fd < 0) {
-        // errno set by `open`.
-        return -1;
+        return;
     }
 
     // Buffer for directory entries - allocated on stack, no heap allocation
@@ -450,49 +449,37 @@ static int _highest_possibly_open_fd_dir_linux(const char *fd_dir) {
             if (errno == EINTR) {
                 continue;
             } else {
-                // `errno` set by _getdents64.
-                highest_fd_so_far = -1;
                 close(dir_fd);
-                return highest_fd_so_far;
+                return;
             }
         }
         if (bytes_read == 0) {
             close(dir_fd);
-            return highest_fd_so_far;
+            return;
         }
         long offset = 0;
         while (offset < bytes_read) {
             struct linux_dirent64 *entry = (struct linux_dirent64 *)(buffer + offset);
-
             // Skip "." and ".." entries
             if (entry->d_name[0] != '.') {
-                int number = _positive_int_parse(entry->d_name);
-                if (number > highest_fd_so_far) {
-                    highest_fd_so_far = number;
+                int fd = _positive_int_parse(entry->d_name);
+                if (fd > STDERR_FILENO && fd != dir_fd) {
+                    int flags = fcntl(fd, F_GETFD);
+                    if (flags >= 0) {
+                        // Set FD_CLOEXEC on every open fd so they are closed after exec()
+                        fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
+                    }
                 }
             }
-
             offset += entry->d_reclen;
         }
     }
-
-    close(dir_fd);
-    return highest_fd_so_far;
 }
 #endif
 
-// This function is only used on systems with Linux kernel 5.9 or lower.
-// On newer systems, `close_range` is used instead.
+// This function is only used on non-Linux systems.
 static int _highest_possibly_open_fd(void) {
-#if defined(__linux__)
-    int hi = _highest_possibly_open_fd_dir_linux("/dev/fd");
-    if (hi < 0) {
-        hi = sysconf(_SC_OPEN_MAX);
-    }
-#else
-    int hi = sysconf(_SC_OPEN_MAX);
-#endif
-    return hi;
+    return sysconf(_SC_OPEN_MAX);
 }
 
 int _subprocess_fork_exec(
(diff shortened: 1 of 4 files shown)


Sources/_SubprocessCShims/process_shims.c
```diff
@@ -431,68 +431,70 @@ static int _positive_int_parse(const char *str) {
...
+        int dir_fd = open(fd_dir, O_RDONLY);
+        if (dir_fd < 0) {
+            return;
+        }
+
+        int found_any_fds = 0;
+
         long bytes_read = _getdents64(dir_fd, (struct linux_dirent64 *)buffer, sizeof(buffer));
         if (bytes_read < 0) {
+            close(dir_fd);
```
jakepetroules: Any particular reason why we call close directly rather than set O_CLOEXEC on the file descriptor?

I notice Android's implementation of POSIX_SPAWN_CLOEXEC_DEFAULT uses close_range with the cloexec flag, as well as falling back to looping, but again setting cloexec rather than directly closing.

jakepetroules: I'm struggling to think of a reason _why_ you wouldn't just close right away -- there is only a single active thread between fork and exec, and this interface controls all the code that runs there (same for Android's posix_spawn), but there must be a reason why they opted for cloexec?

https://android.googlesource.com/platform/bionic/+/master/libc/bionic/spawn.cpp#52

iCharlesHu: I did some searching and it seems in general people prefer `O_CLOEXEC`:

https://tzimmermann.org/2017/08/17/file-descriptors-during-fork-and-exec/
https://danwalsh.livejournal.com/53603.html

It seems the Android fallback is to use `close_range` if possible, then try setting `O_CLOEXEC` instead of `close()` in a loop. I think we should do the same.

iCharlesHu:
`focal` and `amazonlinux2` finished in 2 minutes 🥳 (down from ~36 minutes and 1 hour respectively 💀)

Sources/_SubprocessCShims/process_shims.c
jakepetroules: To be precise, they use `close_range` with `CLOSE_RANGE_CLOEXEC`, then fall back to setting `O_CLOEXEC` in a loop.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -692,13 +679,22 @@ int _subprocess_fork_exec(
         rc = closefrom(pipefd[1] + 1);
         #endif
         if (rc != 0) {
-            // close_range failed (or doesn't exist), fall back to close()
-            for (int fd = STDERR_FILENO + 1; fd <= _highest_possibly_open_fd(); fd++) {
+            #if defined(__linux__)
```
jakepetroules, line 682: Annoying GitHub won't let me comment on the right line but on line 671-672, for consistency with the other changes you just made, you'll want to pass `CLOSE_RANGE_CLOEXEC` to the two close_range calls above on Linux/FreeBSD.

OpenBSD doesn't have any efficient way to set CLOEXEC on a range of fds, so closefrom will have to do.

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -450,49 +449,37 @@ static int _highest_possibly_open_fd_dir_linux(const char *fd_dir) {
...
             if (entry->d_name[0] != '.') {
-                int number = _positive_int_parse(entry->d_name);
-                if (number > highest_fd_so_far) {
-                    highest_fd_so_far = number;
+                int fd = _positive_int_parse(entry->d_name);
+                if (fd > STDERR_FILENO && fd != dir_fd) {
+                    int flags = fcntl(fd, F_GETFD);
+                    if (flags >= 0) {
+                        // Set FD_CLOEXEC on every open fd so they are closed after exec()
+                        fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
```
weissi, line 470: If you're doing a syscall anyway, why don't you just close it?

Sources/_SubprocessCShims/process_shims.c
```diff
@@ -692,13 +679,22 @@ int _subprocess_fork_exec(
...
-            for (int fd = STDERR_FILENO + 1; fd <= _highest_possibly_open_fd(); fd++) {
+            #if defined(__linux__)
+            _set_cloexec_to_open_fds("/dev/fd");
+            #else
+            // close_range failed (or doesn't exist), fall back to setting FD_CLOEXEC
+            int highest_open_fd = _highest_possibly_open_fd();
+            for (int fd = STDERR_FILENO + 1; fd <= highest_open_fd; fd++) {
                 // We must NOT close pipefd[1] for writing errors
                 if (fd != pipefd[1]) {
-                    close(fd);
```
weissi, line 699: `close` feels better here and will be cheaper overall. Otherwise you need to do N syscalls to mark as CLOEXEC and later on there's more work necessary to close them. Why not just close?

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: For two reasons:

1. I've searched around (close vs setting FD_CLOEXEC) and it seems in general `FD_CLOEXEC` is preferred for this situation:

https://utcc.utoronto.ca/~cks/space/blog/unix/ForkFDsAndRaces

https://docs.fedoraproject.org/en-US/defensive-coding/tasks/Tasks-Descriptors/#sect-Defensive_Coding-Tasks-Descriptors-Child_Processes

2. A more practical reason is I don't think we can safely `close()` while traversing `/dev/fd` before `close()` modifies that directory.

Sources/_SubprocessCShims/process_shims.c
iCharlesHu: Ditto to above: it seems the general consensus is to use `FD_CLOEXEC` instead of `close()`. We would have to issue N close() calls here, so without `close_range`, it doesn't really save much.

Sources/_SubprocessCShims/process_shims.c
jakepetroules: > I've searched around (close vs setting FD_CLOEXEC) and it seems in general FD_CLOEXEC is preferred for this situation:

Both of those articles I think are really talking about the pre-fork case. I think in our specific situation where we've already called fork (and thus have exactly one thread and no concurrency considerations), and we can guarantee that we will not call fork again, and that we will call exec imminently, there's no practical difference (that I can see).

> A more practical reason is I don't think we can safely close() while traversing /dev/fd before close() modifies that directory.

This seems like a great point though!

jakepetroules (approved):

Sources/_SubprocessCShims/process_shims.c
weissi: > A more practical reason is I don't think we can safely close() while traversing /dev/fd before close() modifies that directory.

That should _not_ happen and doesn't happen in the original code from `swift-sdk-generator`. How the original code works is:
1. Get the maximum fd
2. Loop from 3 to maximum fd and close

That's completely fine. What is _not_ fine (but that's not what the original code does) is to close whilst iterating of course.

PR #174, opened 2025-09-04
Make Environment keys case-insensitive on Windows

### LINKED ISSUE #134: Environment keys should not be case sensitive on Windows
On Windows, the environment keys are case-insensitive. Instead of using a `[String: String]` type, we should define an `EnvironmentKey` structure that handles the case-insensitivity considerations across platforms.

This is rather important, as we've run into _[real bugs](https://github.com/swiftlang/swift-build/issues/296)_ in the build system as a result of treating the environment as a case sensitive dictionary.

See https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/EnvironmentKey.swift for an example of how we did this in Swift Build.

jakepetroules:
This is how the platform treats them, and helps avoid issues where indexing into an Environment dictionary with the wrong casing fails to return a value.

Closes #134

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -401,8 +401,8 @@ extension Arguments: CustomStringConvertible, CustomDebugStringConvertible {
 /// A set of environment variables to use when executing the subprocess.
 public struct Environment: Sendable, Hashable {
     internal enum Configuration: Sendable, Hashable {
-        case inherit([String: String])
-        case custom([String: String])
+        case inherit([Key: String])
+        case custom([Key: String])
         #if !os(Windows)
         case rawBytes([[UInt8]])
         #endif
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/Configuration.swift
```diff
@@ -448,13 +448,87 @@ extension Environment: CustomStringConvertible, CustomDebugStringConvertible {
...
             }
             return results
         }
     }
 }
 
+/// A key used to access values in an ``Environment``.
+///
+/// This type respects the compiled platform's case sensitivity requirements.
+public struct EnvironmentKey {
```
iCharlesHu: Since this is a public type, I believe Swift prefers the nested `Environment.Key` over the top level `EnvironmentKey`

jakepetroules: Done

iCharlesHu (approved):

PR #176, opened 2025-09-06
Subprocess pipelines

cmcgee1024:
Create complex pipeline stages and run the pipeline configuration using a single I/O configuration based on the usual Input, Output, and Error options provided by the existing run() functions.

Each stage can be either a process or a Swift function. For processes there are the typical executable, arguments, or configuration parameters. There is an additional process option to rewire standard output and standard error in typical ways, such as merging the two to standard output, or replacing standard output with standard error.

A Swift function stage can read from its input, and write to its output/error streams much like a process, but provides the full Swift power in the current process to parse JSON and do other powerful things. Similarly, an exit code can be returned.

Provide a rich set of top-level functions, methods and overloaded operators to construct and execute the pipelines.

```.swift
// Using .finally() method
var config = pipe(
    .name("echo"),
    arguments: ["Hello World"]
).stage(
    executable: .name("wc"),
    arguments: ["-c"]
).finally(
    output: .string(limit: 1024)
)

// Using | and |> operators
config = pipe(
   .name("echo"),
    arguments: ["Hello World"]
)
| (
   .name("wc"),
    arguments: ["-c"]
)
|> .string(limit: 1024)

let result = try await config.run()
print(result.standardOutput) // "12"
```

### DIFF
--- a/PIPE_CONFIGURATION_USAGE.md
+++ b/PIPE_CONFIGURATION_USAGE.md
@@ -26,7 +26,7 @@ This eliminates interim `PipeConfiguration` objects with discarded I/O and makes
 ```swift
 // Using .finally() method
 let config = pipe(
-    executable: .name("echo"),
+    .name("echo"),
     arguments: ["Hello World"]
 ).finally(
     output: .string(limit: .max)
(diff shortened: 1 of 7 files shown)


cmcgee1024:
The macOS build error seems unrelated to the changes here. It's failing on an ambiguous Region type.

Issue #59, opened 2025-06-04
There doesn't seem to be a way to collect merged stdout+stderr


jakepetroules:
The API only allows collecting stdout and stderr separately, not as a single stream. This makes it impossible to build something like a Terminal app where the stdout+stderr of a process should be interleaved in the order of emission.

jakepetroules:
How could that possibly produce the same result w.r.t. ordering?

iCharlesHu:
Ditto to @FranzBusch 's answer. The current design uses `AsyncSequence` specifically so you can utilize the rest of `async-algorithm` for things like this. @jakepetroules can you elaborate what's the issue you are seeing with `merge`?

FranzBusch:
@jakepetroules is right that the ordering with `merge` is not guaranteed to be the same as the subprocess that produced it since it is up to the executors scheduling. If we want to achieve the same ordering as the subprocess then we need to pass probably the same FD to as stdout and stderr and then offer a single async sequence for it.

jakepetroules:
Actually, this approach works with the current API surface:

```swift
let (readEnd, writeEnd) = try FileDescriptor.pipe()
return try await readEnd.closeAfter {
    // Direct both stdout and stderr to the same fd. Only set `closeAfterSpawningProcess` on one of the outputs so it isn't double-closed (similarly avoid using closeAfter for the same reason).
    let result = try await Subprocess.run(
        .path(FilePath(url.filePath.str)),
        arguments: .init(arguments),
        environment: environment.map { .custom(.init($0)) } ?? .inherit,
        workingDirectory: (currentDirectoryURL?.filePath.str).map { FilePath($0) } ?? nil,
        output: .fileDescriptor(writeEnd, closeAfterSpawningProcess: true),
        error: .fileDescriptor(writeEnd, closeAfterSpawningProcess: false))
    // read from `readEnd` here...
}
```

jakepetroules:
Actually I'll leave this open to allow consideration of a higher level API for this (which I think would be a reasonable addition). That said, there is a workaround for now, it's just a little lower level than I'd prefer, especially since the client has to provide their own mechanism to read from the fd in an async-friendly manner.

jakepetroules:
Now that we have AsyncIO with platform-specific implementations, I think this would make even more sense to provide.

luispadron:
+1 to having this, I needed to get merged stdout / stderr just now and it was not obvious how to do this with the existing API

PR #180, opened 2025-09-10
Introudce CombinedErrorOutput

iCharlesHu:
`CombinedErrorOutput` is a concrete error output type for subprocesses that combines the standard error output with the standard output stream.

When `CombinedErrorOutput` is used as the error output for a subprocess, both standard output and standard error from the child process are merged into a single output stream. This is equivalent to using shell redirection like `2>&1`.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/59


Sources/Subprocess/IO/Output.swift
```diff
@@ -286,6 +286,57 @@ extension OutputProtocol where Self == BytesOutput {
...
+
+/// A concrete error output type for subprocesses that combines the standard error
+/// output with the standard output stream.
+///
+/// When `CombinedErrorOutput` is used as the error output for a subprocess, both
+/// standard output and standard error from the child process are merged into a
+/// single output stream. This is equivalent to using shell redirection like `2>&1`.
+///
+/// This output type is useful when you want to capture or redirect both output
+/// streams together, making it easier to process all subprocess output as a unified
```
jakepetroules: nit: I'd replace "easier to process" with "possible to process", since the behavior is inherently different, so it's not as if this API is simply a matter of convenience.

Sources/Subprocess/Configuration.swift
```diff
@@ -1213,3 +1233,32 @@ extension Set {
         return self.remove(element) != nil
     }
 }
+
+#if canImport(WinSDK)
+extension HANDLE {
+    func duplicate() throws -> HANDLE {
```
jakepetroules, line 1239: What is the reason to duplicate the underlying platform file descriptor / HANDLE versus using the same one directly? Is it simply because it makes easier to model things using Swift's lifetime features, or is there another reason?

iCharlesHu: It's easier and cleaner to model with `~Copyable`. When we duplicate, each output gets its own `CreatedPiped`, just like other cases, so we don't have to make special cases to not close fd, etc.

Issue #185, opened 2025-09-22
[Feature] `Environment` does not allow filtering


compnerd:
**Is your feature request related to a problem? Please describe.**
It is not possible to filter certain environment variables from the inherited environment for the subprocess. With Foundation's `Process`, you could set the element to `nil` to unset the variable, but no such equivalent exists for `Environment` in `Subprocess`.

**Describe the solution you'd like**
Either of two (immediately obvious) approaches:
1. permit `.updating(["VARIABLE":nil])`
2. add `.filtering("VARIABLE")`

**Describe alternatives you've considered**
Both of the solutions in the description would be feasible. Given that the core function for this is not exposed, it is not possible to handle this outside of the package.

**Additional context**

PR #193, opened 2025-10-01
Change Environment.updating to accept [Key: String?]

iCharlesHu:
A nil value removes the corresponding key from being inherited by child processes.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/185


Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -310,6 +310,82 @@ extension SubprocessIntegrationTests {
         )
     }
 
+    @Test func testEnvironmentInheritOverrideUnsetValue() async throws {
+        #if os(Windows)
+        // First insert our dummy environment variable
+        "SubprocessTest".withCString(encodedAs: UTF16.self) { keyW in
+            "value".withCString(encodedAs: UTF16.self) { valueW in
+                SetEnvironmentVariableW(keyW, valueW);
```
jakepetroules: nit: some stray semicolons here.

I could also see the set-defer-unset pattern being problematic for concurrently running tests. We had the same problem in Swift Build and adopted a mock-based approach with Swift Concurrency task locals, which works pretty nicely. Not suggesting it's something that needs to be fixed now, but just a thought for future enhancements.

jakepetroules (approved):
Thanks!

Tests/SubprocessTests/IntegrationTests.swift
iCharlesHu: Yeah unfortunately this is technically a global state. Fortunately we only rely on this key being present so removing it is not strictly necessary at this point.

Issue #182, opened 2025-09-15
Swift Subprocess hangs in release builds but works in debug builds


kattouf:
**Describe the bug**

Swift Subprocess library hangs indefinitely when built in release configuration, while working correctly in debug configuration. The `Subprocess.run()` call never completes when running a simple `ls` command in release builds.

**To Reproduce**
What executable was running? `ls`
2. What arguments are passed in? No arguments (just `ls`)
3. What environment values are used? Default environment

Code example:
```swift
let result = try await Subprocess.run(.name("ls"), output: .string(limit: 4096))
```

I reproduced this issue here: https://github.com/kattouf/swift-subprocess-hang-report

**Expected behavior**
The `Subprocess.run()` call should complete successfully and return the process result with output, process identifier, and termination status.

**Environment (please complete the following information):**
 - OS macOS - 15
 - Swift version - swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0

**Additional context**
I reproduced this issue with both release 0.1.0 and the main branch.

kattouf:
I bisected the problem and found that it starts reproducing from commit 6eb91f6651e9d029b96fcf16b047dfb7c26da7f7, which is the merge of https://github.com/swiftlang/swift-subprocess/pull/158

PR #194, opened 2025-10-02
Update WorkQueue.queue to be a reference type instead of an array.

iCharlesHu:
We pass WorkQueue.queue 'by reference' through inout parameters and across thread suspension points. We noticed a stale value issue in release builds when passing Array with inout, likely due to the copy-on-write mechanism. To avoid stale values after a thread resumes from suspension, switch to a dedicated reference type Queue.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/182

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -28,14 +28,14 @@ jobs:
           # Test dependencies
           yum install -y procps
         fi
-      linux_build_command: 'swift-format lint -s -r --configuration ./.swift-format . && swift test && swift test --disable-default-traits'
+      linux_build_command: 'swift-format lint -s -r --configuration ./.swift-format . && swift test && swift test -c release && swift test --disable-default-traits'
       windows_swift_versions: '["6.1", "nightly-main"]'
       windows_build_command: |
         Invoke-Program swift test
         Invoke-Program swift test --disable-default-traits
       enable_macos_checks: true
       macos_xcode_versions: '["16.3"]'
-      macos_build_command: 'xcrun swift-format lint -s -r --configuration ./.swift-format . && xcrun swift test && xcrun swift test --disable-default-traits'
+      macos_build_command: 'xcrun swift-format lint -s -r --configuration ./.swift-format . && xcrun swift test && xcrun swift test -c release && xcrun swift test --disable-default-traits'
       enable_linux_static_sdk_build: true
       linux_static_sdk_versions: '["6.1", "nightly-6.2"]'
       linux_static_sdk_build_command: |
--- a/Sources/Subprocess/Thread.swift
+++ b/Sources/Subprocess/Thread.swift
@@ -77,12 +77,33 @@ private struct BackgroundWorkItem {
 // We can't use Mutex directly here because we need the underlying `pthread_mutex_t` to be
 // exposed so we can use it with `pthread_cond_wait`.
 private final class WorkQueue: Sendable {
-    private nonisolated(unsafe) var queue: [BackgroundWorkItem]
+    // Queue needs to be a reference type because we pass it inout
+    final class Queue: Sendable {
+        internal nonisolated(unsafe) var queue: [BackgroundWorkItem]
+
+        var isEmpty: Bool { self.queue.isEmpty }
+
+        func append(_ item: BackgroundWorkItem) {
+            self.queue.append(item)
+        }
+
+        func removeFirst() -> BackgroundWorkItem {
+            return self.queue.removeFirst()
+        }
+
+        func removeAll() { self.queue.removeAll() }
+
+        init() {
+            self.queue = []
+        }
+    }
+
+    private nonisolated(unsafe) var queue: Queue
     internal nonisolated(unsafe) let mutex: UnsafeMutablePointer<MutexType>
     internal nonisolated(unsafe) let waitCondition: UnsafeMutablePointer<ConditionType>
 
     init() {
-        self.queue = []
+        self.queue = Queue()
         self.mutex = UnsafeMutablePointer<MutexType>.allocate(capacity: 1)
         self.waitCondition = UnsafeMutablePointer<ConditionType>.allocate(capacity: 1)
         #if canImport(WinSDK)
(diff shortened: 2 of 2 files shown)


Sources/Subprocess/Thread.swift
```diff
@@ -77,12 +77,33 @@ private struct BackgroundWorkItem {
 // We can't use Mutex directly here because we need the underlying `pthread_mutex_t` to be
 // exposed so we can use it with `pthread_cond_wait`.
 private final class WorkQueue: Sendable {
-    private nonisolated(unsafe) var queue: [BackgroundWorkItem]
+    // Queue needs to be a reference type because we pass it inout
+    final class Queue: Sendable {
+        internal nonisolated(unsafe) var queue: [BackgroundWorkItem]
```
weissi, line 82: This is incorrect because this type isn't Sendable. It allows mutation without synchronousiarion.

furthermore, the reference gets passed inout.

We should also add a test that shows where precisely you need reference semantics and why

iCharlesHu: > This is incorrect because this type isn't Sendable. It allows mutation without synchronousiarion.

This type is exclusively used under a pthread_mutex_t. It is for that reason `@unchecked Sendable`.

> We should also add a test that shows where precisely you need reference semantics and why

I don't quite understand what do you mean. Could you suggest a test?

weissi: >> This is incorrect because this type isn't Sendable. It allows mutation without synchronousiarion.
> This type is exclusively used under a pthread_mutex_t. It is for that reason @unchecked Sendable.

I get that this is intent but we should prove this to the compiler and not use `@unchecked`, that's unsafe. We have helpers like NIOLockedValueBox, OSAllocatedUnfairLock and Mutex which can make this safe. And we should use those


> I don't quite understand what do you mean. Could you suggest a test?

I assume that the reason for this pull request is that something didn't work as expected. The test should test that before it doesn't work as expected but now (with this change) it does.

iCharlesHu: > I get that this is intent but we should prove this to the compiler and not use @unchecked, that's unsafe. We have helpers like NIOLockedValueBox, OSAllocatedUnfairLock and Mutex which can make this safe. And we should use those

We can't use `OSAllocatedUnfairLock` nor `Mutex` directly here because we need access to the underlying `pthread_mutex_t` to pass to `pthread_cond_wait`. Because of this, I deliberately made `WorkQueue` NOT a generic `LockedBox` type. Instead, we use `Mutex` wherever we can and only use `WorkQueue` for very specific purposes. On the other hand, what's wrong with using `@unchecked Sendable` this way? Isn't this the canonical use case for `@unchecked Sendable` (when you explicitly use external synchronization instead of relying on Swift itself)?

> I assume that the reason for this pull request is that something didn't work as expected. The test should test that before it doesn't work as expected but now (with this change) it does.

This change resolves issue #182. Currently, the `run()` function hangs on release builds on Darwin, and it no longer hangs on release builds with this PR. To prevent future regressions, I have added `swift test -c release` to the CI pipeline.

.github/workflows/pull_request.yml
```diff
@@ -28,14 +28,14 @@ jobs:
           # Test dependencies
           yum install -y procps
         fi
-      linux_build_command: 'swift-format lint -s -r --configuration ./.swift-format . && swift test && swift test --disable-default-traits'
+      linux_build_command: 'swift-format lint -s -r --configuration ./.swift-format . && swift test && swift test -c release && swift test --disable-default-traits'
       windows_swift_versions: '["6.1", "nightly-main"]'
       windows_build_command: |
         Invoke-Program swift test
```
jakepetroules, line 34: Can you add `Invoke-Program swift test -c release` below this line so we have release coverage for Windows as well?

Sources/Subprocess/Thread.swift
```diff
@@ -94,14 +115,14 @@ private final class WorkQueue: Sendable {
         #endif
     }
 
-    func withLock<R>(_ body: (inout [BackgroundWorkItem]) throws -> R) rethrows -> R {
+    func withLock<R>(_ body: (inout Queue) throws -> R) rethrows -> R {
```
jakepetroules, line 118: What's the point of using inout anymore if it's a reference type?

Sources/Subprocess/Thread.swift
jakepetroules, line 149: I think the actual issue is that we sleep the thread here using the condition variable while still holding an exclusive reference to the queue, and resuming the thread later returns to this point and corrupts state.

I wonder if we can resolve this concern by doing something similar to what @weissi said and decoupling the work queue from the condition variable logic using an abstraction similar to what I proposed here: https://forums.swift.org/t/thoughts-on-a-conditionvariable-type/81123

I _think_ that may solve the problem by ensuring we release ownership of the inout before suspending the thread, due to the way the API is structured -- you can't interact with the condition variable and suspend inside the user-provided closure. Though in your case I think you'd want a variant of my `wait` method that lets you suspend prior to running block instead of after.

Then define WorkQueue as:

```swift
private final class WorkQueue: Sendable {
    private let queue: ConditionVariable<[BackgroundItem]>
}
```

jakepetroules (dismissed):

Sources/Subprocess/Thread.swift
weissi: Okay, but if you build a general purpose work queue, then I'd suggest

- put it into its own file
- Make it actually Sendable (currently it is not and just disables compiler checking)
- Add some real tests for it

itingliu: Looking at this code, it looks like `Queue` isn't actually Sendable like @weissi pointed out. The callsites however can use it safely as long as it is always guarded by a pthread_mutex_t. 

Is my understanding correct? If so, then I tend to agree that marking this type `Sendable` is misleading. This code would still work fine even if we remove `Sendable` from `Queue` since you already declared the var as `nonisolated(unsafe)` below:
```swift
private nonisolated(unsafe) var queue: Queue
```

weissi: My position is: Anything `@unchecked`/`unsafe` needs to be avoided. That's just switching off compiler checking.

A narrow exception is if you're creating new, reusable components (such as a generic queue) with an extensive test suite that cannot be used unsafely through their API, then that's fine.

For example, AsyncProcess doesn't contain `@unchecked`s at all

PR #195, opened 2025-10-03
Add extern "C" to header process_shims.h

j-hui:
Previous, we would encounter missing symbol errors like:

          "_subprocess_vm_size()", referenced from:
              closure #1 () -> Swift.Int in variable initialization expression of Subprocess.(_pageSize in _EDE99368FF43462E0C75AF3C9D4D8F2D) : Swift.Int in AsyncBufferSequence.swift.o
           NOTE: found '__subprocess_vm_size' in process_shims.c.o, declaration possibly missing 'extern "C"'

We can reproduce this by running:

        xcrun swift build -Xswiftc -cxx-interoperability-mode=default --build-tests

Adding `extern "C"` fixes these issues.

rdar://161624906

### DIFF
--- a/Sources/_SubprocessCShims/include/process_shims.h
+++ b/Sources/_SubprocessCShims/include/process_shims.h
@@ -35,6 +35,10 @@
 #include <sys/wait.h>
 #endif // TARGET_OS_LINUX || TARGET_OS_FREEBSD
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #if __has_include(<mach/vm_page_size.h>)
 vm_size_t _subprocess_vm_size(void);
 #endif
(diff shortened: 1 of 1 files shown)


Sources/_SubprocessCShims/include/process_shims.h
```diff
@@ -35,11 +35,18 @@
 #include <sys/wait.h>
 #endif // TARGET_OS_LINUX || TARGET_OS_FREEBSD
 
+#ifdef __cplusplus
+#define SWIFTSUBPROCESS_EXPORT extern "C"
+#else
+#define SWIFTSUBPROCESS_EXPORT extern
+#endif
+
```
compnerd, line 41: I wonder if this would be better served with a:

```c++
extern "C" {
```

block rather than annotating each of the declarations.

j-hui: That's what I originally had, and then I saw `#include <Windows.h>` and realized I couldn't just wrap the entire thing with `extern "C"`.

... But on second thought, I could just hoist that up to the top, and do the `extern "C"` thing you suggest (which is a cleaner patch).

compnerd: Right, as long as the includes come outside of the block, it is simpler. That also avoids the pitfall of missing the annotation on a future declaration.

iCharlesHu (approved):

PR #196, opened 2025-10-05
CMake fixes to allow use in dependent projects

owenv:
1. Define SUBPROCESS_ASYNCIO_DISPATCH on macOS/BSDs to fix a build failure
2. Create SwiftSubprocessConfig.cmake to allow dependent CMake projects to find exported targets

### DIFF
--- a/Sources/Subprocess/CMakeLists.txt
+++ b/Sources/Subprocess/CMakeLists.txt
@@ -41,13 +41,13 @@ elseif(APPLE)
     Platforms/Subprocess+Darwin.swift
     Platforms/Subprocess+Unix.swift)
   target_compile_options(Subprocess PRIVATE
-    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-D SUBPROCESS_ASYNCIO_DISPATCH>")
+    "$<$<COMPILE_LANGUAGE:Swift>:-DSUBPROCESS_ASYNCIO_DISPATCH>")
 elseif(FREEBSD OR OPENBSD)
   target_sources(Subprocess PRIVATE
     Platforms/Subprocess+BSD.swift
     Platforms/Subprocess+Unix.swift)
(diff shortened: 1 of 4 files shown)


Sources/Subprocess/CMakeLists.txt
```diff
@@ -40,10 +40,14 @@ elseif(APPLE)
     Platforms/Subprocess+BSD.swift
     Platforms/Subprocess+Darwin.swift
     Platforms/Subprocess+Unix.swift)
+  target_compile_options(Subprocess PRIVATE
+    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-D SUBPROCESS_ASYNCIO_DISPATCH>")
```
compnerd: If you are using the joined form (`-DSUBPROCESS_ASYNCIO_DISPATCH`) which is required for Swift, I don't think that you should use `SHELL`:. The deduplication should be safe and fine.


```suggestion
  target_compile_options(Subprocess PRIVATE
    "$<$<COMPILE_LANGUAGE:Swift>:-DSUBPROCESS_ASYNCIO_DISPATCH>")
```

Sources/Subprocess/CMakeLists.txt
```diff
@@ -40,10 +40,14 @@ elseif(APPLE)
...
 elseif(FREEBSD OR OPENBSD)
   target_sources(Subprocess PRIVATE
     Platforms/Subprocess+BSD.swift
     Platforms/Subprocess+Unix.swift)
+  target_compile_options(Subprocess PRIVATE
+    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-D SUBPROCESS_ASYNCIO_DISPATCH>")
```
compnerd:
```suggestion
    "$<$<COMPILE_LANGUAGE:Swift>:-DSUBPROCESS_ASYNCIO_DISPATCH>")
```

PR #198, opened 2025-10-15
Don't suspend/resume the thread while holding an exclusive inout reference

### LINKED ISSUE #182: Swift Subprocess hangs in release builds but works in debug builds

jakepetroules:
We pass WorkQueue.queue 'by reference' through inout parameters and across thread suspension points. We noticed a stale value issue in release builds when passing Array with inout, likely due to the copy-on-write mechanism. To avoid stale values after a thread resumes from suspension, sleep the worker thread only after releasing the exclusive inout reference, and re-obtain it after resuming.

Resolves #182

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -1166,8 +1166,20 @@ extension Configuration {
             DeleteProcThreadAttributeList(attributeList)
         }
 
-        var handles = Array(inheritedHandles)
-        handles.withUnsafeMutableBufferPointer { inheritedHandlesPtr in
+        let handles = Array(inheritedHandles)
+
+        // Manually allocate an array instead of using withUnsafeMutableBufferPointer, since the
+        // UpdateProcThreadAttribute documentation states that the lpValue pointer must persist
+        // until the attribute list is destroyed using DeleteProcThreadAttributeList.
+        let handlesPointer = UnsafeMutablePointer<HANDLE>.allocate(capacity: handles.count)
+        defer {
+            handlesPointer.deinitialize(count: handles.count)
+            handlesPointer.deallocate()
+        }
+        handlesPointer.initialize(from: handles, count: handles.count)
+        let inheritedHandlesPtr = UnsafeMutableBufferPointer(start: handlesPointer, count: handles.count)
+
+        do {
             _ = UpdateProcThreadAttribute(
                 attributeList,
                 0,
--- a/Sources/Subprocess/Thread.swift
+++ b/Sources/Subprocess/Thread.swift
@@ -101,7 +101,8 @@ private final class WorkQueue: Sendable {
     }
 
     private func withUnsafeUnderlyingLock<R>(
-        _ body: (UnsafeMutablePointer<MutexType>, inout [BackgroundWorkItem]) throws -> R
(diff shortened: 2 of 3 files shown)


jakepetroules:
This is a more targeted alternative to #194

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -155,7 +155,7 @@ extension Configuration {
                                         spawnContext.startupInfo.pointer(to: \.StartupInfo)!,
                                         &processInfo
                                     )
-                                    return (result, processInfo, GetLastError())
+                                    return (result, processInfo, result ? ERROR_SUCCESS : GetLastError())
```
jakepetroules: Most Windows functions do NOT set GetLastError to 0 on success (and the CreateProcessW docs don't seem to indicate that it is one), so we need to ensure we aren't getting false positive failures, which I saw regularly in release mode.

Sources/Subprocess/Thread.swift
```diff
@@ -114,20 +115,22 @@ private final class WorkQueue: Sendable {
             pthread_mutex_unlock(mutex)
         }
         #endif
+        if condition(&queue) {
+            #if canImport(WinSDK)
+            SleepConditionVariableCS(self.waitCondition, mutex, INFINITE)
+            #else
+            pthread_cond_wait(self.waitCondition, mutex)
```
weissi, line 122: This doesn't look right. `pthread_cont_wait` and pretty much all other low-level condition variable implementations I have used can wake up spuriously. So usually they need to be in a loop. I.e. `while condition(&queue) {` instead of `if condition(&queue)`

jakepetroules: I don't think that alone would be correct in this particular case. If I change this to `while`, it'll hang during shutdown(), because shutdown() clears the queue (making it empty) and then wakes the condition variable. It must proceed in that case rather than go back to sleep.

That said, I think you may be right that something is generally amiss because where dequeue() is called it looks like it could end up terminating the work queue early if the condition variable was woken spuriously, since returning nil can't distinguish "no work items right now" versus "entered shutdown mode".

weissi: Yeah, that doesn't seem good. Condition variables without a `while` are almost never right.

And yes, you'll need to add some sentinel to distinguish shutdown vs. empty

PR #199, opened 2025-10-18
Fix `Environment.updating` for `custom` and `rawBytes`

### LINKED ISSUE #191: Environment.updating(_:) on a .custom environment causes inheritance
**Describe the bug**
When using the instance function `Environment.updating(_:)` on any existing instance of `Environment`, Subprocess simply adds the new values to the `.inherit` environment rather than the instance itself.

**To Reproduce**

Consider this program that uses a `.custom` environment and the `env` command to print the subprocess' perceived environment:

```swift
let pristineEnv: Environment = .custom(["A": "1"])

let result1 = try await run(
    .name("env"),
    environment: pristineEnv,
    output: .string(limit: 4096)
)

// Print 1
print(result1.standardOutput ?? "")

let result2 = try await run(
    .name("env"),
    environment: pristineEnv.updating(["B": "2"]),
    output: .string(limit: 4096)
)

// Print 2
print(result2.standardOutput ?? "")
```

**Expected behavior**
- At _Print 1_ 
  - ✅ The printed environment is exactly "A=1\n"
    - This is expected as the environment is `.custom` (i.e. not `.inherit`ed from the calling process)
- At _Print 2_ 
  - ✅ The printed environment includes "B=2"
  - ❌ The printed environment does **not** include "A=1"
  - ❌ The printed environment includes the entire inherited environment (e.g. `USER`, `HOME` etc)

**Environment:**
 - macOS 15.6
 - swiftlang-6.1.0.110.21 clang-1700.0.13.3

**Additional context**
This report is based on my assumption about how the `.updating` function should work based on it's name, apologies if this behaviour is intended.

a7medev:
Implemented `Environment.updating` preserving the existing `Environment` config with its overrides.

Resolves #191.

### DIFF
--- a/Tests/SubprocessTests/IntegrationTests.swift
+++ b/Tests/SubprocessTests/IntegrationTests.swift
@@ -435,6 +435,132 @@ extension SubprocessIntegrationTests {
         )
         #endif
     }
+
+    @Test func testEnvironmentCustomUpdating() async throws {
+        #if os(Windows)
+        let pathValue = ProcessInfo.processInfo.environment["Path"] ?? ProcessInfo.processInfo.environment["PATH"] ?? ProcessInfo.processInfo.environment["path"]
+        let customPath = "C:\\Custom\\Path"
+        let setup = TestSetup(
+            executable: .name("cmd.exe"),
+            arguments: ["/c", "echo %CUSTOMPATH%"],
+            environment: .custom([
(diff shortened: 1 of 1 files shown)


jakepetroules (approved):

jakepetroules:
Please wait for @iCharlesHu's review as well.

PR #211, opened 2025-11-09
Keep order of paths fixed

### LINKED ISSUE #210: Subprocess arbitrarily selects executable when multiple exist in `PATH`
Given multiple executables with the same name in `PATH`, `Subprocess.run` selects an arbitrary one. I [stumbled upon that with Swiftly](https://github.com/swiftlang/swiftly/issues/463) picking either Xcode's `swift` or the one it installed itself.

You can argue, tools should use absolute paths to call binaries. However, relying on the order defined in `PATH` is also valid.

I'll have a PR ready in a minute. It shows the issue better than any words would be able to explaining it here. Also see https://github.com/swiftlang/swiftly/issues/463.

SimplyDanny:
Fixes #210.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Unix.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Unix.swift
@@ -271,15 +271,13 @@ extension Arguments {
 
 // MARK: -  Executable Searching
 extension Executable {
-    internal static var defaultSearchPaths: Set<String> {
-        return Set([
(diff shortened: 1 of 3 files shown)


iCharlesHu (approved):

SimplyDanny:
> @parkera, @iCharlesHu: Can you please cut a new release? This bug is super annoying.

@parkera: Friendly reminder. 🙏

Issue #213, opened 2025-11-24
`ENOTDIR` error when `PATH` includes existing non-directory paths


fpseverino:
**Describe the bug**
Subprocess fails to spawn commands when the user's `PATH` contains an entry that **exists** but is **not a directory**.
This results in the error:
`UnderlyingError(rawValue: 20)`
Error 20 corresponds to `ENOTDIR`.

Discovered by @gwynne in https://github.com/vapor/toolbox/issues/489

**To Reproduce**
1. Add any existing non-directory file to your `PATH`:
```
PATH="$PATH:$HOME/Desktop/.DS_Store"
```
2. Invoke any tool that relies on Subprocess to locate executables along `PATH`. For example (Vapor Toolbox v20.0.0):
```bash
vapor new foo
```
3. Subprocess fails with:
`Failed to spawn the new process with underlying error: UnderlyingError(rawValue: 20)`

**Expected behavior**
Subprocess should skip PATH entries that are not directories, similar to how it already ignores PATH entries that do not exist.

**Environment:**
 - macOS Tahoe 26.1
 - Swift 6.2

**Additional context**
As identified by @gwynne:
- Subprocess appears to manually enumerate the user's `PATH`.
- Nonexistent entries are correctly ignored.
- **Existing non-directory** entries trigger an `ENOTDIR` failure.
- This occurs before the actual executable search completes.

Example:

**Fails (path exists & is not a directory):**
```
PATH="$PATH:$HOME/Desktop/.DS_Store"
vapor new foo
# → ENOTDIR
```
**Works (path does not exist):**
```
PATH="$PATH:/foo"
vapor new foo
# Works normally
```

Thanks to @gwynne for confirming and demonstrating the behavior.

PR #218, opened 2026-01-21
Add ENOTDIR as an acceptable spawn error

iCharlesHu:
The posix_spawn function currently fails with ENOTDIR instead of ENOENT when a component in the executable path is a regular file rather than a directory. This issue may occur in Subprocess if the user’s PATH value includes non-directory paths. To address this, treat ENOTDIR similarly to ENOENT by proceeding to the next available executable paths.

Resolves https://github.com/swiftlang/swift-subprocess/issues/213

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -77,7 +77,8 @@ public struct PlatformOptions: Sendable {
     /// they are sent to `posix_spawn()`.
     public var preSpawnProcessConfigurator:
         (
-            @Sendable (
+            @Sendable
+            (
                 inout posix_spawnattr_t?,
                 inout posix_spawn_file_actions_t?
             ) throws -> Void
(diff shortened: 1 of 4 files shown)


jakepetroules (approved):

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -438,7 +439,7 @@ extension Configuration {
                 }
                 // Spawn error
                 if spawnError != 0 {
-                    if spawnError == ENOENT || spawnError == EACCES {
+                    if [ENOENT, EACCES, ENOTDIR].contains(spawnError) {
```
cthielen, line 442: Does it make sense to share some of this spawn logic code between Darwin and Unix?

iCharlesHu: Unfortunately the underlying spawning mechanism is different. Darwin uses `posix_spawn` directly and `Unix` uses `fork_exec`. We did try to share the `preSpawn` logic.

PR #220, opened 2026-01-28
Subprocess error overhaul

iCharlesHu:
Ensure Subprocess only throws `SubprocessError` as designed by adapting to typed throws for internal methods. Expose static values for `SubprocessError.Code` so clients can switch on code value.

This change formalizes the recommended error handling pattern:

```swift
do {
    let result = try await run(...)
} catch let subprocessError as SubprocessError {
    // Deal with errors thrown by Subprocess itself
    switch subprocessError.code {
        case .spawnFaied: ...
        ...
    }
} catch let myError as MyError {
    // Deal with custom error thrown from execution body
}
```

Raise the macOS CI to Xcode 26.0. We hit an mis-compile issue on older `Xcode 16.3`.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -474,7 +474,7 @@ extension Environment: CustomStringConvertible, CustomDebugStringConvertible {
     ///
     /// This type respects the compiled platform's case sensitivity requirements.
     public struct Key {
-        public var rawValue: String
+        public let rawValue: String
 
         package init(_ rawValue: String) {
             self.rawValue = rawValue
@@ -965,7 +965,7 @@ internal struct IOChannel: ~Copyable, @unchecked Sendable {
         #endif
     }
 
-    @_optimize(none) // rdar://170094646
+    @_optimize(none)
     internal consuming func consumeIOChannel() -> Channel {
         let result = self.channel
         // Transfer the ownership out and therefor
--- a/Sources/Subprocess/IO/AsyncIO+Dispatch.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Dispatch.swift
@@ -43,6 +43,7 @@ final class AsyncIO: Sendable {
         from dispatchIO: DispatchIO,
         upTo maxLength: Int
     ) async throws(SubprocessError) -> DispatchData? {
+        // https://github.com/swiftlang/swift/issues/87810
         return try await _castError {
             return try await withCheckedThrowingContinuation { continuation in
                 var buffer: DispatchData = .empty
@@ -85,6 +86,7 @@ final class AsyncIO: Sendable {
         _ span: borrowing RawSpan,
         to diskIO: borrowing IOChannel
     ) async throws(SubprocessError) -> Int {
+        // https://github.com/swiftlang/swift/issues/87810
         return try await _castError {
             return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Int, any Error>) in
                 span.withUnsafeBytes {
@@ -115,6 +117,7 @@ final class AsyncIO: Sendable {
         _ array: [UInt8],
         to diskIO: borrowing IOChannel
     ) async throws(SubprocessError) -> Int {
+        // https://github.com/swiftlang/swift/issues/87810
         return try await _castError {
             return try await withCheckedThrowingContinuation { continuation in
                 array.withUnsafeBytes {
--- a/Sources/Subprocess/IO/AsyncIO+Linux.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Linux.swift
@@ -305,7 +305,7 @@ final class AsyncIO: Sendable {
             )
             guard rc == 0 else {
                 throw SubprocessError.asyncIOFailed(
-                    reason: "failed to remove \(fileDescriptor.rawValue) to epoll list",
+                    reason: "failed to remove \(fileDescriptor.rawValue) from epoll list",
                     underlyingError: Errno(rawValue: errno)
                 )
             }
@@ -414,7 +414,7 @@ extension AsyncIO {
             // Reset error code to .failedToRead to match other platforms
             guard let originalError = error as? SubprocessError else {
                 throw SubprocessError.failedToReadFromProcess(
-                    withUnderlyingError: error as? SubprocessError.UnderlyingError
+                    withUnderlyingError: nil
                 )
             }
             throw SubprocessError.failedToReadFromProcess(
--- a/Sources/Subprocess/IO/Input.swift
+++ b/Sources/Subprocess/IO/Input.swift
@@ -40,7 +40,7 @@ import FoundationEssentials
 public protocol InputProtocol: Sendable, ~Copyable {
     /// Asynchronously write the input to the subprocess using the
     /// write file descriptor
-    func write(with writer: StandardInputWriter) async throws(SubprocessError)
+    func write(with writer: StandardInputWriter) async throws
 }
 
 /// A concrete input type for subprocesses that indicates the absence
@@ -64,7 +64,7 @@ public struct NoInput: InputProtocol {
 
     /// Asynchronously write the input to the subprocess that uses the
     /// write file descriptor.
-    public func write(with writer: StandardInputWriter) async throws(SubprocessError) {
+    public func write(with writer: StandardInputWriter) async throws {
         fatalError("Unexpected call to \(#function)")
     }
 
@@ -96,7 +96,7 @@ public struct FileDescriptorInput: InputProtocol {
 
     /// Asynchronously write the input to the subprocess that use the
     /// write file descriptor.
-    public func write(with writer: StandardInputWriter) async throws(SubprocessError) {
+    public func write(with writer: StandardInputWriter) async throws {
         fatalError("Unexpected call to \(#function)")
     }
 
@@ -121,7 +121,7 @@ public struct StringInput<
 
     /// Asynchronously write the input to the subprocess that use the
     /// write file descriptor.
-    public func write(with writer: StandardInputWriter) async throws(SubprocessError) {
+    public func write(with writer: StandardInputWriter) async throws {
         guard let array = self.string.byteArray(using: Encoding.self) else {
             return
         }
@@ -140,7 +140,7 @@ public struct ArrayInput: InputProtocol {
 
     /// Asynchronously write the input to the subprocess using the
     /// write file descriptor
-    public func write(with writer: StandardInputWriter) async throws(SubprocessError) {
+    public func write(with writer: StandardInputWriter) async throws {
         _ = try await writer.write(self.array)
     }
 
@@ -154,7 +154,7 @@ public struct ArrayInput: InputProtocol {
 internal struct CustomWriteInput: InputProtocol {
     /// Asynchronously write the input to the subprocess using the
     /// write file descriptor.
-    public func write(with writer: StandardInputWriter) async throws(SubprocessError) {
+    public func write(with writer: StandardInputWriter) async throws {
         fatalError("Unexpected call to \(#function)")
     }
 
@@ -246,6 +246,8 @@ public final actor StandardInputWriter: Sendable {
 
     /// Write an array of 8-bit unsigned integers to the standard input of the subprocess.
     /// - Parameter array: The sequence of bytes to write.
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError` for more details.
     /// - Returns: the number of bytes written.
     public func write(
         _ array: [UInt8]
@@ -257,6 +259,8 @@ public final actor StandardInputWriter: Sendable {
     /// Write a raw span to the standard input of the subprocess.
     ///
     /// - Parameter `span`: The span to write.
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError` for more details.
     /// - Returns: the number of bytes written.
     public func write(_ span: borrowing RawSpan) async throws(SubprocessError) -> Int {
         return try await AsyncIO.shared.write(span, to: self.diskIO)
@@ -267,6 +271,8 @@ public final actor StandardInputWriter: Sendable {
     /// - Parameters:
     ///   - string: The string to write.
     ///   - encoding: The encoding to use when converting string to bytes
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError` for more details.
     /// - Returns: number of bytes written.
     public func write<Encoding: Unicode.Encoding>(
         _ string: some StringProtocol,
@@ -279,6 +285,8 @@ public final actor StandardInputWriter: Sendable {
     }
 
     /// Signal all writes are finished
+    /// - Throws: `SubprocessError` with error code `.asyncIOFailed`.
+    ///     See `.underlyingError` for more detail.
     public func finish() async throws(SubprocessError) {
         try self.diskIO.safelyClose()
     }
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -34,11 +34,11 @@ public protocol OutputProtocol: Sendable, ~Copyable {
 
     #if SubprocessSpan
     /// Convert the output from span to expected output type
-    func output(from span: RawSpan) throws(SubprocessError) -> OutputType
+    func output(from span: RawSpan) throws -> OutputType
     #endif
 
     /// Convert the output from buffer to expected output type
-    func output(from buffer: some Sequence<UInt8>) throws(SubprocessError) -> OutputType
+    func output(from buffer: some Sequence<UInt8>) throws -> OutputType
 
     /// The max amount of data to collect for this output.
     var maxSize: Int { get }
@@ -123,7 +123,7 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOut
 
     #if SubprocessSpan
     /// Create a string from a raw span.
-    public func output(from span: RawSpan) throws(SubprocessError) -> String? {
+    public func output(from span: RawSpan) throws -> String? {
         // FIXME: Span to String
         var array: [UInt8] = []
         for index in 0..<span.byteCount {
@@ -134,7 +134,7 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOut
     #endif
 
     /// Create a String from a sequence of 8-bit unsigned integers.
-    public func output(from buffer: some Sequence<UInt8>) throws(SubprocessError) -> String? {
+    public func output(from buffer: some Sequence<UInt8>) throws -> String? {
         // FIXME: Span to String
         let array = Array(buffer)
         return String(decodingBytes: array, as: Encoding.self)
@@ -188,13 +188,13 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
     #if SubprocessSpan
     /// Create an Array from `RawSpawn`.
     /// Not implemented
-    public func output(from span: RawSpan) throws(SubprocessError) -> [UInt8] {
+    public func output(from span: RawSpan) throws -> [UInt8] {
         fatalError("Not implemented")
     }
     #endif
     /// Create an Array from `Sequence<UInt8>`.
     /// Not implemented
-    public func output(from buffer: some Sequence<UInt8>) throws(SubprocessError) -> [UInt8] {
+    public func output(from buffer: some Sequence<UInt8>) throws -> [UInt8] {
         fatalError("Not implemented")
     }
 
@@ -338,7 +338,7 @@ extension ErrorOutputProtocol where Self == CombinedErrorOutput {
 #if SubprocessSpan
 extension OutputProtocol {
     /// Create an Array from `Sequence<UInt8>`.
-    public func output(from buffer: some Sequence<UInt8>) throws(SubprocessError) -> OutputType {
+    public func output(from buffer: some Sequence<UInt8>) throws -> OutputType {
         guard let rawBytes: UnsafeRawBufferPointer = buffer as? UnsafeRawBufferPointer else {
             fatalError("Unexpected input type passed: \(type(of: buffer))")
         }
@@ -365,7 +365,7 @@ extension OutputProtocol {
     @_disfavoredOverload
     internal func captureOutput(
         from diskIO: consuming IOChannel?
-    ) async throws(SubprocessError) -> OutputType {
+    ) async throws -> OutputType {
         if OutputType.self == Void.self {
             try diskIO?.safelyClose()
             return () as! OutputType
@@ -404,7 +404,7 @@ extension OutputProtocol {
 
         try diskIO.safelyClose()
         if let result, result.count > self.maxSize {
-            throw .outputLimitExceeded(limit: self.maxSize)
+            throw SubprocessError.outputLimitExceeded(limit: self.maxSize)
         }
 
         #if SUBPROCESS_ASYNCIO_DISPATCH
@@ -416,16 +416,16 @@ extension OutputProtocol {
 }
 
 extension OutputProtocol where OutputType == Void {
-    internal func captureOutput(from fileDescriptor: consuming IOChannel?) async throws(SubprocessError) {}
+    internal func captureOutput(from fileDescriptor: consuming IOChannel?) async throws {}
 
     #if SubprocessSpan
     /// Convert the output from raw span to expected output type
-    public func output(from span: RawSpan) throws(SubprocessError) {
+    public func output(from span: RawSpan) throws {
         fatalError("Unexpected call to \(#function)")
     }
     #endif
     /// Convert the output from a sequence of 8-bit unsigned integers to expected output type.
-    public func output(from buffer: some Sequence<UInt8>) throws(SubprocessError) {
+    public func output(from buffer: some Sequence<UInt8>) throws {
         fatalError("Unexpected call to \(#function)")
     }
 }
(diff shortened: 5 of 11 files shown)


Sources/Subprocess/Buffer.swift
```diff
@@ -71,10 +71,14 @@ extension AsyncBufferSequence.Buffer {
     ///   the method creates it. The argument is valid only for the duration of the
     ///   closure's execution.
     /// - Returns: The return value of the body closure.
-    public func withUnsafeBytes<ResultType>(
-        _ body: (UnsafeRawBufferPointer) throws -> ResultType
-    ) rethrows -> ResultType {
-        return try self.data.withUnsafeBytes(body)
+    public func withUnsafeBytes<ResultType, Error: Swift.Error>(
+        _ body: (UnsafeRawBufferPointer) throws(Error) -> ResultType
+    ) throws(Error) -> ResultType {
```
itingliu, line 76: Just so I'm not misreading this PR: This is exactly the use case that this PR aims to address right? 

And this has nothing to do with embedded since `Subprocess.Error` now contains an existential, so it won't on embedded, right? I'm not saying we should update `Subprocess` to support embedded platforms. I just want to make sure I'm not reading too much into this change

itingliu (commented):
I did not read every line of change but overall LGMT. Just got one question to clarify the intention of this PR

Sources/Subprocess/Buffer.swift
iCharlesHu: > This is exactly the use case that this PR aims to address right?

YES! Unfortunately, I don't think Subprocess would be applicable on embedded platforms because it requires locking, async IO, etc. However, it's still important to adopt typed throws independent of embedded work since Subprocess is a new API and it was designed to throw one error type (`SubprocessError`).

rnapier: I don't understand why wrapping the underlying Error in another type is helping the caller. Rather than checking the error directly, now the caller has to first unload the SubprocessError, and then check an `any Error` (which can be anything). What is the benefit here over an untyped throws?

iCharlesHu: @rnapier  We want to use typed throws for several reasons:

1. Swift is generally moving towards typed throws, particularly for non-public code. The majority of this PR focuses on updating instances that currently do not use typed throws.

2. From an API perspective, I believe it is more intuitive to have a specific error code in this manner. Subprocess was designed to only throw `SubprocessError`. The only exception to this rule is when the custom execution body could potentially throw any error. In my opinion, it is clearer to have a specific error code that indicates the body threw an error so you should check “underlying error” rather than inferring it from the type and having to cast.

> Rather than checking the error directly, 

3. How do you "check" the error? Since the `Error` protocol doesn't have much info you'll most likely have to cast the error back to a concrete type anyway:
```
do { ... }
catch {
    if let subprocessError = error as? SubprocessError {
       ...
    }
    if let myError = error as? MyError {
        ...
    }
    // What to do here if the error isn't what you expect?
}
```
Typed throws allows you to get rid of that first cast, since now it'll always be `SubprocessError`. You only need to cast the `underlyingError` to your type if you choose to.

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
```diff
@@ -81,24 +83,26 @@ final class AsyncIO: Sendable {
...
-                let dispatchData = DispatchData(
-                    bytesNoCopy: $0,
-                    deallocator: .custom(
-                        nil,
-                        {
-                            // noop
-                        }
+    ) async throws(SubprocessError) -> Int {
+        return try await _castError {
+            return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Int, any Error>) in
```
parkera, line 91: Should the second argument be `SubprocessError`? Why do we need the cast?

Sources/Subprocess/Platforms/Subprocess+BSD.swift
```diff
@@ -35,28 +41,33 @@ internal func monitorProcessTermination(
...
-        let source = DispatchSource.makeProcessSource(
-            identifier: processIdentifier.value,
-            eventMask: [.exit],
-            queue: .global()
-        )
-        source.setEventHandler {
-            source.cancel()
-            continuation.resume(
-                with: Result(catching: { () throws(SubprocessError.UnderlyingError) -> TerminationStatus in
+    return try await _castError {
```
parkera, line 41: Same question here about what the `_castError` is for - I won't keep repeating it after this though.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -415,7 +429,7 @@ extension Configuration {
                     gidPtr: gidPtr
                 )
                 let (spawnError, pid) = try await runOnBackgroundThread {
-                    return possibleExecutablePath.withCString { exePath in
+                    return possibleExecutablePath._withCString { exePath in
```
parkera, line 422: What's this change about?

Sources/Subprocess/Configuration.swift
```diff
@@ -1103,47 +1120,52 @@ internal struct CreatedPipe: ~Copyable {
...
+            throw SubprocessError(code: errorCode, underlyingError: error)
+        }
         #endif
     }
 }
 
 extension Optional where Wrapped: Collection {
     func withOptionalUnsafeBufferPointer<Result>(
-        _ body: ((UnsafeBufferPointer<Wrapped.Element>)?) throws -> Result
-    ) rethrows -> Result {
```
parkera, line 1122: Why not leave this as rethrowing but with a typed error? We just didn't need it?

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -194,14 +194,17 @@ extension SubprocessAsyncIOTests {
...
-            #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
+            #expect(
+                subprocessError.underlyingError as? Errno == Errno(rawValue: ECANCELED)
+            )
             #else
             // On Linux, depending on timing, either epoll_ctl or read
             // could throw error first
             #expect(
-                subprocessError.underlyingError == .init(rawValue: EBADF) || subprocessError.underlyingError == .init(rawValue: EINVAL) || subprocessError.underlyingError == .init(rawValue: EPERM)
+                subprocessError.underlyingError as? Errno == Errno(rawValue: EBADF) || subprocessError.underlyingError as? Errno == Errno(rawValue: EINVAL) || subprocessError.underlyingError as? Errno == Errno(rawValue: EPERM)
```
parkera: The fact that we need to cast the underlyingError seems a bit counterintuitive with the goal of using typed throws, right?

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
iCharlesHu: Unfortunately `any Error` is defined in the method signature:
```
@inlinable public func withCheckedThrowingContinuation<T>(isolation: isolated (any Actor)? = #isolation, function: String = #function, _ body: (CheckedContinuation<T, any Error>) -> Void) async throws -> sending T
```
`withCheckedThrowingContinuation` will not take a closure that takes (CheckedContinuation<T, SubprocessError>).

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
iCharlesHu: This is my attempt to write my own typed throw version of `withCString`. If not then every level of `withCString` will need to be wrapped like
```
return try _castError {
    return path.withCString {
        return try _castError {
           return args.withCString { .... }
        }
    }
}
```

Sources/Subprocess/Configuration.swift
iCharlesHu: Yes I discovered this method doesn't need to throw.

Tests/SubprocessTests/AsyncIOTests.swift
iCharlesHu: Somewhat but you only need to cast that error if you need to. Like the comment above, in untyped throws you always have to cast the error to something in order to use it:

```swift
do { ... }
catch {
    if let subprocessError = error as? SubprocessError {
       ...
    }
    if let myError = error as? MyError {
        ...
    }
    // What to do here if the error isn't what you expect?
}
```

With this setup you only need to cast the underlying error if you need to:

```swift
do { ... }
catch {
    guard error.code.value != 15 else {
        // For other errors just display the error message:
        print(error.description)
    }
    // Since error code 15 means body threw an error, lets unwrap underlying error to find out why
    if let underlying = error.underlyingError as? Errno { ... }
}
```

Sources/Subprocess/Error.swift
```diff
@@ -21,12 +21,39 @@ import Musl
...
 
+#if canImport(System)
+@preconcurrency import System
+#else
+@preconcurrency import SystemPackage
+#endif
+
 /// Error thrown from Subprocess
-public struct SubprocessError: Swift.Error, Hashable, Sendable {
+public struct SubprocessError: Swift.Error, Sendable, Equatable {
```
parkera:
```suggestion
public struct SubprocessError<UnderlyingError: Swift.Error>: Swift.Error, Sendable, Equatable {
```

and 
```
extension SubprocessError : Equatable where UnderlyingError: Equatable { }
```

Sources/Subprocess/Buffer.swift
rnapier: > Swift is generally moving towards typed throws, particularly for non-public code

I'm not sure what you mean by "non-public" here. This is `public`, right? But the [guidance](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0413-typed-throws.md#motivation) on typed throws is the opposite:

>Even with the introduction of typed throws into Swift, the existing (untyped) throws remains the better default error-handling mechanism for most Swift code. 

This interface doesn't align with any of the criteria described in [When to use typed throws](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0413-typed-throws.md#when-to-use-typed-throws).

I'll buy that `SubprocessError` is doing some work here and could even be ok, but your example isn't how the `catch` should be written, and the way `Code` is implemented exactly recreates the problem you're describing. The normal way to handle the above catch would be:

```
do { ... }
catch let e as SubprocessError { ... }
catch let e as MyError { ... }
// optionally, but not required if this context throws:
catch { /* unexpected error */ }
```

But because `Code.Storage` is hidden, `/* unexpected error*/` still happens in your interface. I can't write a `switch` over all `SubprocessError` values. So I always need a `default` leg for unexpected values. That seems counter to SE-0413. When the possible errors might expand in the future, SE-0413 says that means we should use untyped throws, not hide the enum.

Given how your API is smoothing over cross-platform issues where the same "logical" error may be different types on different platforms, and given Swift's lack of a rich set of core Error types, SubprocessError may still make good sense, but I'm concerned that the reasoning for it doesn't align with the design guidelines. I'm seeing a lot of junior devs reaching for typed throws when they should not and it makes a mess of their code. I want to make sure they have good examples to follow from core frameworks, or at least good explanations of why they shouldn't follow the patterns of a core framework.

rnapier: Thinking about this more, it's worth seeing what typed throws does to the closure error. I've built a little program that runs `find /` and it throws an error if it discovers a directory with `/usr` in it before it finds `/Applications`.

```
struct SubdirectoryError: Error {
    let path: String
}

// If we find /usr, throw. If we find /Applications, return it. If nothing, return nil.
// Also may throw if the Subprocess fails.
func findApplications() async throws(SubprocessError) -> String? {
    try await Subprocess.run(.name("find"), arguments: ["/"]) { _, output in
        for try await line in output.lines() {
            if line.contains("/usr") {
                throw SubdirectoryError(path: line)
            } else if line.contains("/Applications") {
                return line
            }
        }
        return nil
    }.value
}
```

With this PR, that looks like:

         do {
            if let apps = try await findApplications() {
                print("Found it: \(apps)")
            } else {
                print("Not found")
            }
        } catch {
            if case .executionError(let underlying) = error.underlyingError,
               let subdirError = underlying as? SubdirectoryError {
                print("Found forbidden usr: \(subdirError.path)")
            } else {
                throw error
            }
        }

Without typed throws, the `catch` block is just:

        catch let error as SubdirectoryError {
            print("Found forbidden usr: \(error.path)")
        }

In my experience, "just rethrow unexpected errors; we'll deal with them at the top" is a normal situation, and I think typed throws makes it pretty complicated to get down to the errors you're really going to do something about. If there's a way to unload the underlying error as my type directly in the `catch`, I'd probably feel differently, but I don't think Swift supports this. The fact that `underlyingError` isn't the actual error, but yet another wrapper around the actual error makes it even more complicated to get to it.

I haven't tested it out in code yet, but another common case here is CancellationError, which I think is not current thrown (but IMO should be) when the enclosing Task is cancelled. I'm exploring that for task timeouts.

In my experience, any time there's an "underlying error" that the caller will actually work with (as opposed to just log), wrapping it up just forces a lot of boilerplate unwrapping. Worse, if the calling code *also* throws typed errors, then you get a lot of unwrapping/rewrapping, just to in the end switch on the underlying error that you could have thrown in the first place. [edited to remove off-topic discussion]

itingliu: Saying one pattern is definitely better than the other is probably never accurate. But for this particular case, this public function simply rethrows the error that the `body` throws, so it makes sense to constrain that they are the same error. In fact there are two recent proposals for generalizing existing `Data` and `DataProtocol`'s `throw`-ing API: 
https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0034-ContiguousBytes-Span.md
https://github.com/swiftlang/swift-foundation/pull/1649/changes
For this one I think it makes sense to follow the pattern there.

The other `public func run(...)` function is however a different case so we should probably discuss it separately. Thanks for your example @rnapier. (Side note, I'd like to point out that your last paragraph is not super constructive. I'd appreciate if you don't assign identity or group stereotype to a change.)

Sources/Subprocess/IO/Output.swift
```diff
@@ -446,10 +440,12 @@ extension OutputProtocol {
             return try self.output(from: span)
         }
 
-        return try data.withUnsafeBytes { ptr in
-            let bufferPtr = UnsafeRawBufferPointer(start: ptr, count: data.count)
-            let span = RawSpan(_unsafeBytes: bufferPtr)
-            return try self.output(from: span)
+        return try _castError {
```
itingliu: Do we still need this? It seems to build fine when I just return the body as-is

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
itingliu: It sounds like this function should be updated to throw a typed error

```
@inlinable public func withCheckedThrowingContinuation<T, E: Error>(isolation: isolated (any Actor)? = #isolation, function: String = #function, _ body: (CheckedContinuation<T, E>) -> Void) async throws(E) -> sending T
```
Like this? If so, I think we can follow up with @clintonpi, and should we stick an issue to remove `_castError` when this work is done in stdlib?

Sources/Subprocess/IO/Output.swift
```diff
@@ -460,9 +456,11 @@ extension OutputProtocol {
             return try self.output(from: span)
         }
 
-        return try data.withUnsafeBufferPointer { ptr in
-            let span = RawSpan(_unsafeBytes: UnsafeRawBufferPointer(ptr))
-            return try self.output(from: span)
+        return try _castError {
```
itingliu: We probably don't need this `_castError` now that `Data` API throws a typed error

Sources/Subprocess/IO/Input.swift
```diff
@@ -279,7 +279,7 @@ public final actor StandardInputWriter: Sendable {
     }
 
     /// Signal all writes are finished
-    public func finish() async throws {
+    public func finish() async throws(SubprocessError) {
```
itingliu, line 290: Is it intentional that we're throwing this in a public func? If so would it be useful to mention in the header doc that the caller should expect `.asyncIOFailed` error if it fails?

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -99,12 +99,12 @@ extension Execution {
     public func send(
         signal: Signal,
         toProcessGroup shouldSendToProcessGroup: Bool = false
-    ) throws {
-        func _kill(_ pid: pid_t, signal: Signal) throws {
+    ) throws(SubprocessError) {
```
itingliu, line 104: Same question about whether this throwing a concrete error is intentional now and whether it always throws `SubprocessError.processControlFailed`

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
clintonpi: I've taken note of `withCheckedThrowingContinuation`.

Sources/Subprocess/IO/Output.swift
iCharlesHu: Ahh good catch thanks!!

Sources/Subprocess/IO/Input.swift
iCharlesHu: Yes, since this method only throws `SubprocessError` since no execution body error is involved. Sounds good I'll update the header doc.

iCharlesHu: I just realized the the other methods in StandardInputWriter are not marked with `throws(SubprocessError)`. They all should since no user error is involved here. I'll update

Sources/Subprocess/Platforms/Subprocess+Unix.swift
iCharlesHu: Yes it's intentional. I'll update the documentation.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
itingliu: Why do we need another `capturedErrno` here? Why don't we use `errno` directly?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -319,12 +304,9 @@ final class AsyncIO: Sendable {
...
             )
             guard rc == 0 else {
-                throw SubprocessError(
-                    code: .init(
-                        .asyncIOFailed(
-                            "failed to remove \(fileDescriptor.rawValue) to epoll list")
-                    ),
-                    underlyingError: .init(rawValue: errno)
+                throw SubprocessError.asyncIOFailed(
+                    reason: "failed to remove \(fileDescriptor.rawValue) to epoll list",
```
itingliu: Nit: should this say "~**to**~ **from** epoll list"?

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -292,12 +280,9 @@ final class AsyncIO: Sendable {
                     }
 
                     let capturedError = errno
```
itingliu, line 282: same question about using a separate `capturedError`

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -432,11 +413,12 @@ extension AsyncIO {
         } catch {
             // Reset error code to .failedToRead to match other platforms
             guard let originalError = error as? SubprocessError else {
-                throw error
+                throw SubprocessError.failedToReadFromProcess(
+                    withUnderlyingError: error as? SubprocessError.UnderlyingError
```
itingliu: Would this `as?` ever fail?

Sources/Subprocess/IO/Input.swift
```diff
@@ -246,19 +246,23 @@ public final actor StandardInputWriter: Sendable {
 
     /// Write an array of 8-bit unsigned integers to the standard input of the subprocess.
     /// - Parameter array: The sequence of bytes to write.
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError`for more details.
```
itingliu: Nit:
```suggestion
    ///     See `.underlyingError` for more details.
```

Sources/Subprocess/IO/Input.swift
```diff
@@ -246,19 +246,23 @@ public final actor StandardInputWriter: Sendable {
...
+    ) async throws(SubprocessError) -> Int {
         return try await AsyncIO.shared.write(array, to: self.diskIO)
     }
 
     #if SubprocessSpan
     /// Write a raw span to the standard input of the subprocess.
     ///
     /// - Parameter `span`: The span to write.
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError`for more details.
```
itingliu:
```suggestion
    ///     See `.underlyingError` for more details.
```

Sources/Subprocess/IO/Input.swift
```diff
@@ -267,19 +271,23 @@ public final actor StandardInputWriter: Sendable {
     /// - Parameters:
     ///   - string: The string to write.
     ///   - encoding: The encoding to use when converting string to bytes
+    /// - Throws: `SubprocessError` with error code `.failedToWriteToSubprocess`.
+    ///     See `.underlyingError`for more details.
```
itingliu:
```suggestion
    ///     See `.underlyingError` for more details.
```

Sources/Subprocess/Thread.swift
```diff
@@ -39,17 +39,25 @@ private typealias ConditionType = pthread_cond_t
 private typealias ThreadType = pthread_t
 #endif
 
-internal func runOnBackgroundThread<Result>(
-    _ body: @Sendable @escaping () throws -> Result
-) async throws -> Result {
+#if canImport(System)
+@preconcurrency import System
```
itingliu, line 43: I thought we don't need `preconcurrency import` anymore?

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -157,14 +157,14 @@ extension SubprocessAsyncIOTests {
             }
 
             #if os(Windows)
-            #expect(subprocessError.underlyingError == .init(rawValue: DWORD(ERROR_INVALID_HANDLE)))
+            #expect(subprocessError.underlyingError as! SubprocessError.WindowsError == SubprocessError.WindowsError(rawValue: DWORD(ERROR_INVALID_HANDLE)))
```
itingliu: Do we really need to force cast it to compare the two errors?

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -157,14 +157,14 @@ extension SubprocessAsyncIOTests {
...
             #elseif canImport(Darwin) || os(FreeBSD) || os(OpenBSD)
-            #expect(subprocessError.underlyingError == .init(rawValue: ECANCELED))
+            #expect(subprocessError.underlyingError as? Errno == Errno(rawValue: ECANCELED))
```
itingliu: I'm seeing a warning `Conditional downcast from 'SubprocessError.UnderlyingError?' (aka 'Optional<Errno>') to 'Errno' does nothing`

I think you can just write `#expect(subprocessError.underlyingError == Errno(rawValue: ECANCELED))`

That'd also help clarifying Tony's question below

Tests/SubprocessTests/ProcessMonitoringTests.swift
```diff
@@ -199,31 +199,25 @@ extension SubprocessProcessMonitoringTests {
...
+        let expectedError = Errno(rawValue: ECHILD)
         let processIdentifier = ProcessIdentifier(value: .max)
         #endif
-        await #expect(throws: expectedError) {
+        let caughtError = await #expect(throws: SubprocessError.self) {
             _ = try await monitorProcessTermination(for: processIdentifier)
         }
+        #expect(caughtError?.code == .failedToMonitorProcess)
+        let underlying = try #require(caughtError?.underlyingError)
+        #expect(expectedError.equals(to: underlying))
```
itingliu: Why do we need this wrapper? Simply doing `#expect(expectedError == underlying)` seems to work fine for me?

Or even simpler, can we just write? 
```
#expect(caughtError == .failedToMonitor(withUnderlyingError: expectedError))
```

Sources/Subprocess/Platforms/Subprocess+Unix.swift
iCharlesHu: Ahh good catch. We do have good reason in other places to actually capture the error code before some other method might change it, but I don't think it's necessary here.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: These are cases where we actually do want to do this. Since `errno` is a global mutable variable, later calls such as `removeRegistration()` might hit new errors and reset `errno`. We didn't have to do this on the first case you mentioned because there wasn't much code surrounding it. `AsyncIO` is different.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: Ahh good catch.. it would actually never succeed since the inner function only throws `SubprocessError`.

Sources/Subprocess/Thread.swift
iCharlesHu: Yes we don't! I'm removing them in my next PR to get rid of all warnings!

Tests/SubprocessTests/AsyncIOTests.swift
iCharlesHu: Good catch! I forgot to update this test now the underlying error is no longer `any Error`.

Tests/SubprocessTests/AsyncIOTests.swift
iCharlesHu: Yessss I updated all references.

Tests/SubprocessTests/ProcessMonitoringTests.swift
iCharlesHu: Another test that I forgot to update 😓 `SubprocessError` used to be not `Equatable` (which ironically sparked this whole change)

Tests/SubprocessTests/TestSupport.swift
```diff
@@ -72,3 +78,23 @@ extension Trait where Self == ConditionTrait {
         )
     }
 }
+
+#if os(Windows)
```
itingliu: Do we still need these `equals(to:)`?

iCharlesHu: No I removed it. Thanks!!

PR #226, opened 2026-03-17
Add ResourceUsage to report subprocess CPU time and memory consumption

jakepetroules:
Introduce a public ResourceUsage struct that exposes userTime, systemTime (as Duration), and maxRSS (in bytes) for every terminated subprocess. An ExecutionResult protocol provides common access to terminationStatus and resourceUsage across both ExecutionOutcome and ExecutionRecord.

On Unix, resource data is collected via wait4 (BSD) or the Linux kernel's 5-argument waitid syscall, which populates a rusage struct alongside the termination status. A linux_waitid C shim is added because glibc and musl only expose the 4-parameter POSIX waitid that omits rusage. The raw rusage is available as a public property on non-Windows platforms, with maxRSS normalized from KiB to bytes on Linux, FreeBSD, and OpenBSD.

On Windows, GetProcessTimes provides CPU time (converted from FILETIME 100ns units) and GetProcessMemoryInfo provides PeakWorkingSetSize (maxRSS).

This functionality is particularly necessary as part of SwiftSubprocess because collecting rusage information from a terminated subprocess requires the ability to run code when a process has changed state from executing to zombie, but before its pid has been reaped - something not possible with the current Subprocess API. Further, it is notoriously difficult to collect this information across all OSes for an arbitrary PID anyways, at least in a way that doesn't also simultaneously reap the pid. User time, system time, and maxRSS are some of the most common metrics typically extracted from getrusage. Exposing the raw rusage struct provides access to the rest, and on Windows, callers can use DuplicateHandle to get a process descriptor that can outlive Subprocess's control of the process and collect any additional metrics from the process when it is known to be in a terminated state.

### DIFF
--- a/Sources/Subprocess/Result.swift
+++ b/Sources/Subprocess/Result.swift
@@ -15,6 +15,109 @@ import System
 import SystemPackage
 #endif
 
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Android)
+import Android
+#elseif canImport(WinSDK)
+import WinSDK
+#endif
+
+// MARK: - ResourceUsage
+
+/// Resource usage information for a terminated subprocess.
+public struct ResourceUsage: Sendable, Hashable {
+    /// The total amount of time spent executing in user mode.
+    public let userTime: Duration
+    /// The total amount of time spent executing in kernel mode.
+    public let systemTime: Duration
+    /// The peak resident set size (maximum memory used), in bytes.
+    public let maxRSS: Int
+
+    #if !os(Windows)
+    /// The underlying POSIX resource usage information.
+    public let rusage: rusage
+    #endif
+}
+
+extension ResourceUsage {
+    #if os(Windows)
+    internal init(processHandle: HANDLE) {
+        var creationTime = FILETIME()
+        var exitTime = FILETIME()
+        var kernelTime = FILETIME()
+        var userFileTime = FILETIME()
+
+        if GetProcessTimes(
+            processHandle,
+            &creationTime,
+            &exitTime,
+            &kernelTime,
+            &userFileTime
+        ) {
+            self.userTime = Self.duration(from: userFileTime)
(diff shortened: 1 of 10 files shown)


Sources/Subprocess/Result.swift
```diff
@@ -15,6 +15,109 @@ import System
...
+
+    private static func duration(from ft: FILETIME) -> Duration {
+        let hundredNanos = UInt64(ft.dwHighDateTime) << 32 | UInt64(ft.dwLowDateTime)
+        let seconds = Int64(hundredNanos / 10_000_000)
+        let remainder = Int64(hundredNanos % 10_000_000)
+        return Duration(
+            secondsComponent: seconds,
+            attosecondsComponent: remainder * 100_000_000_000
+        )
+    }
```
compnerd: This feels like something that might be better to extract as:

```swift
extension Duration {
  internal init(_ ft: FILETIME) {
    let nanos = UInt64(ft.dwHighDateTime) << 32 | UInt64(ft.dwLowDateTime)
    let seconds = Int64(nanos / 10_000_000)
    let remainder = Int64(nanos % 10_000_000)
    self = Duration(secondsComponents: seconds, attosecondsComponent: remainder * 100_000_000_000)
  }
}
```

Sources/Subprocess/Result.swift
```diff
@@ -15,6 +15,109 @@ import System
...
+            secondsComponent: Int64(usage.ru_utime.tv_sec),
+            attosecondsComponent: Int64(usage.ru_utime.tv_usec) * 1_000_000_000_000
+        )
+        self.systemTime = Duration(
+            secondsComponent: Int64(usage.ru_stime.tv_sec),
+            attosecondsComponent: Int64(usage.ru_stime.tv_usec) * 1_000_000_000_000
+        )
+        #if canImport(Darwin)
+        self.maxRSS = Int(usage.ru_maxrss) // bytes on Darwin
+        #else
```
jakepetroules: note to self: probably should explicitly check for Linux/BSD here and #error in other cases to avoid assumptions on new platform bringup and avoid cases where the unit might be incorrect until someone notices and then people become dependent on the behavior

PR #228, opened 2026-03-20
Add StreamingBehavior to AsyncBufferSequence for throughput/latency control

dannys42:
## Motivation

When streaming subprocess output, users previously had no control over the trade-off between throughput and latency. The sequence always batched data for maximum throughput, which limits use for interactive cases where users want to receive output as soon as it's available.

## Changes

- Add `AsyncBufferSequence.StreamingBehavior` enum with three modes:
    - `.throughput` (default): batch data for best performance (preserves the existing behavior)
    - `.balanced`: batch data with guaranteed 250ms max delivery interval
    - `.latency`: deliver data as soon as it's available (also bounded by 250ms)
- Add `streamingBehavior` parameter to all `run` overloads that expose an `AsyncBufferSequence`, defaulting to `.throughput` for backwards compatibility
- Add unit tests verifying that data is delivered progressively under `.latency` and `.balanced` behaviors

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -166,6 +166,8 @@ public func run<
 ///     as the default buffer size. Larger buffer sizes may improve performance for
 ///     subprocesses that produce large amounts of output, while smaller buffer sizes
 ///     may reduce memory usage and improve responsiveness for interactive applications.
+///   - streamingBehavior: Controls the trade-off between throughput and latency when
+///     streaming output from the subprocess. Defaults to `.throughput`.
 ///   - isolation: the isolation context to run the body closure.
 ///   - body: The custom execution body to manually control the running process.
 /// - Returns: an `ExecutableResult` type containing the return value of the closure.
@@ -178,6 +180,7 @@ public func run<Result, Input: InputProtocol, Error: ErrorOutputProtocol>(
     input: Input = .none,
     error: Error = .discarded,
     preferredBufferSize: Int? = nil,
+    streamingBehavior: AsyncBufferSequence.StreamingBehavior = .throughput,
     isolation: isolated (any Actor)? = #isolation,
     body: ((Execution, AsyncBufferSequence) async throws -> Result)
 ) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void {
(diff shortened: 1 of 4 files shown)


jakepetroules:
Since #259 was merged to rewrite the I/O backends and resolve the correctness issues with buffering, I think this change is likely no longer relevant - I don't think we have a realistic throughput/latency tradeoff to make.

@dannys42 given the current state of the repository do you think you still have an enhancement to propose here, or could this be closed?

dannys42:
Oh cool.  Let me try out the new version and see if it resolves this issue.

dannys42:
@jakepetroules Trying 0.4, I still end up having to put `preferredBufferSize=1` in order to get true streaming behavior, which is an unfortunate performance hit.  Without `preferredBufferSize=1`, the program will not output anything until either the program terminates or there's a page (16kB) worth of data in the stream.


Here's a simple test program:

```swift
import Foundation
import Subprocess

@main
struct ExampleApp {

    static func main() async throws {

        let target = CommandLine.arguments.count > 1
            ? CommandLine.arguments[1]
            : "./test_timing"

        let stdout = FileHandle.standardOutput

            .path(.init(target)),
            preferredBufferSize: 1 // <--- comment out this line to see the problem
        ) { _, outputSequence in
            for try await line in outputSequence.lines() {
                stdout.write(Data(line.utf8))
            }
        }

        print("exit:", result.terminationStatus)
    }
}
```

Also compile this as `test_timing`:
```c
#include <stdio.h>
#include <unistd.h>
#include <time.h>

void print_time(const char* label) {
    struct timespec ts;
    clock_gettime(CLOCK_MONOTONIC, &ts);
    fprintf(stderr, "[%s] %ld.%09ld\n", label, ts.tv_sec, ts.tv_nsec);
    fflush(stderr);
}

int main(void) {
    print_time("start");
    
    fprintf(stdout, "line 1\n");
    fflush(stdout);
    print_time("after line 1");
    
    sleep(1);
    
    fprintf(stdout, "line 2\n");
    fflush(stdout);
    print_time("after line 2");
    
    sleep(2);
    
    fprintf(stdout, "line 3\n");
    fflush(stdout);
    print_time("after line 3");
    
    return 0;
}
```

[Example.zip](https://github.com/user-attachments/files/27812827/Example.zip)

dannys42:
Oh wait... I just realized the fix you mentioned is in `main` and not yet released.

dannys42:
Yes, the fix seems to resolve the issue for me.  Thank you!

PR #232, opened 2026-04-03
Rename AsyncBufferSequence.LineSequence to StringSequence and add custom Unicode scalar separator support

### LINKED ISSUE #212: [Feature] null-delimited iteration
**Is your feature request related to a problem? Please describe.**
Currently `AsyncBufferSequence.LineSequence.AsyncIterator.next()` only splits at newlines.

Since the newline character is permitted within many strings and splitting on newlines thus results in incomplete records, most OS and programming tools support outputting null-delimited records.  Using null-delimited output is considered best practice in scripting, and required for programmatically consuming data from filesystem listing, git log records, etc.

**Describe the solution you'd like**
Support configuration to use the null character (0) as the separator, as well as some way to detect if an iterator is configured to split at the newline or null character.

**Describe alternatives you've considered**
Rejected alternatives:
 
- String or regex: Splitting input using some arbitrary String would be most flexible, but also entails much more complicated scanning.

- Naming: Currently the API is named using the default case of "lines".  While this might seems inaccurate in the case of null-delimited output, I believe it is still clear, and being able to query the delimiter as newline or null is sufficient (and would avoid changing the API).  However, if it's preferred to be more accurate, I would propose "segment" as the base concept supporting both newline- and null-delimited output.

- Duplicated iterator for null-delimited output: one could create a new iterator type with corresponding different names, to preserve the existing lines API and support null-terminated segments.  I assume the iteration itself could be refactored so there would be only one delegate scanning function.

**Additional context**

As an example, the Shwift scripting library was updated to support arbitrary strings, but the driving use-case was null-terminated records:

<https://github.com/GeorgeLyon/Shwift/pull/27>

iCharlesHu:
This PR generalizes `AsyncBufferSequence.LineSequence` to `StringSequence` that splits the buffer into Strings according to custom separator sequence in addition to line breaks. Developers could use this API to perform custom line splitting, similar to `String.split`.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/212

### DIFF
--- a/Sources/Subprocess/AsyncBufferSequence.swift
+++ b/Sources/Subprocess/AsyncBufferSequence.swift
@@ -99,35 +99,68 @@ public struct AsyncBufferSequence: AsyncSequence, @unchecked Sendable {
         )
     }
 
-    /// Creates a line sequence that iterates through this buffer sequence line by line.
-    public func lines() -> LineSequence<UTF8> {
-        return LineSequence(
+    /// Splits the buffer into strings using the specified separator.
+    ///
+    /// - Parameters:
+    ///   - separator: The delimiter to split on. The default
+    ///     value is `.lineBreaks`.
+    ///   - bufferingPolicy: The strategy for handling
+    ///     back-pressure. The default value is
+    ///     `.maxLineLength(128 * 1024)`.
+    /// - Returns: A ``StringSequence`` that iterates through
+    ///   the buffer contents as strings.
+    public func strings(
+        separatedBy separator: StringSequence<UTF8>.Separator = .lineBreaks,
+        bufferingPolicy: StringSequence<UTF8>.BufferingPolicy = .maxLineLength(128 * 1024),
+    ) -> StringSequence<UTF8> {
+        return StringSequence(
             underlying: self,
             encoding: UTF8.self,
-            bufferingPolicy: .maxLineLength(128 * 1024)
+            bufferingPolicy: bufferingPolicy,
+            separator: separator
         )
     }
 
-    /// Creates a line sequence that iterates through this buffer sequence line by line.
+    /// Splits the buffer into strings with the given encoding
+    /// and separator.
+    ///
     /// - Parameters:
-    ///   - encoding: The target encoding for strings.
-    ///   - bufferingPolicy: The strategy for handling back-pressure.
-    /// - Returns: A ``LineSequence`` that iterates through this buffer sequence line by line.
-    public func lines<Encoding: _UnicodeEncoding>(
-        encoding: Encoding.Type,
-        bufferingPolicy: LineSequence<Encoding>.BufferingPolicy = .maxLineLength(128 * 1024)
-    ) -> LineSequence<Encoding> {
-        return LineSequence(underlying: self, encoding: encoding, bufferingPolicy: bufferingPolicy)
+    ///   - separator: The delimiter to split on. The default
+    ///     value is `.lineBreaks`.
+    ///   - bufferingPolicy: The strategy for handling
+    ///     back-pressure. The default value is
+    ///     `.maxLineLength(128 * 1024)`.
+    ///   - encoding: The Unicode encoding to decode with.
+    /// - Returns: A ``StringSequence`` that iterates through
+    ///   the buffer contents as strings.
+    public func strings<Encoding: _UnicodeEncoding>(
+        separatedBy separator: StringSequence<Encoding>.Separator = .lineBreaks,
+        bufferingPolicy: StringSequence<Encoding>.BufferingPolicy = .maxLineLength(128 * 1024),
+        as encoding: Encoding.Type,
+    ) -> StringSequence<Encoding> {
+        return StringSequence(
+            underlying: self,
+            encoding: encoding,
+            bufferingPolicy: bufferingPolicy,
+            separator: separator
+        )
     }
 }
 
 @available(*, unavailable)
 extension AsyncBufferSequence.Iterator: Sendable {}
 
-// MARK: - LineSequence
+// MARK: - StringSequence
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -382,6 +461,59 @@ extension AsyncBufferSequence.LineSequence {
...
+        /// LF:    Line Feed, U+000A
+        /// VT:    Vertical Tab, U+000B
+        /// FF:    Form Feed, U+000C
+        /// CR:    Carriage Return, U+000D
+        /// CR+LF: CR (U+000D) followed by LF (U+000A)
+        /// NEL:   Next Line, U+0085
+        /// LS:    Line Separator, U+2028
+        /// PS:    Paragraph Separator, U+2029
+        /// ```
+        public static var lineBreaks: Self { .init(.lineBreaks) }
```
itingliu: And this is basically `CharacterSet.newlines`

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -99,35 +99,68 @@ public struct AsyncBufferSequence: AsyncSequence, @unchecked Sendable {
...
+    ///   - separator: The delimiter to split on. The default
+    ///     value is `.lineBreaks`.
+    ///   - bufferingPolicy: The strategy for handling
+    ///     back-pressure. The default value is
+    ///     `.maxLineLength(128 * 1024)`.
+    ///   - encoding: The Unicode encoding to decode with.
+    /// - Returns: A ``StringSequence`` that iterates through
+    ///   the buffer contents as strings.
+    public func strings<Encoding: _UnicodeEncoding>(
+        separator: StringSequence<Encoding>.Separator = .lineBreaks,
```
itingliu: I find the name `separator` conflicting with `.lineBreaks`. Separator is a singular noun, so I think it means "any of the separator in the collection of the specified separator*s*", but that's not what the `.unicodeScalars` function do (see the other comment)

Bikeshedding time... what about `separatedBy`? It would match Foundation's `func components<T : StringProtocol>(separatedBy separator: T) -> [String]`

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -382,6 +461,59 @@ extension AsyncBufferSequence.LineSequence {
...
+        /// as U+00E9 (precomposed) doesn't match "é" encoded as
+        /// U+0065 U+0301 (decomposed). Make sure the separator
+        /// scalars use the same representation as the input data.
+        ///
+        /// - Parameter separators: The scalars that form
+        ///   the delimiter.
+        /// - Returns: A separator that matches the given
+        ///   scalar sequence.
+        public static func unicodeScalars(
+            _ separators: some Collection<Unicode.Scalar>
```
itingliu: Just looking at the function name, I would't be sure if it means it'd split if the string matches *all* specified in this collection, or just *any*. Like it's not clear to me if `.unicodeScalars(["?", ":"])`  matches `"?:"` or `"?" | ":"`

I believe you mean the former. In that case would it be reasonable to add an overload that takes a String (and a Substring) for that case? stdlib does this for their `split` functions. Or give this some name that indicates this, like ".sequence"?

```
try await line in standardOutput.strings(separator: "?:"))
```

or 
```
try await line in standardOutput.strings(separator: .sequence(["?", ":"]))
```

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -382,6 +461,59 @@ extension AsyncBufferSequence.LineSequence {
...
+
+        internal init(_ storage: Storage) {
+            self.storage = storage
+        }
+
+        /// Splits on Unicode line break characters.
+        /// The following Unicode characters are recognized as line
+        /// breaks:
+        /// ```
+        /// LF:    Line Feed, U+000A
```
itingliu, line 498: Curious, how do you decide what this list should contain? Do they all belong to the same unicode category?

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: Ahh good idea thanks!

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: What if we change this to `unicodeScalarSequence`? To emphases we are treating the whole thing as a sequence, rather than "or".

Sources/Subprocess/AsyncBufferSequence.swift
iCharlesHu: This is taken from Wikipedia: https://en.wikipedia.org/wiki/Newline

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -392,6 +476,59 @@ extension AsyncBufferSequence.LineSequence {
...
+        /// scalars use the same representation as the input data.
+        ///
+        /// - Parameter separators: The scalars that form
+        ///   the delimiter.
+        /// - Returns: A separator that matches the given
+        ///   scalar sequence.
+        public static func unicodeScalarSequence(
+            _ separators: some Sequence<Unicode.Scalar>
+        ) -> Self {
+            if let array = separators as? Array<Unicode.Scalar> {
```
itingliu: As we talked we can add an overload to take `separator: Array<Unicode.Scalar>` so it takes the fast path

iCharlesHu:
Added a overload to `Separator.unicodeScalarSequence` that accepts `Array` directly as a performance optimization.

itingliu (approved):
cool!

PR #233, opened 2026-04-06
Adopt NonisolatedNonsendingByDefault and rename .standardOutput and .standardError output types to .currentStandardOutput and .currentStandardError

iCharlesHu:
Renaming `.standardOutput` and `.standardError` to `.currentStandardOutput` and `.currentStandardError` makes it more clear that the output is written to the parent process's' standard output rather than the child process's.

### DIFF
--- a/Sources/Subprocess/AsyncBufferSequence.swift
+++ b/Sources/Subprocess/AsyncBufferSequence.swift
@@ -50,7 +50,7 @@ public struct AsyncBufferSequence: AsyncSequence, @unchecked Sendable {
         }
 
         /// Retrieves the next buffer in the sequence, or `nil` if the sequence ended.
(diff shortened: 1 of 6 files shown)


jakepetroules (approved):

Sources/Subprocess/AsyncBufferSequence.swift
```diff
@@ -178,7 +178,7 @@ extension AsyncBufferSequence {
             /// Retrieves the next line, or `nil` if the sequence ended.
             public mutating func next() async throws -> String? {
```
FranzBusch: You probably want to adopt the `next(isolation:)` method in all these `AsyncSequences` as well

iCharlesHu: Will do. Thanks!

PR #234, opened 2026-04-08
[POC] Introduce OutputMethod and ErrorOutputMethod to replace OutputProtocol and ErrorProtocol

itingliu:
Added `public struct OutputMethod<Output>` to replace `OutputProtocol`. This also includes the following changes

Remove all the `OutputProtocol` conforming types. Replace them with static vars and funcs of `OutputMethod`


jakepetroules:
Is this change still relevant after recent refactors or should it be closed?

PR #235, opened 2026-04-08
Add a test for a custom OutputProtocol-conforming type

itingliu:
- Implement unimplemented functions. Replace fatalError wtih comments explaining why the function is no-op
- Add a test for custom output type
- Remove the comment that discourages providing a custom Output type

### DIFF
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -24,9 +24,6 @@ internal import Dispatch
 // MARK: - Output
 
 /// A type that serves as the output target for a subprocess.
-///
-/// Instead of creating custom implementations of ``OutputProtocol``, use the
-/// built-in implementations provided by the `Subprocess` library.
 public protocol OutputProtocol: Sendable, ~Copyable {
     associatedtype OutputType: Sendable
 
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -163,7 +163,7 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
 
     /// Creates an array from a ``RawSpan``.
     public func output(from span: RawSpan) throws -> [UInt8] {
-        span.withUnsafeBytes { Arra($0) }
+        span.withUnsafeBytes { Array($0) }
     }
 
     internal init(limit: Int) {
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -111,12 +111,10 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOut
 
     /// Creates a string from a raw span.
     public func output(from span: RawSpan) throws -> String? {
-        // FIXME: Span to String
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/IO/Output.swift
```diff
@@ -102,20 +109,27 @@ public struct FileDescriptorOutput: OutputProtocol, ErrorOutputProtocol {
...
 
-    /// Creates a string from a raw span.
+    #if SubprocessSpan
+    /// Create a string from a raw span.
     public func output(from span: RawSpan) throws -> String? {
-        // FIXME: Span to String
-        var array: [UInt8] = []
-        for index in 0..<span.byteCount {
-            array.append(span.unsafeLoad(fromByteOffset: index, as: UInt8.self))
+        span.withUnsafeBytes { ptr in
```
iCharlesHu, line 111: (not a issue) This TODO was more about switching to a String initializer that takes a Span directly. I don't think that API is ready yet unfortunately.

Sources/Subprocess/IO/Output.swift
```diff
@@ -163,9 +178,15 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
         #endif
     }
 
-    /// Creates an array from a ``RawSpan``.
+    #if SubprocessSpan
```
iCharlesHu: I think your fork is out of date. We removed this flag/trait when we dropped 6.1 support.

Tests/SubprocessTests/ProtocolConformanceTests.swift
```diff
@@ -0,0 +1,70 @@
...
+import Subprocess
+
+#if canImport(Darwin)
+import Foundation
+#else
+import FoundationEssentials
+#endif
+
+@Suite(.serialized)
+struct ProtocolConformanceTests {
```
iCharlesHu, line 22: (nit) Any particular reason to create a new test file instead of just reusing IntegrationTests?

iCharlesHu (approved):
Thank youuu!

Tests/SubprocessTests/ProtocolConformanceTests.swift
itingliu: I'm used to having a struct for one particular kind of test, so I created another one here. Honestly I didn't look closely what IntegrationTest contains. I can move back if you prefer

Sources/Subprocess/IO/Output.swift
itingliu: Even better. Let me rebase

Sources/Subprocess/IO/Output.swift
itingliu: Yeah I know the intention. I was looking for it so eagerly the other day too. For now at least this change stops us from potentially reallocating the array... still need a Span one though unfortunately.

Tests/SubprocessTests/ProtocolConformanceTests.swift
iCharlesHu: Ahh not at all. It's probably a good idea to split them up anyways.

itingliu:
@swift-ci please test

PR #238, opened 2026-04-11
Enable FreeBSD CI

jakepetroules:
Also fixes FreeBSD/OpenBSD build issues due to different pthread definitions, and an ordering issue where setgroups, setuid, and setgid were called in a different order from the macOS implementations (setgroups needs root so must be called first).


Tests/SubprocessTests/UnixTests.swift
```diff
@@ -95,19 +95,26 @@ extension SubprocessUnixTests {
         }
         var platformOptions = PlatformOptions()
         platformOptions.supplementaryGroups = Array(expectedGroups)
+        // Use /usr/bin/id instead of `swift` to avoid dynamic linker
+        // issues: setgroups() replaces all supplementary groups, which
+        // can prevent the dynamic linker from finding libswiftCore.so
+        // on systems where library paths require specific group access.
         let idResult = try await Subprocess.run(
-            .name("swift"),
-            arguments: [getgroupsSwift.string],
```
iCharlesHu, line 100: Please you also remove this entry from `TestResources.swift`

iCharlesHu (approved):
FreeBSD tests are failing... seems like infra?

jakepetroules:
OK, fixed up a couple test reliability issues that this exposed (needed to ignore SIGPIPE, and testRunawayProcess had nondeterministic timing sensitivity).

Issue #67, opened 2025-06-10
[Feature] Option to terminate any subprocesses launched by the process


dave256:
**Is your feature request related to a problem? Please describe.**
If the created process spawns subprocesses (such as running a shell script and the script spawns a Python program), it seems the teardown method doesn't kill the subprocesses. This is problematic if the spawned subprocess has entered an infinite loop/hang. I made a fork and tried changing the parameter of the `toProcessGroup` parameter in `send(signal:toProcessGroup:)` method to true but that still didn't seem to work.

**Describe the solution you'd like**
There should be an option to also kill any subprocesses spawned by the process.

**Describe alternatives you've considered**
I don't have any alternatives other than perhaps running it in a separate XPC process.

**Additional context**
None

broken-circle:
I ran into the same issue and opened #243 to address it.

If merged, setting `PlatformOptions.sendTeardownSignalsToProcessGroup` to `true` will ensure this behavior during teardown.

One caveat: Unless you also set `createSession` to `true`, or `processGroupID` to a non-inherited value, the targeted process group includes the parent process. This is almost never what you want. Pairing it with `createSession` isolates the subprocess and its descendants in their own session.

dave256:
@broken-circle thanks! I hope this works correctly and is merged. I gave up on subprocess for now but will look into it again if this is merged.

iCharlesHu:
Fixed with https://github.com/swiftlang/swift-subprocess/pull/243

PR #243, opened 2026-04-25
Target process group during teardown when configured

broken-circle:
Post-merge updates:

- Teardown configuration is no longer controlled through `PlatformOptions`, but rather through `TeardownStep.send()` and `TeardownStep.gracefulShutDown()`, which both gain a new `toProcessGroup: Bool` parameter (default `false`).
-  When `toProcessGroup` is `true`, that step's signal targets the subprocess's entire process group instead of only the subprocess itself, so backgrounded grandchildren aren't orphaned.
-  The implicit final SIGKILL inherits toProcessGroup from the last step in the sequence.
Because the subprocess inherits the parent's process group by default, pair toProcessGroup with createSession (or a non-inherited processGroupID) so the group-targeted signals don't also reach the parent process.

The original PR description is below. Please see the review comments for more information regarding the design changes.

---

Add `sendTeardownSignalsToProcessGroup` to Unix `PlatformOptions`, allowing signals emitted during teardown (including the final `.kill`) to target the subprocess's entire process group, rather than only the subprocess itself.

## Motivation

`Execution.send(signal:toProcessGroup:)` exposes the ability to signal either a single process or its entire process group. This flexibility matters when a subprocess spawns descendants of its own. Signaling only the leader leaves backgrounded grandchildren orphaned and still running.

`TeardownStep` offers no equivalent control. Every signal path inside `runTeardownSequence()` hard-codes `toProcessGroup: false`. A user who configures `teardownSequence` expecting robust termination of a subprocess tree gets group-unaware signaling with no way to opt in.

The workaround is to abandon `teardownSequence` entirely and reimplement the timeout/grace-period escalation pattern manually, which defeats the point of the abstraction.

This was reported as #67 in June 2025 and assigned to the Post-1.0 milestone.

## Design

`sendTeardownSignalsToProcessGroup` is a Boolean on Unix `PlatformOptions`. Windows isn't affected, since its final teardown step is `TerminateProcess` with no group concept.

I considered an alternative of adding `toProcessGroup: Bool` as a parameter on `TeardownStep.send()` and `gracefulShutDown()`, but decided against it for two reasons. First, `runTeardownSequence()` always appends an implicit final `.kill` step that has no user-visible construction site to which to attach a per-step flag, so the setting would have to either default to a hard-coded value or be specified on the teardown call itself. Second, scoping the option at `PlatformOptions` parallels the existing `teardownSequence` property. I could revise the PR if the per-step approach is preferred.

## Testing

Added `testTeardownSignalsProcessGroup()` in `UnixTests`, paired with the existing `testRunawayProcess()`.

No other tests needed changes, and all pass when run locally.

### DIFF
--- a/Sources/Subprocess/Teardown.swift
+++ b/Sources/Subprocess/Teardown.swift
@@ -31,9 +31,9 @@ import Musl
 public struct TeardownStep: Sendable, Hashable {
     internal enum Storage: Sendable, Hashable {
         #if !os(Windows)
-        case sendSignal(Signal, allowedDuration: Duration)
+        case sendSignal(Signal, toProcessGroup: Bool, allowedDuration: Duration)
         #endif
-        case gracefulShutDown(allowedDuration: Duration)
+        case gracefulShutDown(toProcessGroup: Bool, allowedDuration: Duration)
         case kill
     }
     var storage: Storage
@@ -43,13 +43,22 @@ public struct TeardownStep: Sendable, Hashable {
     /// before proceeding to the next step.
     ///
     /// The final step in the sequence always sends a `.kill` signal.
+    ///
+    /// - Important: When sending the signal to the process group, unless you
+    /// also set `createSession` to `true`, or `processGroupID` to a
+    /// non-inherited value, the targeted process group includes the parent
+    /// process. This is almost never what you want. Pair `toProcessGroup`
+    /// with `createSession` to isolate the subprocess and its descendants in
+    /// their own session.
     public static func send(
         signal: Signal,
+        toProcessGroup: Bool = false,
         allowedDurationToNextStep: Duration
     ) -> Self {
         return Self(
             storage: .sendSignal(
                 signal,
+                toProcessGroup: toProcessGroup,
                 allowedDuration: allowedDurationToNextStep
             )
         )
@@ -64,11 +73,22 @@ public struct TeardownStep: Sendable, Hashable {
     ///   1. Sends `WM_CLOSE` if the child process is a GUI process.
     ///   2. Sends `CTRL_C_EVENT` to the console.
     ///   3. Sends `CTRL_BREAK_EVENT` to the process group.
+    ///
+    /// - Important: On Unix, when sending the signal to the process group,
+    /// unless you also set `createSession` to `true`, or `processGroupID`
+    /// to a non-inherited value, the targeted process group includes the parent
+    /// process. This is almost never what you want. Pair `toProcessGroup`
+    /// with `createSession` to isolate the subprocess and its descendants in
+    /// their own session. On Windows, the `toProcessGroup` parameter has no
+    /// effect; `WM_CLOSE` and `CTRL_C_EVENT` have no process-group equivalent,
+    /// and `CTRL_BREAK_EVENT` is always sent to the process group.
     public static func gracefulShutDown(
+        toProcessGroup: Bool = false,
         allowedDurationToNextStep: Duration
     ) -> Self {
         return Self(
             storage: .gracefulShutDown(
+                toProcessGroup: toProcessGroup,
                 allowedDuration: allowedDurationToNextStep
             )
         )
@@ -95,6 +115,7 @@ internal enum TeardownStepCompletion {
 
 extension Execution {
     internal func gracefulShutDown(
+        toProcessGroup: Bool,
         allowedDurationToNextStep duration: Duration
     ) async {
         #if os(Windows)
@@ -129,7 +150,7 @@ extension Execution {
         // Send SIGTERM
         try? self.send(
             signal: .terminate,
-            toProcessGroup: false
+            toProcessGroup: toProcessGroup
         )
         #endif
     }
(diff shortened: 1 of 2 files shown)


iCharlesHu:
Hi @broken-circle, thanks for the fix! I agree with the fix itself, but I think we should make some API tweaks: instead of adding `sendTeardownSignalsToProcessGroup` to `PlatformOptions`, can you modify `TeardownStep`?  `TeardownStep` already has a public static func

```swift
struct TeardownStep {
    public static func send(
        signal: Signal,
        allowedDurationToNextStep: Duration
    ) -> Self
}
```

We can simply add a `toProcessGroup` parameter to this type. 

```swift
struct TeardownStep {
    public static func send(
        signal: Signal,
        toProcessGroup: Bool = false,
        allowedDurationToNextStep: Duration
    ) -> Self
}
```

This design allows individual `TeardownStep` to control whether it should send to the process group instead of having a global control.

broken-circle:
Hi @iCharlesHu, thanks for the quick review! Sure thing, `toProcessGroup` is now a per-step parameter on `send()`, defaulting to `false`.

While implementing it, I ran into a consistency problem and wanted to mention it before you re-review. The teardown sequence has three places that send signals: explicit `send()` steps, `gracefulShutDown()` (which sends `SIGTERM` on Unix), and the implicit final `kill` appended by `runTeardownSequence()`.

If only `send()` accepts `toProcessGroup`, then a user enabling that behavior for a subprocess that spawns grandchildren will find that if the initial signal doesn't cause the subprocess to clean up cooperatively, then the framework's `SIGKILL` will go only to the subprocess itself, leaving the grandchildren orphaned.

I made two changes to ensure consistent teardown behavior:

- `gracefulShutDown()` also accepts `toProcessGroup`, defaulting to `false`.
    - This is an unconditional parameter for cross-platform call-site consistency, and I documented it as a no-op on Windows. `WM_CLOSE` and `CTRL_C_EVENT` have no process-group equivalents, and `CTRL_BREAK_EVENT` is always sent to the process group. The alternative would be to gate the parameter behind `#if !os(Windows)` so the API doesn't carry a no-op parameter, at the cost of forcing cross-platform call sites to use platform conditionals. I can switch to the conditional parameter if you'd prefer it.
- The implicit final `kill` inherits `toProcessGroup` from the last user-supplied step. The enum case stays bare.

If these changes go beyond the scope you imagined, I can update the PR to only add `toProcessGroup` to `send()`, and follow it up with another PR for the additional changes after this one is merged.

iCharlesHu:
> Hi @iCharlesHu, thanks for the quick review! Sure thing, `toProcessGroup` is now a per-step parameter on `send()`, defaulting to `false`.
> 
> While implementing it, I ran into a consistency problem and wanted to mention it before you re-review. The teardown sequence has three places that send signals: explicit `send()` steps, `gracefulShutDown()` (which sends `SIGTERM` on Unix), and the implicit final `kill` appended by `runTeardownSequence()`.
> 
> If only `send()` accepts `toProcessGroup`, then a user enabling that behavior for a subprocess that spawns grandchildren will find that if the initial signal doesn't cause the subprocess to clean up cooperatively, then the framework's `SIGKILL` will go only to the subprocess itself, leaving the grandchildren orphaned.
> 
> I made two changes to ensure consistent teardown behavior:
> 
> * `gracefulShutDown()` also accepts `toProcessGroup`, defaulting to `false`.
>   
>   * This is an unconditional parameter for cross-platform call-site consistency, and I documented it as a no-op on Windows. `WM_CLOSE` and `CTRL_C_EVENT` have no process-group equivalents, and `CTRL_BREAK_EVENT` is always sent to the process group. The alternative would be to gate the parameter behind `#if !os(Windows)` so the API doesn't carry a no-op parameter, at the cost of forcing cross-platform call sites to use platform conditionals. I can switch to the conditional parameter if you'd prefer it.
> * The implicit final `kill` inherits `toProcessGroup` from the last user-supplied step. The enum case stays bare.
> 
> If these changes go beyond the scope you imagined, I can update the PR to only add `toProcessGroup` to `send()`, and follow it up with another PR for the additional changes after this one is merged.

Makes sense to me. Thanks for the thoughtful considerations!

Sources/Subprocess/Teardown.swift
```diff
@@ -64,11 +73,22 @@ public struct TeardownStep: Sendable, Hashable {
...
     ///   2. Sends `CTRL_C_EVENT` to the console.
     ///   3. Sends `CTRL_BREAK_EVENT` to the process group.
+    ///
+    /// - Important: On Unix, when sending the signal to the process group,
+    /// unless you also set `createSession` to `true`, or `processGroupID`
+    /// to a non-inherited value, the targeted process group includes the parent
+    /// process. This is almost never what you want. Pair `toProcessGroup`
+    /// with `createSession` to isolate the subprocess and its descendants in
+    /// their own session. On Windows, the `toProcessGroup` parameter has no
+    /// effect; `WM_CLOSE` and `CTRL_C_EVENT` have no process-group equivalent,
```
jakepetroules, line 83: Windows does have Job Objects, which allow some grouping-like operations on processes, could we explore how use of those APIs might fit into this?

broken-circle: @jakepetroules Sure, good call. I spent some time reading up on Job Objects and they fit nicely on `kill` using [`TerminateJobObject`](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-terminatejobobject). I don't think the other steps map directly to Job Objects since they propagate along different axes (`WM_CLOSE` is per-window, `CTRL_C_EVENT` is per-console, and `CTRL_BREAK_EVENT` already targets the Windows process group at spawn).

Adopting this behavior on `kill` means assigning each spawned child to a job. `Subprocess+Windows.swift` has the right shape for that, but the implementation would have its own design decisions worth exploring in depth (handle ownership and inheritance, breakaway behavior, and so on). I'd like to do this in a follow-up PR rather than expand this one. Does that sound good to you and @iCharlesHu?

broken-circle:
@iCharlesHu Thanks for running CI. Looks like it caught a stray tab on a blank line. Fixed now. Should be all green on the next run.

Sources/Subprocess/Teardown.swift
iCharlesHu: I agree. Let's tackle Windows in a separate PR.

iCharlesHu:
> @iCharlesHu Thanks for running CI. Looks like it caught a stray tab on a blank line. Fixed now. Should be all green on the next run.

It looks like it failed again. Please run `swift-format -i -r .` to format it

broken-circle:
@iCharlesHu Done, pushed.

iCharlesHu (approved):
Thanks for the fix!

PR #254, opened 2026-04-28
Fix streaming read hangs by replacing persistent DispatchIO with per-read dispatch_read

### LINKED ISSUE #252: output streaming doesn't appear to work
**Describe the bug**

I wrote a basic program that launches a subprocess which writes back an `OK` and then just spins in a loop. The parent is meant to receive the `OK` and then cancel the task which is then meant to stop the subprocess. Unfortunately, it just hangs, the `OK` never gets received.

**To Reproduce**

Run

```
    func testCancelAllTearsDownProcess() async throws {
        enum WhoReturned {
            case okReceived
            case finished(Subprocess.ExecutionOutcome<Void>)
        }
        struct WrongLineReceived: Error {
            var line: String
        }

        let (okSignal, okSignalContinuation) = AsyncThrowingStream.makeStream(of: Void.self)

        try await withThrowingTaskGroup(of: WhoReturned.self) { group in
            group.addTask {
                let outcome = try await Subprocess.run(
                    .path("/bin/sh"),
                    arguments: ["-c", "echo OK; while true; do sleep 0.1; done"],
                    error: .discarded
                ) { execution, outputSequence in
                    for try await line in outputSequence.strings(separatedBy: .lineBreaks) {
                        if line == "OK" {
                            okSignalContinuation.yield(())
                            okSignalContinuation.finish()
                        } else {
                            XCTFail("unexpected output: \(line)")
                            okSignalContinuation.finish(throwing: WrongLineReceived(line: line))
                        }
                        break
                    }
                }
                return .finished(outcome)
            }

            group.addTask {
                var iterator = okSignal.makeAsyncIterator()
                _ = try await iterator.next()
                return .okReceived
            }

            var finalOutcome: Subprocess.ExecutionOutcome<Void>? = nil
            while let taskResult = try await group.next() {
                switch taskResult {
                case .okReceived:
                    group.cancelAll()
                case .finished(let outcome):
                    finalOutcome = outcome
                }
            }

            XCTAssertNotNil(finalOutcome)
        }
    }
```

**Expected behavior**
Does not hang.

**Additional context**

This is a simplified version of [AsyncProcess's `testShutdownSequenceWorks`](https://github.com/apple/swift-llbuild2/blob/bc6abf3eb3498308745e5f1b43af789f81932c6c/Tests/FXAsyncSupportTests/IntegrationTests.swift#L1151) which would also fail.

jakepetroules:
The Darwin AsyncIO implementation used a single persistent DispatchIO channel with read(length: .max), which only completes at EOF. This caused AsyncBufferSequence to hang whenever a subprocess produced less output than the buffer size, because the read would never return until the pipe closed. The preferredBufferSize parameter existed solely as a workaround — setting it to 1 made reads return byte-by-byte — but the real fix is to make reads return available data without waiting for a full buffer.

Root cause: DispatchIO.read(length: .max) accumulates data internally and only resumes the continuation when done=true (EOF). The low water mark controls handler invocation granularity but not when the operation completes. This is fundamentally incompatible with streaming reads.

The fix replaces the persistent DispatchIO channel with the class-level dispatch_read/dispatch_write convenience APIs, which create a temporary channel per operation and return whatever data is currently available:

Darwin (AsyncIO+Dispatch.swift):
- read() calls DispatchIO.read(fromFileDescriptor:maxLength:) which returns one chunk of available data per call (streaming semantics)
- read(upTo: .max) loops internally until EOF (bulk semantics)
- New readAll(from:upTo:) loops until maxLength bytes or EOF, used by captureOutput for collecting complete subprocess output
- write() uses DispatchIO.write(toFileDescriptor:) with a retry loop for any unwritten data, matching Linux/Windows behavior
- IOChannel on Darwin now stores a FileDescriptor instead of DispatchIO, removing the persistent channel entirely

Linux (AsyncIO+Linux.swift) and Windows (AsyncIO+Windows.swift):
- Refactored read into _read(from:upTo:returnOnFirstRead:) to share implementation between single-chunk and bulk modes
- read() with finite maxLength returns after the first successful read syscall (single-chunk, for streaming)
- read(upTo: .max) loops until EOF (bulk, same as before)
- New readAll() loops until maxLength bytes or EOF

This gives consistent cross-platform semantics:
- read(upTo: finite) → single chunk (used by AsyncBufferSequence)
- read(upTo: .max) → loop until EOF
- readAll(upTo: N) → loop until N bytes or EOF (used by captureOutput)

The preferredBufferSize parameter is now functionally unnecessary and can be removed in a subsequent change.

Closes #252

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Dispatch.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Dispatch.swift
@@ -33,49 +33,81 @@ final class AsyncIO: Sendable {
         from diskIO: borrowing IOChannel,
         upTo maxLength: Int
     ) async throws(SubprocessError) -> DispatchData? {
-        return try await self.read(
-            from: diskIO.channel,
-            upTo: maxLength,
+        if maxLength == .max {
+            return try await self.readAll(from: diskIO, upTo: maxLength)
+        }
+        return try await self.readChunk(
+            fd: diskIO.channel.rawValue,
+            maxLength: maxLength,
         )
     }
 
     internal func read(
-        from dispatchIO: DispatchIO,
+        from fileDescriptor: FileDescriptor,
         upTo maxLength: Int
+    ) async throws(SubprocessError) -> DispatchData? {
+        if maxLength == .max {
+            return try await self.readAll(
+                fd: fileDescriptor.rawValue,
+                upTo: maxLength
+            )
+        }
+        return try await self.readChunk(
+            fd: fileDescriptor.rawValue,
+            maxLength: maxLength,
+        )
+    }
+
+    internal func readAll(
+        from diskIO: borrowing IOChannel,
+        upTo maxLength: Int
+    ) async throws(SubprocessError) -> DispatchData? {
+        return try await self.readAll(
+            fd: diskIO.channel.rawValue,
+            upTo: maxLength
+        )
+    }
+
+    private func readAll(
+        fd: Int32,
+        upTo maxLength: Int
+    ) async throws(SubprocessError) -> DispatchData? {
+        var accumulated: DispatchData = .empty
+        while accumulated.count < maxLength {
+            if Task.isCancelled {
+                throw SubprocessError.asyncIOFailed(
+                    reason: "Cancelled",
+                    underlyingError: nil
+                )
+            }
+            let remaining = maxLength == .max ? .max : maxLength - accumulated.count
+            let chunk: DispatchData? = try await self.readChunk(fd: fd, maxLength: remaining)
+            guard let chunk else { break }
+            accumulated.append(chunk)
+        }
+        return accumulated.isEmpty ? nil : accumulated
+    }
+
+    private func readChunk(
+        fd: Int32,
+        maxLength: Int
     ) async throws(SubprocessError) -> DispatchData? {
         // https://github.com/swiftlang/swift/issues/87810
         return try await _castError {
             return try await withCheckedThrowingContinuation { continuation in
-                var buffer: DispatchData = .empty
-                dispatchIO.read(
-                    offset: 0,
-                    length: maxLength,
-                    queue: DispatchQueue(label: "SubprocessReadQueue")
-                ) { done, data, error in
+                DispatchIO.read(
+                    fromFileDescriptor: fd,
+                    maxLength: maxLength,
+                    runningHandlerOn: .global()
+                ) { data, error in
                     if error != 0 {
                         continuation.resume(
-                            throwing:
-                                SubprocessError
-                                .failedToReadFromProcess(
-                                    withUnderlyingError: Errno(rawValue: error)
-                                )
-                        )
+                            throwing: SubprocessError.failedToReadFromProcess(
+                                withUnderlyingError: Errno(rawValue: error)
+                            ))
                         return
                     }
-                    if let data {
-                        if buffer.isEmpty {
-                            buffer = data
-                        } else {
-                            buffer.append(data)
-                        }
-                    }
-                    if done {
-                        if !buffer.isEmpty {
-                            continuation.resume(returning: buffer)
-                        } else {
-                            continuation.resume(returning: nil)
-                        }
-                    }
+                    continuation.resume(returning: data.isEmpty ? nil : data)
                 }
             }
         }
@@ -147,27 +179,31 @@ final class AsyncIO: Sendable {
         queue: DispatchQueue = .global(),
         completion: @escaping (Int, SubprocessError?) -> Void
     ) {
-        diskIO.channel.write(
-            offset: 0,
-            data: dispatchData,
-            queue: queue
-        ) { done, unwritten, error in
-            guard done else {
-                // Wait until we are done writing or encountered some error
-                return
+        let fd = diskIO.channel.rawValue
+        var totalWritten = 0
+        func writeRemaining(_ data: DispatchData) {
+            DispatchIO.write(
+                toFileDescriptor: fd,
+                data: data,
+                runningHandlerOn: queue
+            ) { unwritten, error in
+                let unwrittenLength = unwritten?.count ?? 0
+                totalWritten += data.count - unwrittenLength
+                if error != 0 {
+                    completion(
+                        totalWritten,
+                        .failedToWriteToProcess(withUnderlyingError: Errno(rawValue: error))
+                    )
+                    return
+                }
+                if let unwritten, !unwritten.isEmpty {
+                    writeRemaining(unwritten)
+                } else {
+                    completion(totalWritten, nil)
+                }
             }
-
-            let unwrittenLength = unwritten?.count ?? 0
-            let writtenLength = dispatchData.count - unwrittenLength
-            guard error != 0 else {
-                completion(writtenLength, nil)
-                return
-            }
-            completion(
-                writtenLength,
-                .failedToWriteToProcess(withUnderlyingError: Errno(rawValue: error))
-            )
         }
+        writeRemaining(dispatchData)
     }
 }
 
--- a/Sources/Subprocess/AsyncBufferSequence.swift
+++ b/Sources/Subprocess/AsyncBufferSequence.swift
@@ -38,9 +38,7 @@ public struct AsyncBufferSequence: AsyncSequence, @unchecked Sendable {
     /// The element type for the asynchronous sequence.
     public typealias Element = Buffer
 
-    #if SUBPROCESS_ASYNCIO_DISPATCH
-    internal typealias DiskIO = DispatchIO
-    #elseif canImport(WinSDK)
+    #if canImport(WinSDK)
     internal typealias DiskIO = HANDLE
     #else
     internal typealias DiskIO = FileDescriptor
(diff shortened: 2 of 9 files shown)


Sources/Subprocess/IO/AsyncIO+Dispatch.swift
```diff
@@ -33,49 +33,81 @@ final class AsyncIO: Sendable {
...
+            if Task.isCancelled {
+                throw SubprocessError.asyncIOFailed(
+                    reason: "Cancelled",
+                    underlyingError: nil
+                )
+            }
+            let remaining = maxLength == .max ? .max : maxLength - accumulated.count
+            let chunk: DispatchData? = try await self.readChunk(fd: fd, maxLength: remaining)
+            guard let chunk else { break }
+            accumulated.append(chunk)
```
weissi, line 86: This will break backpressure and therefore may lead to OOM crashes and uncontrollable resource usage if this is ever called with `maxLength` larger than what we want to tolerate at most.

So I'm very worried about the presence of this function.

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
```diff
@@ -33,49 +33,81 @@ final class AsyncIO: Sendable {
...
         )
     }
 
     internal func read(
-        from dispatchIO: DispatchIO,
+        from fileDescriptor: FileDescriptor,
+        upTo maxLength: Int
+    ) async throws(SubprocessError) -> DispatchData? {
+        if maxLength == .max {
+            return try await self.readAll(
```
weissi, line 50: This sure reads like it will break backpressure and will potentially hoover up infinite amounts of bytes into memory. Could you explain how this is safe?

weissi: The crucial invariant we need to uphold is that we _never_ read more than a clearly defined number of bytes for every time the user calls the `try await outputStream.next()` call.

We must also have a test case that proves this. The way I would suggest to write this test it to launch a shell with this program

```
set -eu

trap '' PIPE

roughly_6mb=$(echo {1..1000000})   # puts 6MB of data in this variable
for f in {1..20}; do
    echo "$f" >> /tmp/asdf
    echo "$roughly_6mb" || exit 2
done
exit 0
```

Then read a little bit, maybe 1kB or so, then _sleep_ for half a second. After that, validate the following:

- After the sleep, the subprocess is still running (because the `echo "$roughly_6mb"` is still blocked)
- Then cancel the Swift Concurrency task (which will tear down the process)
- Finally validate that the task exits in the expected way. The expected way would be
   - SIGKILL  # if we killed it first
   - or exit code 2   # if we closed the reading end of the pipe first which will make the `echo "$roughly_6mb"` fail and that will cause it to exit with code 2

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
```diff
@@ -33,49 +33,81 @@ final class AsyncIO: Sendable {
...
-                dispatchIO.read(
-                    offset: 0,
-                    length: maxLength,
-                    queue: DispatchQueue(label: "SubprocessReadQueue")
-                ) { done, data, error in
+                DispatchIO.read(
+                    fromFileDescriptor: fd,
+                    maxLength: maxLength,
+                    runningHandlerOn: .global()
+                ) { data, error in
```
weissi, line 102: This closure will be invoked _exactly once_, is that correct?

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
```diff
@@ -33,49 +33,81 @@ final class AsyncIO: Sendable {
...
         // https://github.com/swiftlang/swift/issues/87810
         return try await _castError {
             return try await withCheckedThrowingContinuation { continuation in
-                var buffer: DispatchData = .empty
-                dispatchIO.read(
-                    offset: 0,
-                    length: maxLength,
-                    queue: DispatchQueue(label: "SubprocessReadQueue")
-                ) { done, data, error in
+                DispatchIO.read(
```
weissi, line 98: How do we cancel such a read? We must be able to cancel our read, killing the child process is _not enough_. We also need a test for the following scenario:

1. We run a subprocess (process `A`) and stream its stdout (i.e. we create a pipe where we inherit the write end into the child)
2. The subprocess spawns _another_ subprocess (process `B`) and directly inherits on the pipe end that we provided
3. Now we cancel our subprocess which will eventually kill process `A` but `B` _remains alive still has our file descriptor_.
4. The verify the expectations: Despite the fact that our file descriptor's write end is still open (in process `B`), even if nothing more has been written, we are not hanging.

AsyncProcess has a similar test case (that one tests stdin not stdout, we need both)

```
    func testWeDoNotHangIfStandardInputRemainsOpenButProcessExits() async throws {
        // This tests an odd situation: The child exits but stdin is still not closed, mostly happens if we inherit a
        // pipe that we still have another writer to.

        var sleepPidToKill: CInt?
        defer {
            if let sleepPidToKill {
                self.logger.debug(
                    "killing our sleep grand-child",
                    metadata: ["pid": "\(sleepPidToKill)"]
                )
                kill(sleepPidToKill, SIGKILL)
            } else {
                XCTFail("didn't find the pid of sleep to kill")
            }
        }
        do { // We create a scope here to make sure we can leave the scope without hanging
            let (stdinStream, stdinStreamProducer) = AsyncStream.makeStream(of: ByteBuffer.self)
            let exe = ProcessExecutor(
                executable: "/bin/sh",
                [
                    "-c",
                #"""
                # This construction attempts to emulate a simple `sleep 12345678 < /dev/null` but some shells (eg. dash)
                # won't allow stdin inheritance for background processes...
                exec 2>&- # close stderr
                exec 2<&0 # duplicate stdin into fd 2 (so we can inherit it into sleep

                ( # this creates a child of our child which gets inherited our fds
                    exec 0<&2  # map the duplicated fd 2 as our stdin
                    exec 2>&-  # close the duplicated fd2
                    exec sleep 12345678 # sleep (this will now have the origin stdin as its stdin)
                ) & # uber long sleep that will inherit our stdin pipe
                exec 2>&- # close duplicated 2

                read -r line
                echo "$line" # write back the line
                echo "$!" # write back the sleep
                exec >&-
                exit 0
                """#
                ],
                standardInput: stdinStream
            )
            stdinStreamProducer.yield(ByteBuffer(string: "GO\n"))
            stdinStreamProducer.yield(ByteBuffer(repeating: 0x42, count: 16 * 1024 * 1024))
            async let resultAsync = exe.runWithExtendedInfo()
            async let stderrAsync = Array(exe.standardError)
            var stdoutLines = await exe.standardOutput.splitIntoLines().makeAsyncIterator()
            let lineGo = try await stdoutLines.next()
            XCTAssertEqual(ByteBuffer(string: "GO"), lineGo)
            let linePid = try await stdoutLines.next().map(String.init(buffer:))
            let sleepPid = try XCTUnwrap(linePid.flatMap { CInt($0) })
            self.logger.debug("found our sleep grand-child", metadata: ["pid": "\(sleepPid)"])
            sleepPidToKill = sleepPid
            let stderrBytes = try await stderrAsync
            XCTAssertEqual([], stderrBytes)
            let result = try await resultAsync
            XCTAssertEqual(.exit(0), result.exitReason)
            XCTAssertNotNil(result.standardInputWriteError)
            XCTAssertEqual(ChannelError.ioOnClosedChannel, result.standardInputWriteError as? ChannelError)
            stdinStreamProducer.finish()
        }
    }
```

weissi (changes requested):
Requesting changes because I'm pretty sure (happy to be corrected) that this suffers from two crucial issues:

1. Breaks backpressure (might load infinite amounts of data into memory)
2. Lacks the ability to cancel reads from file descriptors inherited into children that get inherited into yet another process

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
jakepetroules: Cancellation wasn't handled previously. Let's tackle that in a separate PR.

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
jakepetroules: Yes

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
jakepetroules: I think the only place where this can be called is for non-streaming output, where you can't stop the user from potentially setting output to `.bytes(limit: .max)`.

But I agree this could be misleading, let me see if I can refactor a bit more.

weissi: Okay thanks for the explainer, this makes sense. My suggestion would be to be stupidly obvious with the intent `readBytesContiguouslyIntoMemoryAsRequestedByUser(userRequestedMaxBytes: Int)` or something equally obvious that we don't end up accidentally calling elsewhere

Sources/Subprocess/IO/AsyncIO+Dispatch.swift
weissi: Okay, fair enough. @jakepetroules would you be able to make a fresh bug detailing the places where we don't handle cancellation which would break under the scenario? If this is a pre-existing issue I'm happy moving forward assuming we have an actual bug for that.

weissi: @jakepetroules Is this about right? https://github.com/swiftlang/swift-subprocess/issues/256

jakepetroules:
Closing; @iCharlesHu is taking this over via #259 instead

PR #259, opened 2026-05-02
Make `AsyncIO.read`'s `maxLength` a cap and rewrite Darwin backend

### LINKED ISSUE #252: output streaming doesn't appear to work

iCharlesHu:
Previously `AsyncIO.read(upTo: maxLength)` treated maxLength as a target length. It would accumulate bytes in a loop until it had gathered that many before returning. This deadlocks when a subprocess writes fewer bytes than `maxLength` and then pauses or waits on stdin: the parent blocks forever on data the child never sends.

This PR changes `maxLength` to a "cap", as its name suggests. Each `read(upTo:)` now returns as soon as the underlying nonblocking `read(2)` (or `ReadFile`) completes. Callers that need a specific byte count accumulate across calls themselves.

Follow-on cleanup enabled by the semantics fix:

- Replace Darwin's `DispatchIO` backend with `DispatchSource` + `read(2)`. The new Darwin path is a thin event-source shim over the same nonblocking read/write loop that Linux and Android already use, sharing implementation in a new `IO/AsyncIO+Unix.swift`.
- Collapse `IOChannel` into `IODescriptor`. `IOChannel` only existed to wrap a `DispatchIO`; with `DispatchIO` gone, a single fd-owning type suffices.
- Drop the public `preferredBufferSize` parameter from every `run()` overload.
    It was introduced to let callers work around the old hang; buffer size is now derived from the pipe's capacity via `F_GETPIPE_SZ` (Linux/Android), `fstat().st_blksize` (Darwin/OpenBSD), or a fixed 64 KB fallback (FreeBSD, Windows).

Resolves: https://github.com/swiftlang/swift-subprocess/issues/252


Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,339 @@
...
+        to diskIO: borrowing IODescriptor
+    ) async throws(SubprocessError) -> Int {
+        guard bytes.count > 0 else {
+            return 0
+        }
+        let fileDescriptor = diskIO.descriptor()
+        let signalStream = self.registerFileDescriptor(fileDescriptor, for: .write)
+        var writtenLength: Int = 0
+        do {
+            /// Outer loop: every iteration signals we are ready to read more data
```
jakepetroules: s/read/write/

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,339 @@
...
+#if canImport(System)
+import System
+#else
+import SystemPackage
+#endif
+
+#if canImport(Darwin)
+import Darwin
+
+let _subprocess_read = Darwin.read
```
jakepetroules: How about just using SystemPackage.FileDescriptor.read instead? This also allows you to handle EINTR more easily, which looks like it is missing.

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,339 @@
...
+                while true {
+                    let written = bytes.withUnsafeBytes { ptr in
+                        let remainingLength = ptr.count - writtenLength
+                        let startPtr = ptr.baseAddress!.advanced(by: writtenLength)
+                        return _subprocess_write(fileDescriptor.rawValue, startPtr, remainingLength)
+                    }
+                    let capturedErrno = errno
+                    if written > 0 {
+                        writtenLength += written
+                        if writtenLength >= bytes.count {
```
jakepetroules: fatalError if writtenLength > bytes.count? that should never be true

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,339 @@
...
+                )
+            }
+            throw SubprocessError.failedToWriteToProcess(
+                withUnderlyingError: originalError.underlyingError
+            )
+        }
+        return 0
+    }
+
+    func write(
```
jakepetroules, line 180: the two write overloads are near identical, I think they should share a common implementation

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,339 @@
...
+            throw SubprocessError.failedToWriteToProcess(
+                withUnderlyingError: originalError.underlyingError
+            )
+        }
+        return 0
+    }
+
+    @inline(__always)
+    private func shouldWaitForNextSignal(with error: CInt) -> Bool {
+        return error == EAGAIN || error == EWOULDBLOCK || error == EINTR
```
jakepetroules: are you sure we shouldn't just handle EINTR immediately, before waiting for another tick from the DispatchSource?

jakepetroules:
I'm confident at this point that we should dump the Dispatch implementation entirely and drop down to raw kqueue. Even DispatchSource is too buggy and appears to have inconsistent behavior between macOS and FreeBSD; I could not get it to work at all.

And if we need to use kqueue on FreeBSD we may as well use it on macOS as well and not maintain two implementations.

Take a look at my attempt here, this eliminated all the hangs including on FreeBSD: https://github.com/jakepetroules/swift-subprocess/commit/118c7e5192f250c60549fc6c954a109bec7f4174

iCharlesHu:
Updated Darwin implementation to `Kqueue` + `read`.

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -0,0 +1,392 @@
...
+
+    private let state: Result<State, SubprocessError>
+    #if canImport(Darwin)
+    private let shutdownFlag: OSAllocatedUnfairLock<Bool> = OSAllocatedUnfairLock(initialState: false)
+    #else
+    private let shutdownFlag: Atomic<UInt8> = Atomic(0)
+    #endif
+
+    internal init() {
+        let kqueueFileDescriptor = kqueue()
```
jakepetroules, line 115: Let's use the `kqueue1` overload which allows atomically setting the close-on-exec flag.

(conditional because this is a NetBSD innovation that the other BSDs copied, but macOS doesn't have it)

```suggestion
        #if os(FreeBSD) || os(OpenBSD)
        let kqueueFileDescriptor = kqueue1(O_CLOEXEC)
        #else
        let kqueueFileDescriptor = kqueue()
        #endif
```

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -0,0 +1,392 @@
...
+#else
+import SystemPackage
+#endif
+
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
```
jakepetroules: nit: can drop the Musl import as kqueue is a BSD-family API, so this file would never compile on Linux anyways

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -0,0 +1,392 @@
...
+        guard kqueueFileDescriptor >= 0 else {
+            let error: SubprocessError = .asyncIOFailed(
+                reason: "kqueue failed",
+                underlyingError: Errno(rawValue: errno)
+            )
+            self.state = .failure(error)
+            return
+        }
+        var shutdownPipe: [Int32] = [0, 0]
+        guard pipe(&shutdownPipe) == 0 else {
```
jakepetroules: this should use SwiftSystem FileDescriptor.pipe API; it provides a nicer interface. there is also a proposal out for adding a close-on-exec parameter that we'll want to adopt later: https://github.com/apple/swift-system/pull/266

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -233,21 +241,8 @@ final class AsyncIO: Sendable {
...
-                    continuation.finish(throwing: error)
-                    return
-                }
-                guard fcntl(fileDescriptor.rawValue, F_SETFL, flags | O_NONBLOCK) != -1 else {
-                    let error: SubprocessError = .asyncIOFailed(
-                        reason: "failed to set \(fileDescriptor.rawValue) to be non-blocking",
-                        underlyingError: Errno(rawValue: errno)
-                    )
-                    continuation.finish(throwing: error)
+                if let nonBlockingFdError = self.setNonblocking(for: fileDescriptor) {
```
jakepetroules, line 244: would be nice if we could create the fds as nonblocking out of the gate.

are _all_ of the file descriptors passed into this function created by `pipe`? if so, we could use `pipe2` on the platforms which have it (so basically everything but macOS), to avoid the extra syscall there

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -0,0 +1,283 @@
...
+            for try await _ in signalStream {
+                /// Inner loop: repeatedly call `.write()` and write more data until:
+                /// 1. We've written bytes.count bytes.
+                /// 2. `FileDescriptor.write()` throwws `Errno` with `EAGAIN` or `EWOULDBLOCK`. In
+                ///     this case we `break` out of the inner loop and wait `.write()` to be
+                ///     ready by `await`ing the next signal in the outer loop.
+                while true {
+                    let written: Int
+                    do {
+                        written = try span.withUnsafeBytes { ptr in
```
jakepetroules: should this use `span.extracing()` to do the math parts, _then_ call withUnsafeBytes in the result of that, and pass the ptr directly to fileDescriptor.write?

That seems like it might be slightly safer

Sources/Subprocess/IO/Input.swift
```diff
@@ -237,12 +237,16 @@ extension InputProtocol {
...
-    internal var diskIO: IOChannel
+    internal var diskIO: IODescriptor
 
-    init(diskIO: consuming IOChannel) {
+    init(diskIO: consuming IODescriptor) {
         self.diskIO = diskIO
     }
 
+    deinit {
+        try? self.diskIO.safelyClose()
```
jakepetroules: could this use assert that it's been closed instead? this way we can have the call site be responsible and not drop the error

a lot of good guidance for Swift generally has been to not use deinit to clean up non-memory resources (and this is a non-memory resource)

Sources/Subprocess/IO/AsyncIO+Linux.swift
jakepetroules: and if this can be a user supplied fd, should we dup and set O_NONBLOCK on that one, instead of modifying the original fd?

jakepetroules (commented):
This is looking really good! Left a few minor points of feedback.

Sources/Subprocess/IO/Input.swift
iCharlesHu: Actually I think I left this in as a debugging thing. I don't think it's necessary to have it.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: All fds passed to `AsyncIO` are created by `pipe`. However, we can't just create the pipe as non-blocking because both ends would be non-blocking, and that usually breaks the child process. I think we are going to have to make the extra `fcntl` call either to add `O_NONBLOCK` to the parent end or remove it from the child end, so we might as well keep the existing behavior.

jakepetroules: ok, sounds good

jakepetroules (approved):

PR #260, opened 2026-05-04
Support Windows `toProcessGroup` teardown

broken-circle:
Post-merge updates:

- Every Windows child is now assigned to a Job Object at spawn, so Subprocess always passes `CREATE_SUSPENDED` to `CreateProcessW`. Subprocess always resumes the child once it's in the job, unless a `preSpawnProcessConfigurator` sets `CREATE_SUSPENDED` itself; that signals that the caller will resume the child, so Subprocess skips its own `ResumeThread` and leaves the thread suspended for the caller to resume via `Execution.processIdentifier.threadHandle`.

The original PR description is below. Please see the review comments for more information regarding the design changes.

---

This PR expands the Unix child process teardown implemented in #243 to Windows.

## Motivation

#243 added a `toProcessGroup: Bool` parameter to `send()` and `gracefulShutDown()` on `TeardownStep`, with the final `kill` inheriting `toProcessGroup` from the last user-supplied step. On Windows, the final `kill` step ignored that parameter and only terminated the immediate child, leaving descendants unaffected. This PR closes that gap.

## Design

Windows has no direct equivalent of Unix process groups, but [Job Objects](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects) provide a similar grouping mechanism. A process assigned to a job, along with all of its descendants, can be terminated together using [`TerminateJobObject`](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-terminatejobobject).

Job Objects fit cleanly into the final `kill` step. Other teardown steps already propagate along different axes and don't benefit from Job Objects: `WM_CLOSE` is per-window, `CTRL_C_EVENT` is per-console, and `CTRL_BREAK_EVENT` targets the Windows process group that was set up at spawn.

To make `TerminateJobObject` available at teardown time, every Windows child process is assigned to a job object at spawn. The job is created in `spawnDirect()` and `spawnAsUser()` on `Configuration`, and lives for the duration of the `Execution`.

### Atomic assignment using `CREATE_SUSPENDED`

`Configuration.generateCreateProcessFlag()` now always sets `CREATE_SUSPENDED`.  After running `preSpawnProcessConfigurator`, both spawn paths verify that the flag is still set and throw `spawnFailed` if a user-supplied configurator removed it. `CREATE_SUSPENDED` makes the assignment atomic with respect to the child, ensuring that between `CreateProcessW` and `AssignProcessToJobObject`, no user code runs in the child and it cannot spawn descendants outside the job. Once the assignment succeeds, `ResumeThread` lets the child begin execution.

In practice, the suspension period is imperceptible, but this constitutes a small behavioral change. Users who explicitly clear `CREATE_SUSPENDED` will now receive an error.

A possible alternative is `PROC_THREAD_ATTRIBUTE_JOB_LIST`, which assigns the job at process creation time without the suspend-then-resume requirement. This requires `lpStartupInfo` to be `STARTUPINFOEX` and the `EXTENDED_STARTUPINFO_PRESENT` process creation flag to be present in `dwCreationFlags`.

[`CreateProcessW` documentation](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) explicitly supports both `STARTUPINFO` and `STARTUPINFOEX`, but [`CreateProcessWithLogonW` documentation](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createprocesswithlogonw) specifies support for only `STARTUPINFO`. Since `spawnAsUser()` calls `CreateProcessWithLogonW`, using `PROC_THREAD_ATTRIBUTE_JOB_LIST` would require two implementations of the same behavior. The suspend-and-resume pattern works uniformly for both paths.

### Handle inheritance

The Job Object handle is created with `nil` security attributes, which makes it non-inheritable. As [Raymond Chen notes](https://devblogs.microsoft.com/oldnewthing/20131209-00/?p=2433):

> We must therefore be careful not to allow this handle to be inherited, because that would create another handle that needs to be closed before the job is terminated.

The [`CreateJobObjectW` documentation](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-createjobobjectw) confirms this approach:

> If `lpJobAttributes` is `NULL`, the job object gets a default security descriptor and the handle cannot be inherited.

`Configuration.createJobObject()` is the single point of creation and enforces this behavior.

### Termination

`Execution.terminate(withExitCode:)` now also accepts `toProcessGroup: Bool`, defaulting to `false`. When `true`, it calls `TerminateJobObject` against the subprocess's Job Object, terminating the child and all its descendants. When `false`, it retains its existing per-process behavior using `TerminateProcess`. This mirrors the shape of the Unix `send(signal:toProcessGroup:)` API for cross-platform consistency.

The Windows branch of the `kill` case in `Execution.runTeardownSequence()` now passes the existing `killProcessGroup` value through to `terminate(withExitCode:toProcessGroup:)`.

`ProcessIdentifier` now also stores the Job Object handle alongside the process and thread handles, and closes it as part of `close()`.

### Nested job behavior

Subprocess assigns each spawned child to its own Job Object regardless of whether the parent process itself is in one. [The Nested Jobs documentation](https://learn.microsoft.com/en-us/windows/win32/procthread/nested-jobs) explains that on Windows 8+:

> If a process that is already in a job is added to another job, the jobs are nested by default if the system can form a valid job hierarchy and neither job sets UI limits.

If those conditions are not met, `AssignProcessToJobObject` fails and Subprocess throws a `spawnFailed` error identifying the parent's job as the likely cause. See the "Future directions" section for a planned opt-in fallback.

## Future directions

The following may be natural follow-up PRs, but are worth their own discussion. I could open subsequent PRs for any or all of these.

- `KILL_ON_JOB_CLOSE`: A flag that automatically terminates everything in a Job Object when the last handle to it closes. This would offer orphan protection if the parent crashes, but changes the current behavior of parent-exit not killing the child.
- `JOB_OBJECT_LIMIT_BREAKAWAY_OK`: A flag that enables descendants to escape the job. This changes the current behavior where every descendant of the spawned process stays in the job.
- User opt-in fallback for environments where job assignment fails (see "Nested job behavior" above): A follow-up PR could add an opt-in fallback that proceeds with descendant-tracking disabled, accepting that `toProcessGroup: true` and process-group teardown will be unavailable for that subprocess.

## Testing

The full test suite passes on `windows-latest` using GitHub Actions.

### DIFF
--- a/Tests/SubprocessTests/WindowsTests.swift
+++ b/Tests/SubprocessTests/WindowsTests.swift
@@ -14,6 +14,7 @@
 @preconcurrency import WinSDK
 import Foundation
 import Testing
+import Synchronization
 
 #if canImport(System)
 import System
@@ -278,7 +279,9 @@ extension SubprocessWindowsTests {
         #expect(stuckProcess.terminationStatus.isSuccess)
     }
 
-    /// Tests a use case for Windows platform handles by assigning the newly created process to a Job Object
+    /// Tests a use case for Windows platform handles by assigning the newly
+    /// created process to a user-supplied Job Object, nested inside
+    /// Subprocess's internal job.
     /// - see: https://devblogs.microsoft.com/oldnewthing/20131209-00/
     @Test func testPlatformHandles() async throws {
         let hJob = CreateJobObjectW(nil, nil)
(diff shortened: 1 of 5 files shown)


Tests/SubprocessTests/WindowsTests.swift
```diff
@@ -286,28 +288,129 @@ extension SubprocessWindowsTests {
         info.BasicLimitInformation.LimitFlags = DWORD(JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE)
         #expect(SetInformationJobObject(hJob, JobObjectExtendedLimitInformation, &info, DWORD(MemoryLayout<JOBOBJECT_EXTENDED_LIMIT_INFORMATION>.size)))
 
-        var platformOptions = PlatformOptions()
-        platformOptions.preSpawnProcessConfigurator = { (createProcessFlags, startupInfo) in
-            createProcessFlags |= DWORD(CREATE_SUSPENDED)
```
jakepetroules: This and the ResumeThread were deliberate illustrative parts of this test; please retain them

Tests/SubprocessTests/WindowsTests.swift
```diff
@@ -407,6 +510,88 @@ extension SubprocessWindowsTests {
...
+    /// Returns `true` if `WaitForSingleObject` on the handle reports the
+    /// process as still running.
+    private static func isProcessAlive(handle: HANDLE) -> Bool {
+        return WaitForSingleObject(handle, 0) == WAIT_TIMEOUT
+    }
+
+    /// Polls until the process referenced by the given handle exits or the
+    /// given timeout elapses, and returns `true` if the process exited
+    /// within the timeout.
+    private static func waitForProcessExit(
```
jakepetroules, line 609: Why poll? See monitorProcessTermination for how to do this

Tests/SubprocessTests/WindowsTests.swift
```diff
@@ -286,28 +288,129 @@ extension SubprocessWindowsTests {
...
-            createProcessFlags |= DWORD(CREATE_SUSPENDED)
-        }
-
         let result = try await Subprocess.run(
             self.cmdExe,
             arguments: ["/c", "echo"],
-            platformOptions: platformOptions,
             output: .discarded
         ) { execution, _ in
+            // Nest the child inside a user-supplied Job Object. Subprocess
```
jakepetroules, line 300: now the process may have been resumed before the jobs have been nested, this might be undesirable?

how about this: instead of throwing if the user removes CREATE_SUSPENDED, treat the user adding CREATE_SUSPENDED in preSpawnProcessConfigurator as a signal that they are going to manually handle calling ResumeThread. and so the presence of CREATE_SUSPENDED means the library _won't_ internally call ResumeThread. but it will always internally set CREATE_SUSPENDED on the flags passed to the actual CreateProcess call, regardless of what preSpawnProcessConfigurator does.

jakepetroules:
This looks great!

One thing: I am curious if there would be a use case for an API to prevent Subprocess from internally assigning processes to jobs. I don't know if there's any potential caveat or downside to doing this unconditionally. @compnerd ?

Tests/SubprocessTests/WindowsTests.swift
broken-circle: @jakepetroules Done! `generateCreateProcessFlag()` no longer sets `CREATE_SUSPENDED`. Both spawn paths now check whether the configurator set it, and pass that signal through to `assignChildToJobObjectAndResume()` so the thread is left suspended for the caller to resume. Subprocess still always sets `CREATE_SUSPENDED` before calling `CreateProcessW`, so the Job Object assignment remains atomic. The `preSpawnProcessConfigurator` docs now also explain this.

Tests/SubprocessTests/WindowsTests.swift
broken-circle: Done, switched to `RegisterWaitForSingleObject` with `withTaskCancellationHandler()`.

broken-circle:
@jakepetroules One potential downside of doing this unconditionally is failure in environments that don't support nesting. Jobs nest by default on Windows 8+, but assignment can still fail when either job has UI limits or the system can't form a valid hierarchy (the docs are vague about what qualifies).

An opt-out could tie into the third future direction. A `PlatformOptions.jobObjectAssignment` property with three behaviors could cover this case:

```swift
extension PlatformOptions {
    /// Describes how Job Object assignment is handled.
    public struct JobObjectAssignment: Sendable, Hashable {
        internal enum Storage: Sendable, Hashable {
            case always
            case bestEffort
            case never
        }

        internal let storage: Storage

        /// Always assign the child to a Job Object, and throw 
        /// `spawnFailed` if assignment fails.
        /// 
        /// This is the default and matches the current behavior.
        public static let always: Self = .init(storage: .always)

        /// Attempt to assign the child to a Job Object, and proceed
        /// without one if assignment fails.
        ///
        /// Descendant tracking and `toProcessGroup: true` are 
        /// unavailable for subprocesses for which assignment fails.
        public static let bestEffort: Self = .init(storage: .bestEffort)

        /// Never assign the child to a Job Object.
        ///
        /// For users that are manually handling job management, or 
        /// who do not need descendant-group teardown.
        public static let never: Self = .init(storage: .never)
    }
}
```

This would raise one question: If the spawned process has no assigned Job Object (either through `never` or through `bestEffort` + assignment failure), what should happen when a teardown step or `terminate()` call specifies `toProcessGroup: true`? Throwing an error is unambiguous, but adds mode dependency. Silently degrading to per-process behavior loses information and could be misleading. My preference would be to throw an error.

I could open a follow-up PR for this once we've settled on the shape. Does this match what you had in mind?

broken-circle:
@iCharlesHu @jakepetroules Rebased onto `main` following your recent changes. Ready for another look when you have a chance. I'm planning to open a follow-up for `PlatformOptions.jobObjectAssignment` to provide the opt-out we discussed.

Tests/SubprocessTests/WindowsTests.swift
```diff
@@ -411,6 +536,215 @@ extension SubprocessWindowsTests {
...
+    /// Asynchronously waits for the given handle to be signaled, using
+    /// `RegisterWaitForSingleObject`. Honors task cancellation by
+    /// unregistering the wait synchronously.
+    private static func waitForHandleSignal(handle: HANDLE) async {
+        // Shared mutable state between the registration scope, the C
+        // callback, and the cancellation handler. Access is serialized by
+        // a Win32 critical section since the callback runs on a thread-pool
+        // thread outside Swift Concurrency.
+        final class State: @unchecked Sendable {
+            var lock = SRWLOCK()
```
jakepetroules: nit: why SRWLOCK + unchecked Sendable? I think you can use Mutex here, right?

jakepetroules (approved):
Looks reasonable to me, thanks!

Tests/SubprocessTests/WindowsTests.swift
broken-circle: Took a couple attempts to get it working with the region checker, but landed on the `AsyncStream` + `Mutex` pattern from `AsyncIO+KQueue.swift`. Now we can drop `@unchecked Sendable` from the `State` class and scope it to just the struct holding the pointers.

jakepetroules:
@compnerd Any feedback from you on this?

broken-circle:
@iCharlesHu Would appreciate a review when you have a moment.

iCharlesHu (approved):
Looks great! Thanks for this change!

iCharlesHu:
@broken-circle  Could you rebase to pick up https://github.com/swiftlang/swift-subprocess/commit/84d6efca66b5461eecc67da58549bba79c950bcd ? We need the CI changes from there otherwise macOS CI will fail.

broken-circle:
@iCharlesHu Sure thing, done.

Issue #249, opened 2026-04-27
API issue: When streaming just `stderr`, then the user closure isn't allowed a return type


weissi:
**Describe the bug**

I wanted to write this function (lightly edited for clarity)

```
func withToolInvocation<R: Sendable>(
    _ executable: FilePath,
    _ arguments: [String],
    _ body: (ToolProgressStream) async throws -> R
) async throws -> R {
    let executionResult = try await Subprocess.run(
        .path(.init(executable.string)),
        arguments: Arguments(arguments),
        input: .none,
        output: .string(limit: 1024 * 1024, encoding: UTF8.self)
    ) { execution, standardError in
        return try await body(
            standardError.lines()  // in the real world I'm JSON decoding each line here
        )
    }

    return executionResult.value
}
```

The intent is that the user's `body` closure will return some value as is custom with `with*`-style functions such as `Subprocess.run`.

But I'm getting a compiler error:

```
Global function 'run(_:arguments:environment:workingDirectory:platformOptions:input:output:preferredBufferSize:isolation:body:)' requires the types 'StringOutput<UTF8>.OutputType' (aka 'Optional<String>') and '()' be equivalent
```

because it carries a `Output.OutputType == Void`:

```
public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
    _ executable: Executable,
    arguments: Arguments = [],
    environment: Environment = .inherit,
    workingDirectory: FilePath? = nil,
    platformOptions: PlatformOptions = PlatformOptions(),
    input: Input = .none,
    output: Output,
    preferredBufferSize: Int? = nil,
    isolation: isolated (any Actor)? = #isolation,
    body: (
        _ execution: Execution,
        _ errorSequence: AsyncBufferSequence
    ) async throws -> Result
) async throws -> ExecutionOutcome<Result> where Output.OutputType == Void {
```

**To Reproduce**
- Paste the above `withToolInvocation` function

**Expected behavior**
compiles

iCharlesHu:
This is the intended design. If you choose to use the closure-based `run()`, you are essentially saying, "I want to gather the output myself". This is why `ExecutionOutcome` only has a `value` field that represents whatever you decide to gather from the closure, instead of separate `.standardOutput` and `.standardError`. If you care about both standard output and standard error values, you have to gather both yourself with the closure approach.

weissi:
> This is the intended design. If you choose to use the closure-based `run()`, you are essentially saying, "I want to gather the output myself".

[...]

Issue #247, opened 2026-04-27
API problem: Number of closure parameters use to distinguish streaming modes


weissi:
**Describe the bug**

The `run` function is really a function that's often called `with*` and in accordance with this pattern it expectedly takes a closure. In Subprocess, we take this closure if we want to stream input and/or output.

For example

```
    let executionResult = try await Subprocess.run(
        .path(.init(executable.string)),
        arguments: Arguments(arguments)
    ) { execution, inputWriter, standardOutput, standardError in
       [...]
    }
```

In the real world, we have a plethora of different cases: We don't _always_ want to stream stdin, stdout and stderr. Sometimes we want to stream stdin & stdout but not stderr. In other cases, we want to stream stdout & stderr but configure stdin to be empty (on UNIX this means mapping /dev/null).

This leads to an API design issue because subprocess overloads the `run(...)` function on the _trailing closure_!

- `run(exe, arguments: args) { e in [...] }` means:
   - no input
   - stdout discarded
   - stderr discarded
- `run(exe, arguments: args) { e, s in [...] }` means:
   - no input
   - *stdout streamed* (in stream `s`)
   - stderr discarded
- `run(exe, arguments: args) { e, s1, s2 in [...] }` means:
   - *stdin streamed* (in `s1`)
   - *stdout streamed* (in stream `s2`)
   - stderr discarded!?
- `run(exe, arguments: args) { e, s1, s2, s3 in [...] }` means:
   - *stdin streamed* (in `s1`)
   - *stdout streamed* (in `s2`)
   - *stderr streamed* (in `s3`)

These can change if you pass `input:`, `output:`, `error:` parameters to the `run()` function.

To me this all feels a little arbitrary that essentially the number of arguments I pass configure what I stream. Especially with two streams passed, why is it `stdin` & `stdout` (discarding `stderr`) and not `stdout` & `stderr`? I think this is also what lead to #246

PR #262, opened 2026-05-13
Collapse run() overloads behind a generic Execution type

iCharlesHu:
Reduce the 14+ run() overloads to two (one executable-based, one configuration-based) by making `Execution` generic over its input, output, and error types. The body closure now takes a single `Execution<Input, Output, Error>`, and type-conditional extensions expose `standardInputWriter`, `standardOutput`, and `standardError` only for the matching stream types.

Companion changes:

- Merge `ExecutionOutcome` into `ExecutionResult<ClosureResult, Output, Error>`, which now carries the body closure's return value via `closureOutput`. Now all `run()` methods return `ExecutionResult`. `ExecutionOutcome` remains as an internal helper.

- Expose `CustomWriteInput` and `SequenceOutput` publicly, along with the `.inputWriter` and `.sequence` factories. Callers opt in per stream:
    - `input: .inputWriter` unlocks `execution.standardInputWriter`;
    - `output: .sequence` unlocks `execution.standardOutput`;
    - `error: .sequence` unlocks `execution.standardError`.

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -19,6 +19,7 @@ public import SystemPackage
 
 /// Runs an executable asynchronously and returns the collected output
 /// of the child process.
+///
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -28,7 +29,7 @@ public import SystemPackage
 ///   - input: The input to send to the executable.
 ///   - output: The method to use for redirecting standard output.
 ///   - error: The method to use for redirecting standard error.
-/// - Returns: An ``ExecutionRecord`` that contains the result of the run.
+/// - Returns: An ``ExecutionResult`` that contains the result of the run.
 public func run<
     Input: InputProtocol,
     Output: OutputProtocol,
@@ -42,7 +43,7 @@ public func run<
     input: Input = .none,
     output: Output,
     error: Error = .discarded
-) async throws -> ExecutionRecord<Output, Error> {
+) async throws -> ExecutionResult<Void, Output, Error> {
     let configuration = Configuration(
         executable: executable,
         arguments: arguments,
@@ -60,6 +61,7 @@ public func run<
 
 /// Runs an executable asynchronously and returns the collected output
 /// of the child process.
+///
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -69,7 +71,7 @@ public func run<
 ///   - input: A span to write to the subprocess's standard input.
 ///   - output: The method to use for redirecting standard output.
 ///   - error: The method to use for redirecting standard error.
-/// - Returns: An ``ExecutionRecord`` that contains the result of the run.
+/// - Returns: An ``ExecutionResult`` that contains the result of the run.
 public func run<
     InputElement: BitwiseCopyable,
     Output: OutputProtocol,
@@ -83,7 +85,7 @@ public func run<
     input: borrowing Span<InputElement>,
     output: Output,
     error: Error = .discarded
-) async throws -> ExecutionRecord<Output, Error> {
+) async throws -> ExecutionResult<Void, Output, Error> {
     let configuration = Configuration(
         executable: executable,
         arguments: arguments,
@@ -101,8 +103,15 @@ public func run<
 
 // MARK: - Custom Execution Body
 
-/// Runs an executable with given parameters and a custom closure
-/// to manage the running subprocess’s lifetime.
+/// Runs an executable asynchronously and lets a closure manage the running subprocess.
+///
+/// Use this overload when you need to interact with the subprocess while it runs,
+/// such as streaming its standard output, writing to its standard input, or sending
+/// signals. The closure runs concurrently with the subprocess and receives an
+/// ``Execution`` value you can use to access these capabilities.
+///
+/// The subprocess must terminate before this method returns.
+///
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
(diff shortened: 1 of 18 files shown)


iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/249
Resolves: https://github.com/swiftlang/swift-subprocess/issues/247

Sources/Subprocess/API.swift
```diff
@@ -233,259 +155,53 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
...
-/// - Returns: An ``ExecutionRecord`` that contains the result of the run.
+/// - Returns: An ``ExecutionResult`` that contains the result of the run.
 public func run<
     InputElement: BitwiseCopyable,
     Output: OutputProtocol,
     Error: ErrorOutputProtocol
 >(
     _ configuration: Configuration,
-    input: borrowing Span<InputElement>,
+    input: Span<InputElement>,
```
jakepetroules: it seems like this _should_ be `borrowing`?

jakepetroules:
This looks really nice, I think it is a big improvement! Summarizing the 6 overloads, there are 3 axes of variation with 2 possibilities each.

| First arg | Input type | Has body: closure? |
|---|---|---|
| Executable | Span | No |
| Executable | InputProtocol | No |
| Executable | InputProtocol | Yes |
| Configuration | Span | No |
| Configuration | InputProtocol | No |
| Configuration | InputProtocol | Yes |

This seems like a fairly reasonable number to me, especially as the overload mechanism no longer relies on the closure arguments.

The only thing I can think of to reduce the overloads further is that you could eliminate the 3 "executable" overloads and have _only_ the "configuration" ones, then move run() from a global to a member of Configuration, which would also address one of the concerns @weissi pointed out w.r.t. global namespace membership. That said, I think that discussion could be had separately from this improvement if that's something you are willing to consider.

Sources/Subprocess/Execution.swift
```diff
@@ -29,16 +29,62 @@ import Musl
...
+        outputStream: AsyncBufferSequence?,
+        errorStream: AsyncBufferSequence?
     ) {
         self.processIdentifier = processIdentifier
+        self.inputWriter = inputWriter
+        self.outputStream = outputStream
+        self.errorStream = errorStream
+    }
+}
+
```
jakepetroules, line 64: You can add an additional extension here:

```swift
extension Execution {
    @available(*, unavailable, message: "this property requires that the input is .standardInput")
    public var standardInputWriter: StandardInputWriter {
        fatalError()
    }

    @available(*, unavailable, message: "this property requires that the output is .sequence")
    public var standardOutput: AsyncBufferSequence {
        fatalError()
    }

    @available(*, unavailable, message: "this property requires that the error is .sequence")
    public var standardError: AsyncBufferSequence {
        fatalError()
    }
}
```

This will hugely improve the error message in certain invalid scenarios. For example, this is invalid code:

```swift
try await Subprocess.run(.name("foo"), input: .string(""), output: .currentStandardError, error: .discarded) { e in
    try await e.standardInputWriter.finish()
}
```

Before adding the proposed code, you get this message on the _input_ parameter:

> Static method 'string' requires the types 'CustomWriteInput' and 'StringInput<String, Unicode.UTF8>' be equivalent

After adding the proposed code, you get this message on the _standardInputWriter_ property:

> 'standardInputWriter' is unavailable: this property requires that the input is .standardInput

Much clearer and more localized to the issue!

jakepetroules:
I also get compiler warnings about AsyncSequence and nonisolated on Input+Foundation.swift on the write() functions at lines 78 and 132. You should be able to fix this by changing:

```swift
for try await data in asyncSequence {
```

to:

```swift
var iterator = self.sequence.makeAsyncIterator()
while let chunk = try await iterator.next(isolation: #isolation) {
```

but that requires a macOS 15 deployment target. FWIW, I think we could realistically consider raising Subprocess's deployment target to macOS 15 at this point.

Sources/Subprocess/Result.swift
```diff
@@ -17,90 +17,148 @@ import SystemPackage
...
+    }
+
+    /// The collected standard error of the subprocess.
+    public var standardError: Error.OutputType {
+        return self.capturedError!
+    }
+}
+
+extension ExecutionResult where ClosureResult == Void {
+    @available(*, unavailable)
```
jakepetroules: This will produce a better diagnostic for users:

```suggestion
    @available(*, unavailable, message: "closureOutput is not available when the closure returns Void")
```

The fatalError I would just leave empty -- it can't ever be called

Sources/Subprocess/Execution.swift
iCharlesHu: Ahh great point. Thanks!!

Sources/Subprocess/Result.swift
```diff
@@ -17,90 +17,148 @@ import SystemPackage
...
+        return self.capturedOutput!
+    }
+
+    /// The collected standard error of the subprocess.
+    public var standardError: Error.OutputType {
+        return self.capturedError!
+    }
+}
+
+extension ExecutionResult where ClosureResult == Void {
```
jakepetroules: I think this particular extension should be removed; there's no reason to make the Void case unavailable and it just makes the code less generic and requires downstream special cases when other generic functions wrap Subprocess.run.

jakepetroules: In fact, this never gets called at all because overload resolution isn't going to ever select it. Same for the other two.

I tried and I don't think @_specialize can work here either.

Honestly, I would just ensure that closureResult, capturedOutput, and capturedError are always non-nil, and just pass the void value `()` in cases where they aren't expected to be present. Then you'll get void properties that do nothing and won't crash if they're called.

Slightly janky but better than crashing with a force unwrap because technically there's nothing _wrong_ about calling these APIs with a Void type, it just isn't _useful_.

jakepetroules (changes requested):

Sources/Subprocess/Result.swift
iCharlesHu: Ahh good call. Actually that's what we went with originally.

iCharlesHu:
Make `ExecutionResult.{standardOutput|standardError|closureOutput}` non-conditional and non-optional. For `Output` with `Void OutputType` these properties will just be `Void`, which Xcode correctly warns you about.

warning

jakepetroules (approved):

PR #264, opened 2026-05-14
Executable.resolveExecutablePath should be async

### LINKED ISSUE #263: Executable.resolveExecutablePath should be async
This method requires hitting the filesystem and should therefore be `async` as it can block - I think we should resolve this before 1.0 since it's not possible to change later.

We might also want to provide an overload that allows the caller to provide their own `exists` function.

jakepetroules:
This method requires hitting the filesystem and should therefore be async as it can block

Closes #263


iCharlesHu (approved):

iCharlesHu:
Make sense. Thanks!

Issue #157, opened 2025-08-21
testDoesNotReapUnrelatedChildProcess nondeterministically hangs


jakepetroules:
Tests should probably add more timeouts in order to help debug what's actually going wrong. Could be the termination monitoring or the waitid perhaps?

iCharlesHu:
I was looking to see if `swift-testing` supports timeout feature but it seems like it doesn't yet. We'll need to implement our own timeout..

iCharlesHu:
So far seems to only happen on Linux.

jakepetroules:
> I was looking to see if `swift-testing` supports timeout feature but it seems like it doesn't yet. We'll need to implement our own timeout..

https://github.com/swiftlang/swift-build/blob/main/Sources/SWBTestSupport/Timeout.swift

broken-circle:
I wasn't able to reproduce this in 5,000 iterations on `ubuntu-latest` (Swift 6.2 release), and process monitoring has been substantially rewritten since this was filed. I think we can close this as fixed.

On `main`, only the specific child being awaited is reaped, and nothing does a process-wide `waitpid(-1)`, so awaiting one child can't reap or disturb an unrelated one.

PR #267, opened 2026-05-17
Add timeout to `testDoesNotReapUnrelatedChildProcess()`

broken-circle:
This PR adds a timeout to prevent non-deterministically hangs which consume the full CI job timeout with no actionable signal.

Addresses #157, which should remain open until the reason for the underlying hang is determined. I've ran this test 5,000+ times locally on macOS and haven't been able to reproduce the hang. Unfortunately I can't run it at that scale on Linux.

### DIFF
--- a/Tests/SubprocessTests/ProcessMonitoringTests.swift
+++ b/Tests/SubprocessTests/ProcessMonitoringTests.swift
@@ -229,7 +229,8 @@ extension SubprocessProcessMonitoringTests {
         }
     }
 
-    @Test func testDoesNotReapUnrelatedChildProcess() async throws {
+    @Test(.timeLimit(.minutes(1)))
+    func testDoesNotReapUnrelatedChildProcess() async throws {
         // Make sure we don't reap child exit status that we didn't spawn
         let child1 = self.immediateExitProcess(withExitCode: 0)
         let child2 = self.immediateExitProcess(withExitCode: 0)


jakepetroules:
I actually don't remember why I suggested using the Swift Build code as Swift Testing has had the testTimeout trait for well over a year. @iCharlesHu do you remember what the issue was here?

broken-circle:
Yeah, `.timeLimit(.minutes(1))` is simpler. I can update the PR to use that and drop the helper. One thing worth noting is the [coarse granularity](https://developer.apple.com/documentation/testing/trait/timelimit(_:)#discussion):

> Test timeouts do not support high-precision, arbitrarily short durations due to variability in testing environments. You express the duration in minutes, with a minimum duration of one minute.

The 1 minute timeout is fine for this test, but if finer-grained timeouts come up later, the helper would be more useful. I'd say we can drop it now and come back to it later if we need it.

broken-circle:
@iCharlesHu @jakepetroules I was thinking that since we now have a few tests that use Swift Testing time limits, it would probably be best to use the same feature here. If that sounds good, I can update the PR.

jakepetroules:
Makes sense

broken-circle:
Great, this is rebased and ready to go.

jakepetroules (approved):

iCharlesHu (approved):

PR #268, opened 2026-05-18
Make `Configuration` introspectable

### LINKED ISSUE #206: [Feature] Make Configuration, Arguments, and Environment more open for testing purposes
**Is your feature request related to a problem? Please describe.**

One of the benefits of generating Subprocess Configuration instead of statically constructing run(_:Executable,arguments:Arguments) is that you can make command-line generators in Swift code that can also be unit tested very nicely without actually spawning processes. The swiftly project makes use of this extensively because there are dozens of different kinds of command-lines that it uses. Actually spawning the processes for each test makes them slower and more brittle.

Inspecting the executable of the configuration is straightforward since the Executable enum is public and the cases can be matched easily. However, the Arguments is entirely opaque to any client module. There's no way to easily iterate over the argument list to verify correctness since the storage is internal. The best that one can use is the description, but it's more difficult to discern argument boundaries and there is some custom text inserted. Being able to iterate the arguments in some way, or even convert it into an array of strings would allow one to check what was provided to the init afterwards for verification purposes.

Similarly, Environment cannot be switched on its internal configuration to determine whether its custom, or inherited and what are the associated value dictionaries assigned to them. The description computed property for the CustomStringConvertible protocol is available, but then it becomes a parsing problem to find specific values in it.

Opening up the OS path key will be good too so that one can pull out any override or customization to the path environment.

**Describe the solution you'd like**
I would like for Arguments to be iterable as strings that were used to construct it for verification purposes. Also, I would like the ability to decompose Environment as the inherit or custom cases used to construct it along with the associated value dictionaries too. Also, the OS path key so that any custom or overrides to it can be detected. This will make it easier to build unit testing around configuration building for subprocess.

**Describe alternatives you've considered**
An alternative would be to construct my own configuration-like type that then converts to subprocess configuration later on. But, this would involve considerable duplication. Also, the conversion code would need to be tested on top of the configuration building logic.

broken-circle:
Closes #206.

This PR adds public introspection of `Arguments`, `Executable`, and `Environment` so callers generating subprocess configurations can unit-test them without spawning a subprocess. The current storage of these types is opaque, leaving callers with only string-based descriptions to parse.

- `Arguments` conforms to `RandomAccessCollection<Arguments.Value>`, where `Value` is a public mirror of the internal `StringOrRawBytes`. A `Value` enum was chosen over lossy UTF-8 decoding so callers constructing arguments from `[UInt8]` on POSIX can verify the exact bytes.

- `Arguments.executablePathOverride` is exposed as a separate public property rather than folded into iteration. Folding would make `argv[0]` ambiguous between an override and a regular first argument, and would prevent callers from distinguishing the two cases in tests. The internal stored property is renamed to `_executablePathOverride` to free the name for the public accessor.

- `Executable` and `Environment` expose their contents through new `Representation` mirror enums rather than exposing their internal storage types directly, decoupling the public introspection surface from internal storage. `Environment.Key.path` is promoted from `package` to `public` so callers can use it as a canonical reference to the `PATH` key.

The Swift tools version is bumped to 6.2.3 to support `@nonexhaustive` (SE-0487).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 6.2
+// swift-tools-version: 6.2.3
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -450,6 +450,33 @@ extension Executable: CustomStringConvertible, CustomDebugStringConvertible {
     }
 }
 
+// MARK: - Executable Introspection
+
+extension Executable {
+    /// The public representation of an executable's contents.
+    ///
+    /// Use this to introspect how an ``Executable`` was constructed (for
+    /// example, to verify in tests that a configuration builder produced the
+    /// expected executable reference without launching a subprocess).
+    @frozen
+    public enum Representation: Sendable, Hashable {
+        /// The executable is referenced by name and resolved against `PATH`.
+        case name(String)
+        /// The executable is referenced by an absolute or relative file path.
+        case path(FilePath)
+    }
+
+    /// The contents of this executable.
+    public var representation: Representation {
+        switch self.storage {
+        case .executable(let name):
+            return .name(name)
+        case .path(let path):
+            return .path(path)
+        }
+    }
+}
+
 // MARK: - Arguments
 
 /// A collection of arguments to pass to the subprocess.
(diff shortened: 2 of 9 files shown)


jakepetroules:
The capability is nice, but I'm a little concerned about the use of enums in public API -- usually that's discouraged. I'm not sure what other good options there are, though.

allevato:
Making it a `@nonexhaustive enum` would avoid the issues that tended to cause problems with evolution for enum-based APIs, I think?

broken-circle:
Nice, I was just wondering if something like `@nonexhaustive` existed. Reading [SE-0487](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0487-extensible-enums.md), it seems appropriate here.

We could annotate `Arguments.Value` and `Executable.Representation` with `@frozen`, since their cases are exhaustive. `Environment.Representation` can be `@nonexhaustive`.

iCharlesHu:
@broken-circle thanks for the PR! Let's do this post 1.0 (it was originally scheduled for post 1.0)

broken-circle:
This is ready for review now that we're post-1.0.

Earlier I thought `Package.swift` was fine as-is, but the tools version needed bumping to 6.2.3, since that's when `@nonexhaustive` was implemented. The CMake build doesn't declare a Swift floor; should we add a `CMAKE_Swift_COMPILER_VERSION` guard?

If the version bump isn't acceptable, we can make `Environment.Representation` `@frozen` instead, although this enum's set of members isn't necessarily closed. For example, there's currently no way to inherit the parent environment and override a value with raw bytes, and filling that gap would need a new case, and therefore a new major version.

Issue #245, opened 2026-04-27
Type clash 'AsyncBufferSequence' swift-subprocess vs. swift-async-algorithms


weissi:
**Describe the bug**

In almost any streaming use case, `swift-subprocess` will be used together with `swift-async-algorithms`.
For example, it's common to want

```
            merge( // from swift-async-algorithms
                standardOutput.lines(),
                standardError.lines()
            )
```

However, if you just import `AsyncAlgorithms` & `Subprocess`, you'll always get

```
error: 'AsyncBufferSequence' is ambiguous for type lookup in this context
```

**To Reproduce**

```
import AsyncAlgorithms
import Subprocess

typealias OutputStream = AsyncMerge2Sequence<
    AsyncBufferSequence.LineSequence<UTF8>,
    AsyncBufferSequence.LineSequence<UTF8>
>
```

```
error: 'AsyncBufferSequence' is ambiguous for type lookup in this context
```

The user can work around of course by qualifying the type names, e.g.

```
typealias OutputStream = AsyncMerge2Sequence<
    Subprocess.AsyncBufferSequence.LineSequence<UTF8>,
    Subprocess.AsyncBufferSequence.LineSequence<UTF8>
>
```

but this is cumbersome.

**Expected behavior**
Base packages should not publicly export types with the same name.

jakepetroules:
> In almost any streaming use case, `swift-subprocess` will be used together with `swift-async-algorithms`.

This is orthogonal to the issue you reported, but I think approach would be incorrect and is why I filed #59, which was implemented in #180.

jakepetroules:
I'm not sure it's fair to call this a _bug_, but I agree it's probably a good idea to call the type from this package `SubprocessBufferSequence` or similar.

PR #269, opened 2026-05-19
Use `Int64` and parameter labels for byte-denominated limits

### LINKED ISSUE #248: API issue: Unit (e.g. bytes/characters/code units) omitted when configuring limits/sizes
**Describe the bug**

Some APIs like `.string(...)`, `.bytes(...)` and `size` parameters take integers of units unknown to the programmer.

For example:

```
        output: .string(limit: 12345, encoding: UTF16.self)
```

What's the limit here? 12345 bytes, 12345 UTF16 code units, 12345 `String.Character`s? This is unnecessarily unclear. I would argue that the limit should be explicit. Compare this for example to NIOFileSystem's API:

```
        let plan = try await ByteBuffer(
            contentsOf: "/Users/hal9000/demise-of-dave.txt",
            maximumSizeAllowed: .mebibytes(1)
        )
```

which is much clearer than `maximumSizeAllowed: 1024 * 1024` or similar.

broken-circle:
Closes #248.

This PR adds a new public `ByteCount` type to make byte-denominated limits across the API unambiguous.

## Design

The shape is adapted from [swift-nio's `ByteCount`](https://github.com/apple/swift-nio/blob/6338e9fb5e136af8337b48eee65f8868d571a3d7/Sources/NIOFS/ByteCount.swift), except with no `.unlimited` static property. Per the discussion in #248, callers should make an explicit choice rather than opting into unbounded allocation.

`CustomStringConvertible` conformance produces human-readable descriptions such as `"2 MiB (2097152 bytes)"`, with the following behavior:

- Values smaller than 1 KiB render as plain bytes (for example, `"512 bytes"`).
- Values are truncated rather than rounded, so the displayed value never overstates the underlying count.
- IEC labels (KiB, MiB, GiB) are used over JEDEC labels (KB, MB, GB) for technical correctness.

## `BufferingPolicy.maxLineLength`

`BufferingPolicy.maxLineLength` is currently measured in `Encoding.CodeUnit`, not bytes. This PR labels the associated value as `maxLineLength(codeUnits: Int)` to make the unit explicit, and adds `SubprocessError.Context.codeUnitCount(count: Int, encodingName: String)` plus a matching `outputLimitExceeded(codeUnits:encoding:)` factory method so the error description renders the count alongside the encoding name.

Switching the buffer itself to byte-denominated limits would be a behavioral change. I can do that in a follow-up PR if there's interest.

## Testing

This PR adds a new test suite and updates existing tests to use the new type, without changing any existing semantics. For example, `.string(limit: .max)` changes to `.string(limit: .bytes(.max))`.

### DIFF
--- a/Sources/Subprocess/Error.swift
+++ b/Sources/Subprocess/Error.swift
@@ -111,7 +111,8 @@ extension SubprocessError {
 
     private enum Context: Sendable, Hashable {
         case string(String)
-        case int(Int)
+        case byteCount(Int64)
+        case codeUnitCount(count: Int, encodingName: String)
         case processControlOperation(ProcessControlOperation)
     }
 
@@ -178,13 +179,17 @@ extension SubprocessError: CustomStringConvertible, CustomDebugStringConvertible
                 return "Failed to write bytes to the child process."
             }
         case .outputLimitExceeded:
-            if let context = self.context[self.code],
-                case .int(let limit) = context
-            {
-                return "Child process output exceeded the limit of \(limit) bytes."
-            } else {
-                return "Child process output exceeded the limit."
+            if let context = self.context[self.code] {
+                switch context {
+                case .byteCount(let limit):
+                    return "Child process output exceeded the limit of \(limit) bytes."
+                case .codeUnitCount(let limit, let encodingName):
+                    return "Child process output exceeded the line length limit of \(limit) \(encodingName) code units."
+                default:
+                    break
+                }
             }
+            return "Child process output exceeded the limit."
         case .asyncIOFailed:
             let context = self.context[self.code]
             switch (self.underlyingError, context) {
@@ -286,12 +291,28 @@ extension SubprocessError {
         )
     }
 
-    internal static func outputLimitExceeded(limit: Int) -> Self {
+    internal static func outputLimitExceeded(byteLimit: Int64) -> Self {
         return SubprocessError(
             code: .outputLimitExceeded,
             underlyingError: nil,
             context: [
-                .outputLimitExceeded: .int(limit)
+                .outputLimitExceeded: .byteCount(byteLimit)
+            ]
+        )
+    }
+
+    internal static func outputLimitExceeded<Encoding: _UnicodeEncoding>(
+        codeUnits limit: Int,
+        encoding: Encoding.Type
+    ) -> Self {
+        return SubprocessError(
+            code: .outputLimitExceeded,
+            underlyingError: nil,
+            context: [
+                .outputLimitExceeded: .codeUnitCount(
+                    count: limit,
+                    encodingName: String(describing: encoding)
+                )
             ]
         )
     }
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -29,12 +29,12 @@ public protocol OutputProtocol: Sendable, ~Copyable {
     func output(from span: RawSpan) throws -> OutputType
 
     /// The maximum number of bytes to collect.
-    var maxSize: Int { get }
+    var maxSize: Int64 { get }
 }
 
 extension OutputProtocol {
     /// The maximum number of bytes to collect.
-    public var maxSize: Int { 128 * 1024 }
+    public var maxSize: Int64 { 128 * 1024 }
 }
 
 /// An output type that discards output from the child process.
@@ -102,7 +102,7 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOut
     /// The type for this output.
     public typealias OutputType = String?
     /// The maximum number of bytes to collect.
-    public let maxSize: Int
+    public let maxSize: Int64
 
     /// Creates a string from a raw span.
     public func output(from span: RawSpan) throws -> String? {
@@ -112,8 +112,8 @@ public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOut
         }
     }
 
-    internal init(limit: Int, encoding: Encoding.Type) {
-        self.maxSize = limit
+    internal init(byteLimit: Int64, encoding: Encoding.Type) {
+        self.maxSize = byteLimit
     }
 }
 
@@ -122,7 +122,7 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
     /// The output type for this output option.
     public typealias OutputType = [UInt8]
     /// The maximum number of bytes to collect.
-    public let maxSize: Int
+    public let maxSize: Int64
 
     internal func captureOutput(
         from diskIO: consuming IODescriptor,
@@ -130,18 +130,23 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
     ) async throws(SubprocessError) -> [UInt8] {
         var result: [UInt8] = []
         do {
-            var maxLength = self.maxSize
-            if maxLength != .max {
+            var maxLength: Int64 = self.maxSize
+            let isBounded = maxLength != .max
+            if isBounded {
                 // Read one extra byte to detect output that exceeds the
                 // limit.
                 maxLength += 1
+            }
+            // Clamp to `Int.max` on 32-bit platforms.
+            let cappedMaxLength = Int(exactly: maxLength) ?? Int.max
+            if isBounded {
                 // Reserve capacity to avoid reallocations.
-                result.reserveCapacity(maxLength)
+                result.reserveCapacity(cappedMaxLength)
             }
             let bufferSize = AsyncIO.queryPipeBufferSize(for: diskIO.descriptor())
 
-            while result.count < maxLength {
-                let remaining = maxLength - result.count
+            while result.count < cappedMaxLength {
+                let remaining = cappedMaxLength - result.count
                 guard
                     let chunk = try await AsyncIO.shared.read(
                         from: diskIO,
@@ -159,8 +164,8 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
         }
         try diskIO.safelyClose()
 
-        if result.count > self.maxSize {
-            throw .outputLimitExceeded(limit: self.maxSize)
+        if Int64(result.count) > self.maxSize {
+            throw .outputLimitExceeded(byteLimit: self.maxSize)
         }
         return result
     }
@@ -170,7 +175,7 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
         span.withUnsafeBytes { Array($0) }
     }
 
-    internal init(limit: Int) {
+    internal init(limit: Int64) {
         self.maxSize = limit
     }
 }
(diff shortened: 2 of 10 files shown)


Sources/Subprocess/Error.swift
```diff
@@ -111,7 +111,8 @@ extension SubprocessError {
 
     private enum Context: Sendable, Hashable {
         case string(String)
-        case int(Int)
+        case byteCount(ByteCount)
+        case codeUnitCount(Int)
```
jakepetroules: Should this also carry the encoding as a second parameter? a code unit count without the context of a specific encoding isn't very meaningful

broken-circle: Yeah, good call. Updated to carry the encoding name rendered with `String(describing:)`.

Sources/Subprocess/ByteCount.swift
```diff
@@ -0,0 +1,172 @@
...
+/// .bytes(limit: .kibibytes(64))
+/// ```
+///
+/// The decimal factories (``kilobytes(_:)``, ``megabytes(_:)``,
+/// ``gigabytes(_:)``) use powers of 1,000. The binary factories
+/// (``kibibytes(_:)``, ``mebibytes(_:)``, ``gibibytes(_:)``) use powers
+/// of 1,024.
+public struct ByteCount: Sendable, Hashable {
+    /// The number of bytes.
+    public let bytes: Int
```
weissi: This is one of the rare cases where I'd suggest to use `Int64`. Or else, on 32bits, 2GB becomes the largest expressible quantity

Sources/Subprocess/ByteCount.swift
```diff
@@ -0,0 +1,172 @@
...
+    /// One mebibyte is 1,048,576 bytes.
+    ///
+    /// - Parameter count: The number of mebibytes.
+    public static func mebibytes(_ count: Int) -> ByteCount {
+        return ByteCount(_bytes: 1_024 * 1_024 * count)
+    }
+
+    /// Returns a ``ByteCount`` representing the given number of gibibytes.
+    ///
+    /// One gibibyte is 1,073,741,824 bytes.
```
weissi: might be worth documenting it as 2^30 too because that's the difference to gigabyte

Sources/Subprocess/ByteCount.swift
```diff
@@ -0,0 +1,172 @@
...
+/// ```swift
+/// .string(limit: .mebibytes(1))
+/// .bytes(limit: .kibibytes(64))
+/// ```
+///
+/// The decimal factories (``kilobytes(_:)``, ``megabytes(_:)``,
+/// ``gigabytes(_:)``) use powers of 1,000. The binary factories
+/// (``kibibytes(_:)``, ``mebibytes(_:)``, ``gibibytes(_:)``) use powers
+/// of 1,024.
+public struct ByteCount: Sendable, Hashable {
```
weissi: Should we nest this type? This is likely gonna clash

Sources/Subprocess/ByteCount.swift
```diff
@@ -0,0 +1,172 @@
...
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+// Adapted from swift-nio.
```
weissi: IANAL but I believe that you can't just strip NIO's license header. You just put yours on top of NIO's similar to this: https://github.com/apple/swift-nio/blob/6338e9fb5e136af8337b48eee65f8868d571a3d7/Sources/NIOPosix/StructuredConcurrencyHelpers.swift#L1-L29

Sources/Subprocess/ByteCount.swift
```diff
@@ -0,0 +1,172 @@
...
+// https://github.com/apple/swift-nio/blob/6338e9fb5e136af8337b48eee65f8868d571a3d7/Sources/NIOFS/ByteCount.swift
+
+/// A measurement of memory in bytes.
```
weissi: Also CC @glbrntt who introduced NIO's `ByteCount` type. Anything you learned since then that you wish you had done differently?

Sources/Subprocess/ByteCount.swift
broken-circle: Good call. Updated to `Int64` and added a clamp pattern at the `Int`-consuming call sites (with documentation on `ByteCount.bytes` explaining the behavior).

Sources/Subprocess/ByteCount.swift
broken-circle: Worth it for the clarity. Added the power notation to all the factories for consistency.

Sources/Subprocess/ByteCount.swift
broken-circle: This is parallel to the discussion in #245 about possibly renaming `AsyncBufferSequence` to `SubprocessBufferSequence`. For `ByteCount`, a rename feels less appropriate, since `SubprocessByteCount` implies something Subprocess-specific about the type's behavior. Nesting works, but I can't find a semantically-natural parent type (`Configuration.ByteCount` is the closest, but still feels arbitrary). I'm leaning toward leaving it top-level and relying on module namespacing if clashes occur, but I'm open to either a rename or nesting.

Tagging @jakepetroules since he's involved in that discussion. Curious which direction you'd prefer here.

weissi: If we just did the IMHO obvious thing and nested the `run` methods into a type, this would be obvious too. It would be `Subprocess.BufferSequence` / `Subprocess.ByteCount` (and here `Subprocess` would be the type name and not the module). This will come up time and time again that we'd like a namespace but we're lacking one

Sources/Subprocess/Error.swift
jakepetroules: Can it be a `String.Encoding` instead of a ` String`?

broken-circle: I don't think there's a clean mapping between `_UnicodeEncoding` and `String.Encoding`, since custom `_UnicodeEncoding` conformances have no equivalent. We'd also need to change `Error.swift` to depend on `SubprocessFoundation`.

If we want to avoid the string-based shape, I was thinking of using a small `internal enum CodeUnitEncoding` with three UTF cases and one `other(String)` case. That covers the common cases with type safety and stable display names. Worth doing?

iCharlesHu:
I don't think `Subprocess` should vend its own `ByteCount`. I agree with the fact that having a concrete type is easier to use, but at the same time using `Int` (will be updated to `Int64`) as byte count is a very common pattern and used everywhere, and it'll be silly to define a `ByteCount` for every module that might need this concept.

Instead of defining our own, we should punt this until the standard library or swift-system exposes a canonical type for us to use.

broken-circle:
> Instead of defining our own, we should punt this until the standard library or swift-system exposes a canonical type for us to use.

Got it, I'll drop `ByteCount`. We'll keep the `Int64` and `BufferingPolicy.maxLineLength(codeUnits:)` changes.

While we're making these changes, do you want to also go with the other option proposed in #248 to rename the byte count label from `limit` to `byteLimit`? It would add some clarity to the call sites in the meantime. `OutputProtocol.bytes(limit:)` and `BytesOutput.init(limit:)` would stay as-is, since the factory/type names already disambiguate the unit.

Sources/Subprocess/ByteCount.swift
glbrntt: I'd probably have the APIs be generic over whatever the appropriate integer protocol is rather than taking `Int64` unconditionally.

broken-circle:
@iCharlesHu Just checking in on this. I have all three options ready to go:

1. Keep `ByteCount` (current PR)
2. Drop `ByteCount`, switch to `Int64`
3. Drop `ByteCount`, switch to `Int64`, rename `limit` to `byteLimit` on byte-denominated APIs

Ready to push whichever of the latter you prefer.

Edit: Pushed these as separate commits so I can rebase #260. Easy to drop either depending on the direction we go with. Will have to rebase this one too once we're ready.

jakepetroules:
I favor Option 3. @iCharlesHu?

iCharlesHu:
Agreed. Let's go with Option 3. Thanks @broken-circle!

Sources/Subprocess/SubprocessOutputSequence.swift
```diff
@@ -501,10 +501,11 @@ extension AsyncBufferSequence.StringSequence {
     public enum BufferingPolicy: Sendable {
         /// Adds to the buffer without imposing a limit on line length.
         case unbounded
-        /// Imposes a maximum line length limit.
+        /// Imposes a maximum line length limit, measured in code units of
+        /// the sequence's `Encoding`.
         ///
         /// The subprocess throws an error if a line exceeds this limit.
-        case maxLineLength(Int)
+        case maxLineLength(codeUnits: Int)
```
iCharlesHu, line 513: Should we change this one to Int64 as well?

broken-circle: `Int` is right here since `maxLineLength` is compared against `buffer.count` (where `buffer` is `[Encoding.CodeUnit]`). Since `Array.count` is always `Int`, the buffer can't hold more than `Int.max` code units, and the `Int64` range wouldn't be reachable.

broken-circle:
Great, thanks! I went ahead and rebased this onto `main` and squashed the commits.

broken-circle:
Looks like CI failed on Android (Linux) due to the same race condition I fixed for Darwin in #276:

> ✘ Test testSuspendResumeProcess() recorded an issue at LinuxTests.swift:42:6: Caught error: Failed to monitor the child process exit status with underlying error: No child processes

I'll update that PR to also fix the issue on Linux, then rebase this to include those changes, if we can merge #276.

iCharlesHu:
@jakepetroules are we ignoring all Android build failures for now?

jakepetroules (dismissed):
I think we can ignore those Android failures for now, yeah. At least one of them passed, which gives the signal we need.

Sources/Subprocess/IO/Output.swift
```diff
@@ -122,26 +122,31 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
...
-            if maxLength != .max {
+            var maxLength: Int64 = self.maxSize
+            let isBounded = maxLength != .max
+            if isBounded {
                 // Read one extra byte to detect output that exceeds the
                 // limit.
                 maxLength += 1
+            }
+            // Clamp to `Int.max` on 32-bit platforms.
+            let cappedMaxLength = Int(exactly: maxLength) ?? Int.max
```
iCharlesHu, line 141: Sorry, I think I'll need to think over this one a bit more. Looking at the code here, I realized that since we need to clamp max length on 32-bit platforms, offering `Int64` as `maxLength` doesn't change anything. You are still capped at most ` Int32.max` bytes. In addition, now we silently truncate `maxLength`, so you might get unexpected results.

Maybe we don't need to explicitly use `Int64` after all. Let me think more about this.

Sources/Subprocess/IO/Output.swift
```diff
@@ -122,26 +122,31 @@ public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
     /// The output type for this output option.
     public typealias OutputType = [UInt8]
     /// The maximum number of bytes to collect.
-    public let maxSize: Int
+    public let maxSize: Int64
 
     internal func captureOutput(
         from diskIO: consuming IODescriptor,
         for processIdentifier: ProcessIdentifier
     ) async throws(SubprocessError) -> [UInt8] {
```
iCharlesHu, line 130: Also since `Array.count` is `Int` there is no way for us to return an array with more than Int32.max elements on 32-bit platforms.

broken-circle: Yeah, good catch. And on 64-bit platforms, `Int` is `Int64`, so there's no constraint to avoid. I can switch back to `Int`, drop the clamp, and keep the `byteLimit` labels.

broken-circle:
Rebased this onto `main` following the recent Android/Unix test changes, so the next CI run should be all green.

iCharlesHu:
@broken-circle  Thanks so much for this PR!

I took this to my team (Foundation) for discussion, and after weighing the trade-offs we've decided to not use `ByteCount` or `Int64` for now:

- ByteCount is very nice, but it shouldn't be vended by Subprocess. Foundation already uses `Int` as "length" throughout its API surface so it's not unprecedented.
- Int64 as a "max length" really only matters when writing a large number of bytes directly to a file descriptor. For our use case here, where we capture output as [UInt8], Int64 offers no advantage on 32-bit systems, so we'd prefer to stay with Int.

We should definitly consider switching to `ByteCount` once standard library or swift-system could vend a canical copy.

broken-circle:
@iCharlesHu Makes sense! Would you be open to a PR that just changes `limit` to `byteLimit` where applicable, or should we leave it as-is for now?

iCharlesHu:
> @iCharlesHu Makes sense! Would you be open to a PR that just changes `limit` to `byteLimit` where applicable, or should we leave it as-is for now?

Let's keep it `limit` for now, since that spelling was already accepted in prior API reviews.

PR #270, opened 2026-05-20
Replace call to `CreatePipe()` with call to `_pipe()` on Windows.

grynspan:
Our call to `CreatePipe()` is effectively recreating the logic of `_pipe()`, but risks leaking handles at the various points of failure. Let's just use the UCRT implementation instead.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -1298,18 +1298,19 @@ extension FileDescriptor {
         writeEnd: FileDescriptor
     ) {
         var fds: [2 of CInt] = [-1, -1]
-        try fds.mutableSpan.withUnsafeMutableBufferPointer { fds in
+        var span = fds.mutableSpan
+        try span.withUnsafeMutableBufferPointer { fds throws(SubprocessError) in
             guard 0 == _pipe(fds.baseAddress!, 0, _O_BINARY) else {
                 throw SubprocessError.asyncIOFailed(
                     reason: "Failed to create pipe",
                     underlyingError: nil // FIXME: Errno(rawValue: _subprocess_windows_get_errno())
                 )
             }
-            return (
-                readEnd: FileDescriptor(rawValue: fds[0]),
-                writeEnd: FileDescriptor(rawValue: fds[1])
-            )
         }
+        return (
+            readEnd: FileDescriptor(rawValue: fds[0]),
+            writeEnd: FileDescriptor(rawValue: fds[1])
+        )
     }
 
     var platformDescriptor: PlatformFileDescriptor {
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -1298,16 +1298,18 @@ extension FileDescriptor {
         writeEnd: FileDescriptor
     ) {
         var fds: [2 of CInt] = [-1, -1]
-        guard 0 == _pipe(&fds, 0, _O_BINARY) else {
-            throw SubprocessError.asyncIOFailed(
-                reason: "Failed to create pipe",
-                underlyingError: nil // FIXME: Errno(rawValue: _subprocess_windows_get_errno())
+        try fds.span.withUnsafeBufferPointer { fds in
+            guard 0 == _pipe(fds.baseAddress!, 0, _O_BINARY) else {
+                throw SubprocessError.asyncIOFailed(
+                    reason: "Failed to create pipe",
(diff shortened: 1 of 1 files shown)


jakepetroules:
Discussed with @grynspan and this makes sense; I can't recall why the Win32 APIs were used directly in this case. Looks like there is a compilation error with errno though.

grynspan:
> C:\source\Sources\Subprocess\Platforms\Subprocess+Windows.swift:1315:38: error: cannot convert value of type 'Errno' to expected argument type 'SubprocessError.UnderlyingError' (aka 'SubprocessError.WindowsError')

This is wrong, some Windows API (UCRT/CRT stuff) sets `errno` rather the Win32-domain last error. We should allow for either. @compnerd Thoughts?

compnerd:
> > C:\source\Sources\Subprocess\Platforms\Subprocess+Windows.swift:1315:38: error: cannot convert value of type 'Errno' to expected argument type 'SubprocessError.UnderlyingError' (aka 'SubprocessError.WindowsError')
> 
> This is wrong, some Windows API (UCRT/CRT stuff) sets `errno` rather the Win32-domain last error. We should allow for either. @compnerd Thoughts?

Yes, we should allow both; I do this already for the canonical windows error type (in https://github.com/compnerd/swift-win32).

I'm not sure about this: `_pipe` and `CreatePipe` work in different domains (and this is generally a problem). Changes made with the C API sometimes do not reflect the state in win32 and vice-versa. e.g. `setenv` and `SetEnvironmentVariable` have observable differences.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -1308,37 +1308,18 @@ extension FileDescriptor {
...
-        guard CreatePipe(&readHandle, &writeHandle, &saAttributes, 0),
-            readHandle != INVALID_HANDLE_VALUE,
-            writeHandle != INVALID_HANDLE_VALUE,
-            let readHandle: HANDLE = readHandle,
-            let writeHandle: HANDLE = writeHandle
-        else {
-            throw SubprocessError.asyncIOFailed(
-                reason: "Filed to create pipe",
-                underlyingError: SubprocessError.WindowsError(rawValue: GetLastError())
+        try withUnsafeTemporaryAllocation(of: CInt.self, capacity: 2) { fds in
```
compnerd: Can we get away with `InlineArray<2, CInt>`?

grynspan: Depends how far back the package supports.

grynspan:
> I'm not sure about this: `_pipe` and `CreatePipe` work in different domains (and this is generally a problem).

The current implementation is just `_pipe()` with bugs. D.R.Y.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
grynspan: `InlineArray` is in 6.2, right?

jakepetroules: Looks like it requires macOS 26 -- too high.

iCharlesHu:
> This is wrong, some Windows API (UCRT/CRT stuff) sets `errno` rather the Win32-domain last error. We should allow for either. @compnerd Thoughts?

Ahh good point. Maybe `UnderlyingError` should be an enum of either `WindowsError` or `Errno`?

compnerd:
> > This is wrong, some Windows API (UCRT/CRT stuff) sets `errno` rather the Win32-domain last error. We should allow for either. @compnerd Thoughts?
> 
> Ahh good point. Maybe `UnderlyingError` should be an enum of either `WindowsError` or `Errno`?

It needs to be an enum, something like this (which I already use in the canonical error representation):

```swift
  public enum ErrorCode: Sendable {
    case nt(NTSTATUS)
    case win32(DWORD)
    case ucrt(errno_t)
    case hresult(HRESULT)
  }
```

iCharlesHu (approved):

iCharlesHu:
@grynspan could you rebase to pick up test config changes? That should unblock CI.

I'll update `WindowsError` in a separate PR.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
grynspan: Oops. Okay.

grynspan:
@iCharlesHu Looks like CI is still angry.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
grynspan: Oh wait this is Windows-only code anyway. :)

PR #272, opened 2026-05-21
Cancel AsyncIO when the child process exits

### LINKED ISSUE #256: stdin/out/err writes and reads must be cancellable such that we won't hang if we kill child but streams don't EOF
**Describe the bug**

@jakepetroules  tells me (and I've looked at the code which seems to confirm) that at the moment, Subprocess makes the flawed assumption that killing the child process will EOF the file descriptors inherited into the child.

This assumption is _not_ correct.

Consider the following scenario:

- Parent Process `P` uses `Subprocess` to spawn a child process `A` streaming its output
- `A` starts up and starts another child `B` (e.g. using `fork()`) inheriting over the file descriptor
- `A` never writes anything to stdout
- `B` sleeps for 1 hour and then writes `hello world` and exits
- `P` now wants to abort its subprocess so it uses cancellation
- Cancellation will now kill `A` even if `A` isn't responsive
- Subprocess however is still reading from the pipe inherited into `A` and is waiting for EOF
- EOF won't arrive for an hour because that file desriptor is now in `B` which is a process completely unknown to `P`

What needs to happen is what `AsyncProcess` does: The reading/writing needs to be cancelled too.

**To Reproduce**

Three (one for stdin, one for stdout and one for stderr)  test cases similar to

```
    func testWeDoNotHangIfStandardInputRemainsOpenButProcessExits() async throws {
        // This tests an odd situation: The child exits but stdin is still not closed, mostly happens if we inherit a
        // pipe that we still have another writer to.

        var sleepPidToKill: CInt?
        defer {
            if let sleepPidToKill {
                self.logger.debug(
                    "killing our sleep grand-child",
                    metadata: ["pid": "\(sleepPidToKill)"]
                )
                kill(sleepPidToKill, SIGKILL)
            } else {
                XCTFail("didn't find the pid of sleep to kill")
            }
        }
        do { // We create a scope here to make sure we can leave the scope without hanging
            let (stdinStream, stdinStreamProducer) = AsyncStream.makeStream(of: ByteBuffer.self)
            let exe = ProcessExecutor(
                executable: "/bin/sh",
                [
                    "-c",
                #"""
                # This construction attempts to emulate a simple `sleep 12345678 < /dev/null` but some shells (eg. dash)
                # won't allow stdin inheritance for background processes...
                exec 2>&- # close stderr
                exec 2<&0 # duplicate stdin into fd 2 (so we can inherit it into sleep

                ( # this creates a child of our child which gets inherited our fds
                    exec 0<&2  # map the duplicated fd 2 as our stdin
                    exec 2>&-  # close the duplicated fd2
                    exec sleep 12345678 # sleep (this will now have the origin stdin as its stdin)
                ) & # uber long sleep that will inherit our stdin pipe
                exec 2>&- # close duplicated 2

                read -r line
                echo "$line" # write back the line
                echo "$!" # write back the sleep
                exec >&-
                exit 0
                """#
                ],
                standardInput: stdinStream
            )
            stdinStreamProducer.yield(ByteBuffer(string: "GO\n"))
            stdinStreamProducer.yield(ByteBuffer(repeating: 0x42, count: 16 * 1024 * 1024))
            async let resultAsync = exe.runWithExtendedInfo()
            async let stderrAsync = Array(exe.standardError)
            var stdoutLines = await exe.standardOutput.splitIntoLines().makeAsyncIterator()
            let lineGo = try await stdoutLines.next()
            XCTAssertEqual(ByteBuffer(string: "GO"), lineGo)
            let linePid = try await stdoutLines.next().map(String.init(buffer:))
            let sleepPid = try XCTUnwrap(linePid.flatMap { CInt($0) })
            self.logger.debug("found our sleep grand-child", metadata: ["pid": "\(sleepPid)"])
            sleepPidToKill = sleepPid
            let stderrBytes = try await stderrAsync
            XCTAssertEqual([], stderrBytes)
            let result = try await resultAsync
            XCTAssertEqual(.exit(0), result.exitReason)
            XCTAssertNotNil(result.standardInputWriteError)
            XCTAssertEqual(ChannelError.ioOnClosedChannel, result.standardInputWriteError as? ChannelError)
            stdinStreamProducer.finish()
        }
    }
```

should do the trick

**Expected behavior**

When a Subprocess.run gets cancelled, it should return control back to the caller within a short and reasonable, user-controllable amount of time. It must cancel the read/writes that communicate on the file descriptors inherited down.

iCharlesHu:
Subprocess used to assume that a child's exit closes the pipe FDs it inherited. When the child forked a grandchild that also holds those FDs, the kernel keeps the pipes alive after the child exits and AsyncIO's read/write loops never see EOF or EPIPE — run() hangs.

Split process monitoring into two phases:
- waitForProcessTermination waits for the kernel to report exit.
- reapProcess consumes the zombie and returns the status.

Configuration.run now spawns a parallel monitor task and races it against the body closure via an AtomicCounter. If the process exits while the body is still doing I/O, the monitor calls AsyncIO.cancelAsyncIO(for:) to unblock every pending read or write for that process.

Each platform's AsyncIO backend gains a Registration struct that groups file descriptors (or IOCP completion keys on Windows) by process. Cancellation walks those descriptors and either deletes them from epoll/kqueue or calls CancelIoEx. A cancelledProcesses set rejects late registrations so a register-vs-cancel race cannot leak a stream. run clears the marker before reapProcess so PID reuse is safe.

Adds regression tests covering stdin, stdout, and stderr on Unix (POSIX shell + sleep grandchild) and Windows (PowerShell + Start-Process -NoNewWindow grandchild).

Resolves: https://github.com/swiftlang/swift-subprocess/issues/256

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -38,14 +38,14 @@ public struct Configuration: Sendable {
     public var environment: Environment
     /// The working directory to use when running the executable.
     ///
-    /// If this property is `nil`, the subprocess will inherit
-    /// the working directory from the parent process.
+    /// If this property is `nil`, the subprocess inherits the working
+    /// directory from the parent process.
     public var workingDirectory: FilePath?
     /// The platform-specific options to use when
     /// running the subprocess.
     public var platformOptions: PlatformOptions
 
-    /// Creates a Configuration with the parameters you provide.
+    /// Creates a configuration with the parameters you provide.
     /// - Parameters:
     ///   - executable: The executable to run.
     ///   - arguments: The arguments to pass to the executable.
@@ -107,37 +107,78 @@ public struct Configuration: Sendable {
         let processIdentifier = spawnResults.processIdentifier
 
         defer {
-            // Close process file descriptor now we finished monitoring
+            // Close the process file descriptor now that monitoring has finished.
             processIdentifier.close()
         }
 
         return try await withAsyncTaskCleanupHandler { () throws -> ExecutionOutcome<Result> in
-            let inputIO = spawnResults.inputWriteEnd()
-            let outputIO = spawnResults.outputReadEnd()
-            let errorIO = spawnResults.errorReadEnd()
+            // The counter coordinates a race between two finishers: the body
+            // closure and the process-termination monitor. Whichever side
+            // increments first observes a value of `1` and owns the response;
+            // the other side sees `2` and stays out of the way.
+            //
+            // If the monitor wins, the child has exited while the body is
+            // still reading or writing. The body might be blocked on a pipe
+            // that an inherited grandchild keeps open, so the monitor calls
+            // `cancelAsyncIO` to unblock it. If the body wins, the I/O
+            // already finished cleanly and no cancellation is needed.
+            let taskFinishFlag = AtomicCounter()
 
-            let result: Swift.Result<Result, any Swift.Error>
-            do {
-                // Body runs in the same isolation
-                let bodyResult = try await body(processIdentifier, inputIO, outputIO, errorIO)
+            let (result, monitorError) = try await withThrowingTaskGroup(
+                of: SubprocessError?.self,
+                returning: (Swift.Result<Result, any Swift.Error>, SubprocessError?).self
+            ) { group in
+                group.addTask {
+                    do throws(SubprocessError) {
+                        try await waitForProcessTermination(for: processIdentifier)
+                        if taskFinishFlag.addOne() == 1 {
+                            // The body closure hasn't finished but the child
+                            // process has terminated. Cancel all active
+                            // AsyncIO now.
+                            try AsyncIO.shared.cancelAsyncIO(for: processIdentifier)
+                        }
+                        return nil
+                    } catch {
+                        try? AsyncIO.shared.cancelAsyncIO(for: processIdentifier)
+                        return error
+                    }
+                }
 
-                result = .success(bodyResult)
-            } catch {
-                result = .failure(error)
+                let inputIO = spawnResults.inputWriteEnd()
+                let outputIO = spawnResults.outputReadEnd()
+                let errorIO = spawnResults.errorReadEnd()
+
+                let result: Swift.Result<Result, any Swift.Error>
+                do {
+                    // Body runs in the same isolation.
+                    let bodyResult = try await body(processIdentifier, inputIO, outputIO, errorIO)
+                    taskFinishFlag.addOne()
+                    result = .success(bodyResult)
+                } catch {
+                    result = .failure(error)
+                }
+
+                // Wait for the monitor child task to finish.
+                let monitorError = try await group.next() ?? nil
+                return (result, monitorError)
             }
 
-            // Ensure that we begin monitoring process termination after `body` runs
-            // and regardless of whether `body` throws, so that the pid gets reaped
-            // even if `body` throws, and we are not leaving zombie processes in the
-            // process table which will cause the process termination monitoring thread
-            // to effectively hang due to the pid never being awaited
-            let terminationStatus = try await monitorProcessTermination(
-                for: processIdentifier
-            )
+            // Drop the cancellation marker before reaping the zombie. After
+            // `reapProcess` runs the kernel can immediately reuse this PID,
+            // so the marker must be gone first; otherwise a concurrent
+            // `run()` that happens to inherit the same PID would see the
+            // stale entry and reject its registrations.
+            AsyncIO.shared.cleanup(processIdentifier: processIdentifier)
 
-            return ExecutionOutcome(
+            let terminationStatus = try reapProcess(with: processIdentifier)
+
+            if let monitorError {
+                throw monitorError
+            }
+
+            return try ExecutionOutcome(
                 terminationStatus: terminationStatus,
-                value: try result.get()
+                value: result.get()
             )
         } onCleanup: {
             let execution = Execution<Input, Output, Error>(
@@ -146,7 +187,7 @@ public struct Configuration: Sendable {
                 outputStream: nil,
                 errorStream: nil
             )
-            // Attempt to terminate the child process
+            // Attempt to terminate the child process.
             await execution.runTeardownSequence(
                 self.platformOptions.teardownSequence
             )
--- a/Sources/Subprocess/IO/AsyncIO+KQueue.swift
+++ b/Sources/Subprocess/IO/AsyncIO+KQueue.swift
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// AsyncIO implementation based on kqueue
+/// AsyncIO implementation based on kqueue.
 
 #if SUBPROCESS_ASYNCIO_KQUEUE
 
@@ -51,10 +51,7 @@ private let _kevent:
     ) -> Int32 = kevent
 
 private let _kqueueEventSize = 256
-private let _registration:
-    _Mutex<
-        [PlatformFileDescriptor: SignalStream.Continuation]
-    > = _Mutex([:])
+private let _registration: _Mutex<Registration> = _Mutex(Registration())
 
 private func _makeKevent(
     ident: UInt,
@@ -166,10 +163,11 @@ final class AsyncIO: Sendable {
         do throws(Errno) {
             thread = try pthread_create {
                 func reportError(_ error: SubprocessError) {
-                    _registration.withLock { store in
-                        for continuation in store.values {
-                            continuation.finish(throwing: error)
-                        }
+                    let continuations = _registration.withLock { store in
+                        return store.allContinuations()
+                    }
+                    for continuation in continuations {
+                        continuation.finish(throwing: error)
                     }
                 }
 
@@ -213,10 +211,7 @@ final class AsyncIO: Sendable {
                         }
 
                         let continuation = _registration.withLock { store -> SignalStream.Continuation? in
-                            if let continuation = store[targetFileDescriptor] {
-                                return continuation
-                            }
-                            return nil
+                            return store.continuation(for: targetFileDescriptor)
                         }
                         continuation?.yield(true)
                     }
@@ -294,6 +289,7 @@ final class AsyncIO: Sendable {
 
     internal func registerFileDescriptor(
         _ fileDescriptor: FileDescriptor,
+        processIdentifier: ProcessIdentifier,
         for event: Event
     ) -> SignalStream {
         return SignalStream { (continuation: SignalStream.Continuation) -> () in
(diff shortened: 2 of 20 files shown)


Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -311,76 +308,155 @@ final class AsyncIO: Sendable {
...
+            let cancelledContinuations: [SignalStream.Continuation] = _registration.withLock { storage in
+                let previousRegistrations = storage.cancel(processIdentifier: processIdentifier)
+                guard !previousRegistrations.isEmpty else {
+                    return []
+                }
+                var toBeCancelled: [SignalStream.Continuation] = []
+                for registration in previousRegistrations {
+                    toBeCancelled.append(registration.continuation)
+
+                    var kev = _makeKevent(
```
jakepetroules: nit: one of these styles would be a lot clearer / less repetitive:

```swift
for filter in [EVFILT_READ, EVFILT_WRITE] {
    var kev = _makeKevent(
        ident: UInt(registration.fileDescriptor),
        filter: Int16(filter),
        flags: UInt16(EV_DELETE)
    )
    _ = _kevent(
        state.kqueueFileDescriptor,
        &kev,
        1,
        nil,
        0,
        nil
    )
}
```

```swift
func deleteFilter(_ filter: Int) {
    var kev = _makeKevent(
        ident: UInt(registration.fileDescriptor),
        filter: Int16(filter),
        flags: UInt16(EV_DELETE)
    )
    _ = _kevent(
        state.kqueueFileDescriptor,
        &kev,
        1,
        nil,
        0,
        nil
    )
}
deleteFilter(EVFILT_READ)
deleteFilter(EVFILT_WRITE)
```

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -292,27 +309,78 @@ final class AsyncIO: Sendable {
...
+                    return nil
+                }
+
+                let rc = epoll_ctl(
+                    state.epollFileDescriptor,
+                    EPOLL_CTL_DEL,
+                    fileDescriptor.rawValue,
+                    nil
+                )
+                // Only inspect `errno` when the call actually failed; on
```
jakepetroules: nit: is this comment really necessary? That's just errno guidance in general, nothing particularly specific to epoll_ctl...

```
// Only inspect `errno` when the call actually failed; on
// success, `errno` may carry stale values from earlier
// syscalls on this thread.
```

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -237,19 +250,56 @@ final class AsyncIO: @unchecked Sendable {
...
+        let completionKeys: Set<UInt64> = _registration.withLock { storage in
+            return storage.cancel(processIdentifier: processIdentifier)
+        }
+        for key in completionKeys {
+            guard let handle = HANDLE(bitPattern: UInt(truncatingIfNeeded: key)) else {
+                continue
+            }
+            // `CancelIoEx` with a `NULL` overlapped cancels every pending
+            // I/O on the handle. `ERROR_NOT_FOUND` means there was nothing
+            // to cancel; `ERROR_INVALID_HANDLE` means the caller already
```
jakepetroules: Are you sure ERROR_INVALID_HANDLE is benign in this case? The docs don't mention it, but in the Unix analogue, that seems akin to receiving EBADF when calling a function like close() -- which should be an immediate fatalError right there, something you cannot ever recover from.

CC @weissi @compnerd

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -401,26 +479,174 @@ final class AsyncIO: @unchecked Sendable {
...
+
+    /// The result of a single attempt to register a handle.
+    enum RegisterResult {
+        /// The first time this completion key was seen. The caller still
+        /// needs to attach the handle to the IOCP via
+        /// `CreateIoCompletionPort`.
+        case registered
+        /// The completion key was already present. Windows can't
+        /// unregister from an IOCP, so the existing attachment is reused
+        /// and only the continuation is swapped.
```
jakepetroules: What does it mean the continuation is "swapped"? What happens to the old one; do we end up with leaking continuations that are never resumed?

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -66,12 +66,12 @@ extension Int32 {
 
 // MARK: - Process Monitoring
 @Sendable
-internal func monitorProcessTermination(
+internal func waitForProcessTermination(
```
jakepetroules, line 69: side question: I wonder if we should expose this as API now (but as a member function of Execution, probably). Calling this multiple times should be idempotent, and it may be useful for users to be able to run code once the process is guaranteed to have exited but before the zombie has been reaped / HANDLE has been closed.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -684,6 +684,22 @@ extension PlatformOptions: CustomStringConvertible, CustomDebugStringConvertible
 }
 #endif // !canImport(Darwin)
 
+@Sendable
+internal func reapProcess(
+    with processIdentifier: ProcessIdentifier
+) throws(SubprocessError) -> TerminationStatus {
+    do throws(Errno) {
+        // NOTE_EXIT may be delivered slightly before the process becomes reapable,
```
jakepetroules: This comment was Dispatch/kqueue specific and doesn't make sense here out of context. You could generalize it a bit, like:

> "On some platforms, the process exit notification (in particular NOTE_EXIT from kqueue) may be delivered slightly before the process becomes reapable, [...]"

Tests/SubprocessTests/AsyncIOTests.swift
```diff
@@ -346,3 +346,41 @@ extension SubprocessAsyncIOTests.TestBed {
...
+/// real `ProcessIdentifier` to pass. The tests only call `read` and
+/// `write` (never `cancelAsyncIO`), so this stand-in is sufficient.
+private var _currentPID: ProcessIdentifier {
+    #if os(Windows)
+    return ProcessIdentifier(
+        value: GetCurrentProcessId(),
+        processDescriptor: GetCurrentProcess(),
+        threadHandle: GetCurrentThread()
+    )
+    #elseif os(macOS)
```
jakepetroules: I'd flip the elseif-macos and else, and make the condition Linux || Android || FreeBSD.

macOS and OpenBSD don't have process descriptors; as-is this would fail to compile on OpenBSD

Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -2286,6 +2286,330 @@ extension SubprocessIntegrationTests {
     }
 }
 
+// MARK: - Cancellable IO Tests
+
+// Regression tests for https://github.com/swiftlang/swift-subprocess/issues/256.
+// Subprocess wrongly assumed that the death of the child process closes
```
jakepetroules: nit: seems weird for a comment to refer to past tense version of the code

jakepetroules:
Looks pretty good overall! Pointed out a few things...

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -292,27 +309,78 @@ final class AsyncIO: Sendable {
...
+                // success, `errno` may carry stale values from earlier
+                // syscalls on this thread. `ENOENT` means the descriptor
+                // is already gone from the epoll set (because the
+                // descriptor was closed or `cancelAsyncIO` removed it),
```
weissi: This sounds scary because it sounds like the state tracking is incomplete which is almost certainly makes it incorrect. 

Don't we have ABA issues here?

As in:
1. File descriptor (say) `17` gets `cancelAsyncIO`'d and `close`d
2. We trigger `removeRegistration`
3. File descriptor `17` gets reassigned to a new concurrently spawned process
4. `removeRegistration` (wrongly) removes it

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: I think it's benign here because technically, `CancelIoEx` is racing with the body closure, which means it also races with the body closure closing the handles naturally after the read. We call `cancelAsyncIO` immediately after the monitor task finishes, which has to run concurrently with the body closure because if it runs after, then it can't unblock the body from hanging. This race means it's possible that `CancelIoEx` is called right after the body closes the handle, which would trigger ERROR_INVALID_HANDLE.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Good catch. I made sure to finish the old one before the update.

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: The comment and `errno != ENOENT` check here are incorrect, and I removed them. Both `cancelAsyncIO` and `removeRegistration` are atomic operations wrapped in a lock. If fd `17` gets `cancelAsyncIO`, its reference will be removed by `cancelAsyncIO` first,  and `removeRegistration` will be a no-op.

Sources/Subprocess/IO/AsyncIO+Windows.swift
iCharlesHu: Actually after thinking through @weissi's comment below I think you are right that they are not benign. This situation shouldn't happen because we call `removeRegistration` in `_safelyClosed` so the reference should have been gone. I'll update.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
iCharlesHu: This is a good idea.. let's explore it separately in a different PR.

Sources/Subprocess/Configuration.swift
```diff
@@ -96,7 +96,8 @@ public struct Configuration: Sendable {
                 consuming IODescriptor?,
                 consuming IODescriptor?
             ) async throws -> Result
-        )
+        ),
+        isolation: (any Actor)? = #isolation
```
FranzBusch: Why are we adding this instead of adopting `nonisolated(nonsending)`?

Sources/Subprocess/Configuration.swift
```diff
@@ -116,28 +117,60 @@ public struct Configuration: Sendable {
...
+            // increments first observes a value of `1` and owns the response;
+            // the other side sees `2` and stays out of the way.
+            //
+            // If the monitor wins, the child has exited while the body is
+            // still reading or writing. The body might be blocked on a pipe
+            // that an inherited grandchild keeps open, so the monitor calls
+            // `cancelAsyncIO` to unblock it. If the body wins, the I/O
+            // already finished cleanly and no cancellation is needed.
+            let taskFinishFlag = AtomicCounter()
+            let monitorTask = Task<SubprocessError?, Never> {
```
FranzBusch: I am really worried about the usage of an unstructured task here. Subprocess management should not require the usage of an unstructured task at all. I expect the entire package to stay strictly within structured concurrency

iCharlesHu:
Move process monitoring to task group.

jakepetroules (approved):

Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -292,27 +310,75 @@ final class AsyncIO: Sendable {
...
+                for registration in previousRegistrations {
+                    toBeCancelled.append(registration.continuation)
+
+                    let rc = epoll_ctl(
+                        state.epollFileDescriptor,
+                        EPOLL_CTL_DEL,
+                        registration.fileDescriptor,
+                        nil
+                    )
+                    if rc != 0 && errno != ENOENT {
```
weissi, line 361: @iCharlesHu You said the check is incorrect and you removed it. But it looks like it has been merged with this in there?

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -311,34 +307,53 @@ final class AsyncIO: Sendable {
...
+                        1,
+                        nil,
+                        0,
+                        nil
                     )
+
+                    if rc != 0 {
+                        let capturedError = errno
+                        _ = storage.removeRegistration(for: fileDescriptor.rawValue)
+                        let error: SubprocessError = .asyncIOFailed(
```
weissi, line 341: We should crash here, no? If `kevent` fails to remove a registration, then something is _severely_ wrong. If we just send this error to the user, it's entirely not actionable to them. This would be programmer error in subprocess --> we need to crash

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -350,39 +365,84 @@ final class AsyncIO: Sendable {
...
+                for registration in previousRegistrations {
+                    toBeCancelled.append(registration.continuation)
+
+                    for filter in [EVFILT_READ, EVFILT_WRITE] {
+                        var kev = _makeKevent(
+                            ident: UInt(registration.fileDescriptor),
+                            filter: Int16(filter),
+                            flags: UInt16(EV_DELETE)
+                        )
+                        _ = _kevent(
```
weissi, line 419: let's not ignore error. Let's check and crash if there's an error

Sources/Subprocess/IO/AsyncIO+KQueue.swift
```diff
@@ -350,39 +365,84 @@ final class AsyncIO: Sendable {
...
+                    return nil
+                }
+
+                for filter in [EVFILT_READ, EVFILT_WRITE] {
+                    var kev = _makeKevent(
+                        ident: UInt(fileDescriptor.rawValue),
+                        filter: Int16(filter),
+                        flags: UInt16(EV_DELETE)
+                    )
+                    _ = _kevent(
```
weissi, line 383: let's not ignore errors. let's check them and crash on error

PR #273, opened 2026-05-27
Fix process monitoring hang on Linux kernels < 5.4

### LINKED ISSUE #271: subprocesses hanging with older Linux kernels
**Describe the bug**
 On kernels > 5.4, swift-subprocess uses pidfd_open to monitor child process exit. On older kernels, it falls back to a self-pipe signal handler approach (Subprocess+Linux.swift) which is broken on older kernels. When a child process exits, SIGCHLD fires but is discarded, and the awaiting continuation is never resumed causing a permanent hang in the parent process

**To Reproduce**
Run this source on an older ~ 4.14 kernel
```
  import Subprocess

      .name("sleep"),
      arguments: ["1"],
      input: .standardInput,
      output: .standardOutput,
      error: .standardError
  )
```

**Expected behavior**
Expect the process to exit cleanly and not hang

 - kernel: 4.14.327-246.539.amzn2.aarch64
- Swift version 6.2.4 (swift-6.2.4-RELEASE)
- Target: aarch64-unknown-linux-gnu

**Additional context**

iCharlesHu:
Install a SIGCHLD handler via a new `_subprocess_install_sigchld_handler` C shim so the kernel notifies the monitor thread when children exit. Pass an empty sigmask to `epoll_pwait` so `SIGCHLD` is deliverable during the wait, and mark both ends of the self-pipe `O_NONBLOCK` so the drain loop in `_notifyAllKnownChildProcesses` exits cleanly.

Call close_range via a direct syscall wrapper on Linux, with fallback definitions for `SYS_close_range` and `CLOSE_RANGE_CLOEXEC`, so the build works against glibc < 2.34 and older kernel-headers (e.g. Amazon Linux 2 with kernel 4.14).

Manually tested on running a AmazonLinux 2 VM: `4.14.355-282.729.amzn2.x86_64`.

Resolves: https://github.com/swiftlang/swift-subprocess/issues/271

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Linux.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Linux.swift
@@ -103,7 +103,9 @@ internal func waitForProcessTermination(
                         }
                         // Now save the registration
                         var newState = storage
-                        newState.continuations[processIdentifier.processDescriptor] = continuation
+                        var list = newState.continuations[processIdentifier.processDescriptor] ?? []
+                        list.append(continuation)
+                        newState.continuations[processIdentifier.processDescriptor] = list
                         state = .started(newState)
                         // No state to resume
                         return nil
@@ -114,10 +116,14 @@ internal func waitForProcessTermination(
                         // is not the case and only save continuation then.
                         switch Result(catching: { () throws(Errno) -> Bool in try processIdentifier.peekIfExited() }) {
                         case let .success(processExited):
-                            // Save this continuation to be called by signal handler
-                            var newState = storage
-                            newState.continuations[processIdentifier.value] = continuation
-                            state = .started(newState)
+                            if !processExited {
+                                // Save this continuation to be called by signal handler
+                                var newState = storage
+                                var list = newState.continuations[processIdentifier.processDescriptor] ?? []
+                                list.append(continuation)
+                                newState.continuations[processIdentifier.value] = list
+                                state = .started(newState)
+                            }
                             return .success(processExited)
                         case let .failure(underlyingError):
                             let error: SubprocessError = .failedToMonitor(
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -114,10 +114,12 @@ internal func waitForProcessTermination(
...
                         switch Result(catching: { () throws(Errno) -> Bool in try processIdentifier.peekIfExited() }) {
                         case let .success(processExited):
-                            // Save this continuation to be called by signal handler
-                            var newState = storage
-                            newState.continuations[processIdentifier.value] = continuation
-                            state = .started(newState)
+                            if !processExited {
+                                // Save this continuation to be called by signal handler
+                                var newState = storage
+                                newState.continuations[processIdentifier.value] = continuation
```
jakepetroules: What about the existing continuation, if any?

iCharlesHu: Updated to keep a list of continuations. This aligns with our plan to make `waitForProcessTermination` idempotent.

jakepetroules: Thanks. It's worth checking if there are any other places in the codebase that are potentially dropping continuations without resuming them, as that would be a violation of the API contract.

jakepetroules (approved):
testStreamingErrorOutput failed on Windows (nondeterministic?) but that's definitely not a regression from this change.

iCharlesHu:
Yeah a rerun came out green. I'll keep an eye out.. although I haven't run into this one locally on Windows 🤔

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -333,6 +334,18 @@ private let setup: () = {
         do {
             let (readEnd, writeEnd) = try FileDescriptor.pipe()
             _signalPipe = (readEnd.rawValue, writeEnd.rawValue)
+            // Make the pipe non-blocking. The read end MUST be non-blocking
+            // so the drain loop in _notifyAllKnownChildProcesses exits cleanly
+            // with EAGAIN once the pipe is empty.
+            for fd in [readEnd.rawValue, writeEnd.rawValue] {
+                let existing = fcntl(fd, F_GETFL)
+                if existing < 0 {
+                    throw Errno(rawValue: errno)
```
vlm, line 343: This leaks `readEnd`, and `writeEnd`.

PR #276, opened 2026-05-28
Make Unix `testSuspendResumeProcess()` deterministic

broken-circle:
`testSuspendResumeProcess()` fails 211/5000 times (~4.2%) on Darwin with the following error:

> "Failed to monitor the child process exit status with underlying error: No child processes" (ECHILD)

It also fails on Android (Linux) as in #269.

The original test spawns `cat` with `input: .none`, redirecting `stdin` to `/dev/null`. With no input, `cat` reads EOF immediately and exits, racing the test's `send(signal: .suspend)`. When `cat` wins the race, it terminates before being suspended. 

The test then calls `waitpid()` to inspect and consume the child's status. In the intended case, `cat` is suspended and `waitpid()` consumes the stopped-state notification. In the race-loser case, `cat` has already exited and `waitpid()` consumes the exit notification, reaping the zombie before Subprocess can do so.

`Configuration.run()` subsequently calls `waitid()` via `monitorProcessTermination()`, but the kernel no longer has a record of the child, so Subprocess throws `failedToMonitorProcess` wrapping `ECHILD`.

### Design

This PR shifts verification from kernel-state inspection to observable behavior. The revised test:

1. Spawns `cat` with manually-managed input and output pipes.
2. Confirms `cat` echoes before any signals are sent.
3. Stops `cat`, writes two more lines, waits 200ms, and confirms no output appears.
4. Resumes `cat` and confirms both buffered lines emerge in order.

Regarding the 200ms timeout in step 3, any value large enough to outlast normal scheduling variance would work. Generally, a longer timeout strengthens the assertion.

This test introduces the use of Swift Testing's built-in timeout, which isn't currently used elsewhere in the test suite. In #267, @jakepetroules and I were talking about whether to use Swift Testing's timeouts (which have coarse granularity of at least 1 minute), or introducing a `withTimeout()` helper that accepts arbitrarily small timeouts. This test doesn't materially benefit from a smaller timeout, so I prefer using the built-in functionality.

I considered alternatives such as polling `proc_pidinfo` for `pbi_status == SSTOP` (rejected polling in general) or a local kqueue with `EVFILT_PROC | NOTE_SIGNAL` (requires building notification infrastructure for a single test).

This test passes 100% of the time across 2500 local iterations on Darwin. I ran it on Android via GitHub Actions ~50 times and some runs pass, but others fail with infrastructure issues setting up the emulator. Will be more reliable running on the real Subprocess CI.

### Windows

Windows uses Powershell and does not seem to have the same problem. It may be worth it to rewrite it anyway in a follow-up PR for consistency.

### DIFF
--- a/Tests/SubprocessTests/DarwinTests.swift
+++ b/Tests/SubprocessTests/DarwinTests.swift
@@ -75,30 +75,6 @@ struct SubprocessDarwinTests {
         }
         #expect(FilePath(currentDir) == expected)
     }
-
-    @Test func testSuspendResumeProcess() async throws {
-        _ = try await Subprocess.run(
-            // This will intentionally hang
-            .path("/bin/cat"),
-            input: .none,
-            output: .discarded,
-            error: .discarded
-        ) { subprocess in
-            // First suspend the process
-            try subprocess.send(signal: .suspend)
-            var suspendedStatus: Int32 = 0
-            waitpid(subprocess.processIdentifier.value, &suspendedStatus, WNOHANG | WUNTRACED)
-            #expect(_was_process_suspended(suspendedStatus) > 0)
-            // Now resume the process
-            try subprocess.send(signal: .resume)
-            var resumedStatus: Int32 = 0
-            waitpid(subprocess.processIdentifier.value, &resumedStatus, WNOHANG | WUNTRACED)
-            #expect(_was_process_suspended(resumedStatus) == 0)
-
-            // Now kill the process
-            try subprocess.send(signal: .terminate)
-        }
-    }
 }
 
 #endif // canImport(Darwin)
--- a/Tests/SubprocessTests/LinuxTests.swift
+++ b/Tests/SubprocessTests/LinuxTests.swift
@@ -29,7 +29,6 @@ import SystemPackage
 import FoundationEssentials
 
 import Testing
-import _SubprocessCShims
 @testable import Subprocess
 
 // MARK: PlatformOption Tests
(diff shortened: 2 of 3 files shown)


Tests/SubprocessTests/DarwinTests.swift
```diff
@@ -76,29 +76,106 @@ struct SubprocessDarwinTests {
...
             try subprocess.send(signal: .resume)
-            var resumedStatus: Int32 = 0
-            waitpid(subprocess.processIdentifier.value, &resumedStatus, WNOHANG | WUNTRACED)
-            #expect(_was_process_suspended(resumedStatus) == 0)
+            try #require(try await readLine(from: outputPipe.readEnd, timeout: .seconds(2)) == "first")
+            try #require(try await readLine(from: outputPipe.readEnd, timeout: .seconds(2)) == "second")
 
-            // Now kill the process
+            // Tear down. Closing stdin causes cat to exit naturally.
+            try inputPipe.writeEnd.close()
```
jakepetroules: Can you use the closeAfter wrapper to ensure the fd is closed even in error cases?

Tests/SubprocessTests/DarwinTests.swift
```diff
@@ -76,29 +76,106 @@ struct SubprocessDarwinTests {
...
+        let inputPipe = try FileDescriptor.pipe()
+        let outputPipe = try FileDescriptor.pipe()
+
+        // Make the read end non-blocking so readLine() can poll with a
+        // deadline instead of blocking indefinitely.
+        let flags = fcntl(outputPipe.readEnd.rawValue, F_GETFL)
+        try #require(fcntl(outputPipe.readEnd.rawValue, F_SETFL, flags | O_NONBLOCK) == 0)
+
+        defer {
+            try? inputPipe.writeEnd.close()
```
jakepetroules: This will double-close with the call on line 129, which is invalid

Tests/SubprocessTests/DarwinTests.swift
```diff
@@ -76,29 +76,106 @@ struct SubprocessDarwinTests {
...
+            // Tear down. Closing stdin causes cat to exit naturally.
+            try inputPipe.writeEnd.close()
             try subprocess.send(signal: .terminate)
         }
     }
 }
 
+// MARK: - Utilities
+/// Writes a line plus newline to the given file descriptor.
+private func writeLine(_ line: String, to fd: FileDescriptor) throws {
```
jakepetroules: nit: I'd make these extension methods on `FileDescriptor` rather than globals, e.g.

```swift
extension FileDescriptor {
    private func writeLine(_ line: String) throws { ... }
    private func readLine(timeout: Duration) async throws -> String? { ... }
}
```

jakepetroules:
I think this now makes the test cross-platform, right?

Can you move this into UnixTests and remove the Linux copy? It also calls waitpid and has the same race issue, as you mentioned.

broken-circle:
@jakepetroules Yeah, it's cross-platform for Unix now. I went with a new `ProcessControlTests` suite over `UnixTests`, since the latter is gated `#if canImport(Darwin) || canImport(Glibc)`, which excludes Android. Widening `UnixTests`'s gate would change the platform set for every test in the file, which felt out of scope for this fix, but I can widen the gate and fold this in if you'd prefer.

Also fixed the other issues you brought up. Thanks!

jakepetroules:
This looks great, thanks!

I think keeping it narrow for now makes sense. We should explore in a separate PR extending those tests to Android.

broken-circle:
Great! I'll start working on that and open it as a follow-up PR after this one lands.

iCharlesHu (approved):
Thanks or another great PR!!

PR #277, opened 2026-05-28
Update pull request workflow to use main branch

jakepetroules:
0.0.11 is unstable and broken for Android

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   tests:
     name: Test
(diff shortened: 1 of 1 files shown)


iCharlesHu:
It looks like Android is still failing.

Is this change temporary? Should we switch back to a release (as opposed to `main`) once the Android build fix is in?

jakepetroules:
Yes. Talking with @shahmishal about getting a new tag in, and we also need https://github.com/swiftlang/github-workflows/pull/276 to land.

jakepetroules:
And @broken-circle's test fix in #276 should fix the other Android failures.

iCharlesHu (approved):

PR #278, opened 2026-05-29
Expand Unix tests to Android

broken-circle:
This PR follows up on #276 by expanding the existing Unix test suite to also run on Android. Most tests run and pass on Android as-is; the only change beyond the two per-test fixes below is aligning this file's platform gate and imports with the library's (`Darwin || Glibc || Android || Musl`).

Two tests required specific changes:

- `testSubprocessPlatformOptionsCreateSession()`: Android's `ps` doesn't support `-o pid,pgid,tpgid`, so on Android this test reads the same fields directly from `/proc/$$/stat` instead, using a new helper. This test doesn't validate any Android-specific behavior, so I also considered just disabling it on Android, but the change is small and the coverage is slightly stronger (checking session leadership with `pid == session`), so I kept it.

- `testExitSignal()`: When terminated by a catchable signal, Android's `/bin/sh` [exits normally with a status of `128+n`](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Signal-Handling.html), so the result was `.exited(130)` vs `.signaled(2)` on Darwin/Linux. Added an Android-specific branch.

This PR also moves `ProcessControlTests.testSuspendResumeProcess()` into the Unix test suite, since it was previously separated only due to the Unix test suite not running on Android.


Tests/SubprocessTests/UnixTests.swift
```diff
@@ -9,13 +9,15 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if canImport(Darwin) || canImport(Glibc)
+#if canImport(Darwin) || canImport(Glibc) || canImport(Android) || canImport(Musl)
```
jakepetroules: Maybe this should just be `#if !os(Windows)`?

This insulates the code against future changes, like if we rename the FreeBSD libc module from Glibc to FreeBSD, or future OSes that might not use Glibc as their libc module name, and ensures they aren't unintentionally skipped.

jakepetroules (approved):

Tests/SubprocessTests/UnixTests.swift
broken-circle: Yeah, that makes sense. Updated.

One wrinkle: `!os(Windows)` also matches WASI, which has no process model, so a Unix-only file gated that way will fail to build on Wasm instead of being cleanly skipped the way it would with a full `canImport` gate. That pattern shows up 22 times across 7 files. Subprocess doesn't build for Wasm today, but since Wasm became a supported target as of Swift 6.2, it'd be reasonable future-proofing. Would you be open to a follow-up PR that tightens those conditions to `!os(Windows) && !os(WASI)`?

iCharlesHu:
@broken-circle please rebase to pick up https://github.com/swiftlang/swift-subprocess/commit/d963ca30a8bf39ce3b29c39f480168ad87a84d8c. Android build should start passing with it.

broken-circle:
@iCharlesHu Great, thanks! Rebased now.

Tests/SubprocessTests/UnixTests.swift
jakepetroules: iOS is in the same boat, and when I discussed this with @iCharlesHu a while back his position was that the package should just fail to compile on iOS since the underlying APIs aren't available.

I expect we'd do the same for WASI since none of the API surface would be useful on that platform either.

broken-circle: Makes sense. Failing to compile is the clearest signal on unsupported platforms.

iCharlesHu (approved):
Thanks!

PR #280, opened 2026-05-29
Test old Linux kernels using Qemu

cmcgee1024:
Fixes the following:
* Increases test confidence with older Linux kernels (4.18, and 5.10)
* Closes file handles that were leaking on Linux
* Better soft limit detection so that testConcurrentRun() can avoid exhausting system open file limit
* Handle the epoll(DEL) case when a process is being shut down for Linux with older kernels


Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -552,6 +545,7 @@ internal func _isWaitprocessDescriptorSupported() -> Bool {
         // If we can not retrieve pidfd, the system does not support waitid(P_PIDFD)
         return false
     }
+    defer { try? FileDescriptor(rawValue: selfPidfd).close() }
```
jakepetroules, line 575: Can you use closeAfter instead?

cmcgee1024: I don't think that it would work because we would lose the errno comparison if the close of the file descriptor fails, which we want to be just a best effort here.

Sources/Subprocess/Thread.swift
```diff
@@ -138,17 +139,25 @@ private final class WorkQueue: Sendable {
         return body(mutex, &queue)
     }
 
-    // Only called in worker thread. Sleeps the thread if there's no more item
+    // Only called in worker thread. Sleeps the thread if there's no more item.
+    // Uses an explicit while loop to handle spurious pthread_cond_wait wakeups,
+    // which are common on FreeBSD and permitted by POSIX.
     func dequeue() -> BackgroundWorkItem? {
-        return self.withUnsafeUnderlyingLock { queue in
```
jakepetroules, line 143: It's concerning that this was changed; why couldn't this be expressed in terms of the withUnsafeUnderlyingLock abstraction? We should quite explicitly keep the business logic separate from the generic lock mechanics; if the abstraction needs changes, let's make them.

cmcgee1024: This was done to address some of the hanging and stability issues that were encountered on FreeBSD. Now that the FreeBSD stability improvements are merged I'll try without these changes that were made to handle the spurious `pthread_cond_wait` wakeups encountered on that platform, that is apparently still allowed by POSIX.

cmcgee1024: I've merged the changes with the latest in main branch by reverting the FreeBSD specifics. The tests didn't hang this time.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -528,6 +528,12 @@ extension Configuration {
                 // Spawn error
                 if spawnError != 0 {
                     if [ENOENT, EACCES, ENOTDIR].contains(spawnError) {
+                        // clone3(CLONE_PIDFD) allocates a pidfd before exec runs.
+                        // If exec fails we retry with the next candidate path, so
+                        // close the pidfd here to avoid leaking it across retries.
+                        if processDescriptor > 0 {
+                            try? FileDescriptor(rawValue: processDescriptor).close()
```
iCharlesHu: We shouldn't swallow the close error here. We can convert it to SubprocessError and throw it directly.

Sources/Subprocess/Configuration.swift
```diff
@@ -118,8 +118,16 @@ public struct Configuration: Sendable {
...
+                            // AsyncIO now.  Use try? rather than try: the
+                            // process has already exited, so any epoll DEL
+                            // failure here is harmless — dropped continuations
+                            // will terminate the I/O streams on their own.
+                            // Propagating this error as monitorError would
+                            // trigger onCleanup → SIGKILL on an already-dead
+                            // process, and if Subprocess.run() then throws,
+                            // the caller's task group cascades cancellation to
+                            // other live processes.
+                            try? AsyncIO.shared.cancelAsyncIO(for: processIdentifier)
```
iCharlesHu: Should we make `cancelAsyncIO` recognize `DEL` failure as harmless here rather than using `try?`? Same as above I worry that we accidentally swallow unintended errors.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -528,6 +528,12 @@ extension Configuration {
                 // Spawn error
                 if spawnError != 0 {
                     if [ENOENT, EACCES, ENOTDIR].contains(spawnError) {
+                        // clone3(CLONE_PIDFD) allocates a pidfd before exec runs.
+                        // If exec fails we retry with the next candidate path, so
+                        // close the pidfd here to avoid leaking it across retries.
+                        if processDescriptor > 0 {
```
jakepetroules: To be pedantic, I think this should be `>= 0` or `!= 0` because _technically_ 0 is in the range of valid file descriptors.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -447,36 +446,45 @@ private func _unregisterProcessDescriptorAndNotify(_ pidfd: CInt, context: Monit
...
-        let rc = epoll_ctl(
+        // Remove this pidfd from epoll to prevent further notifications.
+        // Ignore the return value: if DEL fails (e.g., ENOENT due to a
+        // concurrent removal, or a transient kernel error on older 5.x kernels),
+        // the process has still exited and the continuation must be resumed
+        // normally.  The fd is removed from epoll automatically by the kernel
+        // when processIdentifier.close() closes it, so a failed DEL here is
+        // never permanent.  Propagating a DEL error as a monitoring failure
+        // would trigger onCleanup → SIGKILL against an already-dead process.
+        _ = epoll_ctl(
```
iCharlesHu: Should we specifically ignore `DEL` as opposed to ignore all errors?

Tests/SubprocessTests/UnixTests.swift
```diff
@@ -895,52 +895,60 @@ extension FileDescriptor {
...
+        // varies across platforms and Swift versions, so calling getrlimit
+        // directly from Swift is not reliably portable.
+        // Cap at 4096: Docker containers can report limits like 2^20.
+        let softLimit = Int(min(_subprocess_nofile_soft_limit(), UInt64(4096)))
+
+        // On Linux, account for any fds already open (e.g. from prior tests in
+        // the same suite) to avoid hitting EMFILE during the concurrent spawn
+        // burst.  /proc/self/fd lists every open descriptor; subtracting the
+        // current count plus a small margin gives the true available headroom.
+        #if os(Linux)
```
jakepetroules: I assume we also want to take this code path on Android? If so you also need `|| os(Android)` since os Linux only covers GNU/Linux

cmcgee1024: It should work with Android too, so I've added the condition here.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
cmcgee1024: Yes, this makes sense that it could be conceptually zero, so I've updated it to be >= 0.

cmcgee1024: I had to revert this to `> 0` because it was causing some of the macOS tests to fail.

jakepetroules: Why not update the tests?

cmcgee1024: It's because the tests are verifying correct behaviour. Allowing zero has an impact to macOS.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
cmcgee1024: I've wrapped any errors thrown here in SubprocessError in the newest revision.

Sources/Subprocess/Configuration.swift
cmcgee1024: I've removed this bit of Linux-specific knowledge in the comment and the attempt to discard any errors thrown here. The epoll_ctl(DEL) case is now fully encapsulated, see comments further down for more detail.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
cmcgee1024: According to the Linux manual the only other errors, other than the ENOENT observed during older kernel testing, would be various flavours of programming errors, such as bad file descriptor, or already closed. I've captured the rc and errno to form an assert that only gets activated in debug/test binaries so that in testing we can gather details about the programming error, and it has no effect in production.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -447,36 +446,53 @@ private func _unregisterProcessDescriptorAndNotify(_ pidfd: CInt, context: Monit
...
-        if rc != 0 {
-            let epollErrno = errno
-            let error = SubprocessError.failedToMonitor(
-                withUnderlyingError: Errno(rawValue: epollErrno)
-            )
-            return (continuationList, error)
-        }
-
-        return (continuationList, nil)
+        assert(delRC == 0 || errno == ENOENT, "epoll_ctl(DEL) failed unexpectedly: \(errno)")
```
iCharlesHu: Subprocess doesn't really use `assert`. We should follow the old format and resume with an error to propagate the error. If we want to only ignore `ENOENT` we can do something like

```
if delRC != 0 && errno != ENOENT {
    let error = SubprocessError.failedToMonitor(
        withUnderlyingError: Errno(rawValue: epollErrno)
    )
    return (continuationList, error)
}

return (continuationList, nil)
```

Sources/Subprocess/Platforms/Subprocess+Unix.swift
iCharlesHu: (nit) I've added a `.invalidDescriptor` in https://github.com/swiftlang/swift-subprocess/commit/b11e6838fde69f7bca87f2298102f43ddbf19327 which you can use here: `if processDescriptor != .invalidDescriptor`. But the original way is correct too.

.swift-version
```diff
@@ -1 +1 @@
-6.2.0
\ No newline at end of file
+6.3.2
```
iCharlesHu, line 1: Just curious: what's our policy to update this file when new Swift is released? Do we always try to use the current release?

iCharlesHu (approved):

.swift-version
jakepetroules: Honestly, I end up deleting it in my working tree half the time because it causes problems when using a different version.

This is really meant for "apps" which build with one canonical version only, not libraries which build with multiple. [/rant]

cmcgee1024: In term of the policy of swift-subprocess and the swift version file, it's up to the code owners here what to do with it.

The swift version file is there to record that at least one person, and ideally the CI too, has verified that the project can be developed on a specific toolchain version with a degree of confidence that the package builds, tests will build and pass, the language server works, and things like code formatting work without extraneous diffs. The hope is to lessen "it works at my desk" kinds of problems with a measure of reproducibility, and to encourage new developers to a package because things just work right away.

By no means does this file indicate that this is the _only_ version of the toolchain that can be used with this package, or the only one it supports. This is just the default that gets developers working with it to a known good configuration for most things and swiftly can help them with that.

Also, it has no effect on dependent packages. This is for development of this package, even if the package is a library package like this one.

It's worth noting that this file instructs what toolchain version to use for testing with the Linux kernel versions added as GH workflows in this PR as it stands. The Qemu setup script uses of swiftly to install the swift toolchain into the Linux VM that normally won't have it. It's currently configured to use this file as it's visible, easy to update, and swiftly can use it.

Please let me know if you'd prefer the test script to always use "latest", or something that's statically coded into the test script, or the GH workflow. I think that this file is simpler, more visible, and more reproducible than those alternatives.

Sources/Subprocess/Platforms/Subprocess+Linux.swift
cmcgee1024: Thanks, I've adopted this approach.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
cmcgee1024: Thanks, I've switched it over to use this constant and it's passing on macOS now.

PR #281, opened 2026-05-29
Make graceful shutdown return as soon as the subprocess exits

iCharlesHu:
Race process-termination monitoring against the teardown sequence in a task group: if the monitor finishes first, the child has exited, so cancel the teardown task and return immediately; if teardown finishes first (after its terminal ), wait for the monitor to observe the exit.

Make `waitForProcessTermination` on all platforms `idempotent`, since now we are calling it multiple times.

Resolves: rdar://174278515

### DIFF
--- a/Tests/SubprocessTests/UnixTests.swift
+++ b/Tests/SubprocessTests/UnixTests.swift
@@ -223,6 +223,71 @@ extension SubprocessUnixTests {
     }
 }
 
+// MARK: - Teardown Timing
+extension SubprocessUnixTests {
+    /// Spawns a child that prints `ready` and then blocks in `sleep`, waits
+    /// for that marker so the child is known to be running, then cancels the
+    /// run to trigger teardown and returns how long the teardown took.
+    private func measureCancelledTeardown(
+        using teardownSequence: [TeardownStep]
+    ) async -> Duration {
+        let (readyStream, readyContinuation) = AsyncStream.makeStream(of: Void.self)
+        return await withTaskGroup(
+            of: Void.self,
+            returning: Duration.self
+        ) { group in
+            group.addTask {
+                var platformOptions = PlatformOptions()
+                // Isolate the child in its own session so teardown can't reach
+                // anything but the process we spawned.
+                platformOptions.createSession = true
+                platformOptions.teardownSequence = teardownSequence
+                let configuration = Configuration(
+                    .path("/bin/sh"),
+                    // `exec` so the monitored child becomes `sleep` itself,
+                    // which dies instantly on SIGTERM/SIGKILL.
+                    arguments: ["-c", "echo ready; exec sleep 10"],
+                    platformOptions: platformOptions
+                )
+                _ = try? await Subprocess.run(
+                    configuration,
+                    input: .none,
+                    output: .sequence,
+                    error: .discarded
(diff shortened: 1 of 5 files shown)


Tests/SubprocessTests/UnixTests.swift
```diff
@@ -239,6 +223,69 @@ extension SubprocessUnixTests {
...
+                    platformOptions: platformOptions
+                )
+                _ = try? await Subprocess.run(
+                    configuration,
+                    output: .string(limit: 1024),
+                    error: .string(limit: 1024)
+                )
+            }
+            // Give the child time to spawn, then cancel to trigger teardown.
+            try? await Task.sleep(for: .milliseconds(200))
```
jakepetroules: To eliminate timing sensitivity issues, how about running a shell script which echoes 'hello' and then runs sleep, and waiting to read 'hello' from the output before invoking cancellation?

Tests/SubprocessTests/UnixTests.swift
```diff
@@ -239,6 +223,69 @@ extension SubprocessUnixTests {
...
+            group.cancelAll()
+        }
+        return clock.now - start
```
jakepetroules: Prefer [Clock.measure](https://developer.apple.com/documentation/swift/clock/measure(isolation:_:))

Tests/SubprocessTests/UnixTests.swift
jakepetroules: or, you could use the body overload with an AsyncStream? e.g.

```swift
let (stream, continuation) = AsyncStream.makeStream(of: Void.self)
group.addTask {
    try await Subprocess.run(..., {
        continuation.finish()
    })
}
// Give the child time to spawn, then cancel to trigger teardown.
await stream.makeAsycIterator().next()
group.cancelAll()
```

Tests/SubprocessTests/UnixTests.swift
```diff
@@ -239,6 +223,69 @@ extension SubprocessUnixTests {
...
+    }
+
+    @Test func testKillTeardownReturnsAsSoonAsProcessExits() async {
+        let elapsed = await self.measureCancelledTeardown(using: [
+            .send(signal: .kill, allowedDurationToNextStep: .seconds(5))
+        ])
+        #expect(elapsed < .milliseconds(500), "SIGKILL is uncatchable; teardown should not wait")
```
jakepetroules: also timing sensitive, but probably overall probably much less likely to flake

Tests/SubprocessTests/UnixTests.swift
iCharlesHu: Great point! Will do.

Tests/SubprocessTests/UnixTests.swift
iCharlesHu: Yeah... but unfortunately the idea is this test is to test timing.  I'll increase it to less than 5 I think that's good enough of a signal.

iCharlesHu: I also removed `testShortTeardownTimeoutCompletesPromptly` because it's too timing sensitive. The other two tests should give us enough signal.

jakepetroules (approved):

PR #284, opened 2026-05-31
Avoid leaking implementation in `PlatformOptions` description

broken-circle:
`PlatformOptions.description(withIndent:)` renders  `ConsoleBehavior` and `WindowStyle` with `String(describing:)`, which exposes the private `storage` wrapper (for example, `"ConsoleBehavior(storage: inherit)"`) instead of just `"inherit"`, leaking an implementation detail that isn't part of the public API.

This PR gives both types `CustomStringConvertible` conformance and interpolates them directly.

I considered making the two nested `Storage` enums have `String` raw values, but this seems more consistent with the existing style.

## Future directions

Should the description of `UserCredentials` redact `password`? I think it might be a good idea. If there's interest, I'll open a follow-up PR to render it the same way, with a redacted password.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -453,14 +453,58 @@ public struct PlatformOptions: Sendable {
     public init() {}
 }
 
+extension PlatformOptions.ConsoleBehavior: CustomStringConvertible, CustomDebugStringConvertible {
+    /// A textual representation of how the console appears when spawning a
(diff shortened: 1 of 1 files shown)


jakepetroules:
> Should the description of UserCredentials redact password? I think it might be a good idea. If there's interest, I'll open a follow-up PR to render it the same way, with a redacted password.

I actually think we should remove those APIs for now, to give us more time to consider whether that's the right API shape in the first place - opened #285.

iCharlesHu (approved):

PR #288, opened 2026-06-01
Add back standardOutput and standardError as deprecated renamed properties

marcprux:
The rename of `standardOutput ` to `currentStandardOutput ` and `standardError ` to `currentStandardError ` in https://github.com/swiftlang/swift-subprocess/pull/233 is causing breakage in various projects like https://github.com/swiftlang/swift-java/pull/775

This PR brings them back as a stopgap deprecated/renamed property to prevent projects from having to pin to 0.4.


iCharlesHu (approved):
Thanks!

PR #289, opened 2026-06-02
Make `testConcurrentSlowExitsDoNotHang()` deterministic and leak-free

broken-circle:
`testConcurrentSlowExitsDoNotHang()` is flaky, and under repeated runs it also leaks processes until spawning fails with `EAGAIN`.

The test spawns 16 bash children running `trap 'sleep 1; exit 0' TERM; sleep 300 & wait`, sends each `SIGTERM` after 100ms, and asserts every child exits `0`. It fails in two ways, both rooted in the trap/signal/background-sleep design.

First, the exit status is not deterministic. `SIGTERM` interrupts the `wait`, which returns `128+SIGTERM` (`143`); the trap's `exit 0` doesn't reliably override that status, so the run occasionally reports `.exited(143)`. On Darwin's system bash (3.2) this reproduces about 1/50 times.

Second, when the trap fires `exit 0`, the backgrounded `sleep 300` is orphaned, reparented to `launchd`, and runs for its full 300 seconds. Nothing reaps it, so each iteration strands 16 `sleep 300` processes; when run repeatedly, they accumulate until the process limit is reached and every subsequent attempt fails with `EAGAIN`, producing a cascade of subsequent failures. I confirmed this by inspection when running the test; after a single run, 16 orphaned `sleep 300` (PPID 1) remain and persist past the test process's own exit, so nothing on the success path could have reaped them.

This PR uses 16 body-less `/bin/sleep 1` children. The reaper keys on child exit (`SIGCHLD`), which is identical whether a child exits on a timer or a signal, so the children still exit in a tight burst and flood the reaper. The body-less runs keep all 16 monitors blocked on termination for the full second, so the exits land while every monitor is waiting, reproducing the condition the test guards against. This approach has the benefit of being deterministic, leaking nothing, and not relying on bash. The time limit is added to avoid hanging in case of reaper regressions. This test passes when run 1,000 times consecutively.

### DIFF
--- a/Tests/SubprocessTests/UnixTests.swift
+++ b/Tests/SubprocessTests/UnixTests.swift
@@ -733,39 +733,26 @@ extension SubprocessUnixTests {
         }
     }
 
-    // Ensure Subprocess does not hang on Linux when several concurrent
-    // `Subprocess.run` calls saw their children exit in a tight burst.
-    //
-    // Spawns 16 `bash` children, each with a SIGTERM trap that sleeps one second
-    // before exiting. Every body closure sends SIGTERM at roughly the same instant,
-    // so all 16 children finish their trap and become zombies inside a single small.
-    @Test(.requiresBash) func testConcurrentSlowExitsDoNotHang() async throws {
-        // 16 concurrent slow-to-exit children is enough to flood the
-        // monitor and trigger the burst on Linux.
+    @Test(.timeLimit(.minutes(1)))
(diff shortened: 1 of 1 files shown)


iCharlesHu:
Please rebase to pick up https://github.com/swiftlang/swift-subprocess/pull/290

broken-circle:
@iCharlesHu Done, rebased onto `main`.

iCharlesHu (approved):

Issue #287, opened 2026-06-01
FreeBSD CI times out


broken-circle:
The FreeBSD CI job is timing out at the 6-hour limit on recent PRs (#282, #284, #285). The build completes in ~60s, so the hang is in the test suite, not compilation. The suite runs in parallel, so the timeout lands in a different test on each run.

- https://github.com/swiftlang/swift-subprocess/actions/runs/26690751131/job/78767563149?pr=282
- https://github.com/swiftlang/swift-subprocess/actions/runs/26722620295/job/78763346043?pr=284
- https://github.com/swiftlang/swift-subprocess/actions/runs/26728321726/job/78767382593?pr=285

The hanging runs are on `Swift version 6.3-dev (LLVM 972b62858355d07, Swift 3f8c798cfe25bbb)`.

#277 switched from `swift_package_test.yml@0.0.11` to `@main` to fix issues on Android, but I don't think this was the cause, since the `freebsd-build` job in `swift_package_test.yml` is unchanged between `0.0.11` and `main` (only diffs are for Android/Wasm). It installs the toolchain from `https://download.swift.org/tmp-ci-nightly/development/freebsd-14_ci_latest.tar.gz` and accepts only `nightly-main`, which might have a regression.

The FreeBSD step there might also benefit from a `timeout-minutes` like other steps.

cc @iCharlesHu @jakepetroules

iCharlesHu:
I'm working on a fix.

broken-circle:
Looks like #280 may resolve this.

iCharlesHu:
Resolved by https://github.com/swiftlang/swift-subprocess/pull/290

PR #290, opened 2026-06-02
Monitor process termination with kqueue instead of DispatchSource

iCharlesHu:
On FreeBSD, Subprocess spawns children with `pdfork(2)`. Per the man page such a child will not cause `SIGCHLD` on termination, which means `DispatchSource` does not work with processes spawned via `pdfork`.

Replace the `DispatchSource` implementation with a kqueue-based monitor shared by all kqueue platforms, mirroring the Linux epoll/pidfd backend. This change also allows us to completely remove the use of `Dispatch` from Subprocess.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+BSD.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+BSD.swift
@@ -11,19 +11,47 @@
 
 #if os(macOS) || os(FreeBSD) || os(OpenBSD)
 
-#if canImport(Darwin)
-import Darwin
-#elseif canImport(Glibc)
-import Glibc
-#endif
-
 #if canImport(System)
 import System
 #else
 import SystemPackage
 #endif
 
-internal import Dispatch
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#endif
+
+#if canImport(os)
+import os
+#endif
+
+import Synchronization
+
+#if canImport(Darwin)
+private typealias _Mutex = OSAllocatedUnfairLock
+#else
+private typealias _Mutex = Synchronization.Mutex
+#endif
+
+// Selects the kqueue filter and identifier used to watch a child for exit.
+//
+// macOS and OpenBSD spawn ordinary children, monitored by PID with
+// `EVFILT_PROC`. FreeBSD spawns children with `pdfork()` and monitors the
+// resulting process descriptor with `EVFILT_PROCDESC`. If `pdfork()` was
+// unavailable and the spawn fell back to `fork()` (so there is no descriptor),
+// it monitors by PID just like the other platforms.
+private func _monitorTarget(
+    for processIdentifier: ProcessIdentifier
+) -> (filter: Int16, identifier: CInt) {
+    #if os(FreeBSD)
+    if processIdentifier.processDescriptor != .invalidDescriptor {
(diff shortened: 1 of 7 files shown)


iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/287

Sources/Subprocess/Platforms/Subprocess+BSD.swift
```diff
@@ -11,19 +11,47 @@
...
+// macOS and OpenBSD spawn ordinary children, monitored by PID with
+// `EVFILT_PROC`. FreeBSD spawns children with `pdfork()` and monitors the
+// resulting process descriptor with `EVFILT_PROCDESC`. If `pdfork()` was
+// unavailable and the spawn fell back to `fork()` (so there is no descriptor),
+// it monitors by PID just like the other platforms.
+private func _monitorTarget(
+    for processIdentifier: ProcessIdentifier
+) -> (filter: Int16, identifier: CInt) {
+    #if os(FreeBSD)
+    if processIdentifier.processDescriptor >= 0 {
```
jakepetroules: nit: might be worth checking anywhere we use a sentinel value for an unset processDescriptor and make sure that's using -1 instead of 0

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -440,7 +440,7 @@ extension Configuration {
         errorPipe: consuming CreatedPipe
     ) async throws -> SpawnResult {
         // Ensure the waiter thread is running.
-        #if os(Linux) || os(Android)
+        #if os(Linux) || os(Android) || os(FreeBSD) || os(OpenBSD)
```
jakepetroules: #if !os(macOS) would be more future proof

jakepetroules: Also, the ifdef in SubprocessProcessMonitoringTests.init should probably now just become #if !os(Windows)

iCharlesHu: Actually we can drop `#if` here altogether.

Sources/Subprocess/Platforms/Subprocess+BSD.swift
jakepetroules: nit: could be `!= .invalidDescriptor here` too

Sources/Subprocess/Platforms/Subprocess+BSD.swift
```diff
@@ -41,19 +69,353 @@ internal func waitForProcessTermination(
...
+                            storage.kqueueFileDescriptor,
+                            &event,
+                            1,
+                            nil,
+                            0,
+                            nil
+                        )
+                        if rc != 0 {
+                            let capturedErrno = errno
+                            #if !os(FreeBSD)
```
jakepetroules: Don't you potentially still need this path on FreeBSD when not using a process descriptor? Should this add that the `processDescriptor == .invalidDescriptor` and remove the compile-time guard?

iCharlesHu: I think `filter == Int16(EVFILT_PROC)` itself should suffice. We only use this filter on non-FreeBSD because we don't have process file descriptor.

jakepetroules (approved):

PR #296, opened 2026-06-04
Drain buffered output on Windows after cancellation

### LINKED ISSUE #96: Subprocess dropping output
Summary:
The experimental Subprocess implementation drops part of the output in some cases

Steps To Reproduce:
1. Download attached programs
2. Run `./sample` (do `go build sample.go` first if you want to build it from source)
3. Run `swift run Repro`

Results:
When invoked directly the program outputs three lines:

% ./sample
          State            Type            Name       Start Time         End Time                             Frame In                            Frame Out                              Logs ID
        Started       Terrible1          Foobar  2024-09-30T08:29:13+09:00  2024-09-30T08:29:13+09:00                  0914390189048023984                                                                   abc-123
      Completed       Terrible2          Foobar  2024-09-30T08:29:13+09:00  2024-09-30T08:29:13+09:00                  0914390189048023984                                                                   abc-123

When invoked through Subprocess, the output only contains one line:

% swift run Repro
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'Repro' complete! (0.31s)
          State            Type            Name       Start Time         End Time                             Frame In                            Frame Out                              Logs ID

Seen on the main branch of https://github.com/iCharlesHu/swift-experimental-subprocess, commit 0f93295

Reported in 136937639

broken-circle:
Closes #96. The issue was already resolved on non-Windows platforms.

When a child exits, the run cancels its pending I/O via `cancelAsyncIO()`. On Windows, a read issued after that cancellation returned `nil` without consuming bytes already buffered in the pipe, so output that the child had already written could be dropped. The kqueue and epoll backends drain the descriptor on this path; the IOCP backend had no equivalent. A read or write reaching the cancelled path also issued an overlapped operation and returned without awaiting it, leaving the kernel to write into a buffer the call had already released.

`registerHandle()` now returns its `RegistrationOutcome` so `read()` and `write()` can branch on the cancelled state before issuing any I/O:

- `read()` drains what's buffered via `drainBufferedDataAfterCancellation()`, which polls with `PeekNamedPipe` and only issues a `ReadFile` when data is present, so it can't block waiting for an EOF that may never arrive. That read is registered with `bypassCancellation` and awaited through the completion port.
- `write()` reports a zero-length write, since the subprocess's standard input no longer has a reader.

Neither path issues an overlapped operation that it doesn't await.

`read()` drains on the in-flight cancellation exit as well as when the process was cancelled before the read began. The caller stops on the first `nil`, so that exit is the only place left to recover bytes that landed in the buffer with no read pending to receive them. The cost is one `PeekNamedPipe` at normal EOF, which returns `ERROR_BROKEN_PIPE` immediately.

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Windows.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Windows.swift
@@ -178,73 +178,220 @@ final class AsyncIO: @unchecked Sendable {
     private func registerHandle(
         _ handle: HANDLE,
         processIdentifier: ProcessIdentifier
-    ) -> SignalStream {
-        return SignalStream { (continuation: SignalStream.Continuation) in
-            switch self.ioCompletionPort {
-            case .success(let ioPort):
-                // Make sure thread setup also succeeded.
-                if case .failure(let error) = monitorThread {
-                    continuation.finish(throwing: error)
-                    return
-                }
-                let completionKey = UInt64(UInt(bitPattern: handle))
+    ) -> (stream: SignalStream, outcome: RegistrationOutcome) {
+        let (stream, continuation) = SignalStream.makeStream()
 
-                // Hold the lock across both the map insert and
-                // `CreateIoCompletionPort` so a concurrent `cancelAsyncIO`
-                // cannot mark the process cancelled between the two steps
-                // and leave behind a HANDLE that's bound to the IOCP but
-                // missing the cancellation marker.
-                let outcome: RegistrationOutcome = _registration.withLock { storage in
-                    let result = storage.register(
-                        completionKey: completionKey,
-                        continuation: continuation,
-                        processIdentifier: processIdentifier
-                    )
-                    switch result {
-                    case .alreadyCancelled:
-                        return .alreadyCancelled
-                    case .updated:
-                        // The handle was already attached to the IOCP by a
-                        // previous read or write; the new continuation
-                        // replaces the previous one in the map.
-                        return .registered
-                    case .registered:
-                        // Per the Windows documentation, the function
-                        // returns the handle of the existing I/O completion
-                        // port on success.
-                        guard
-                            CreateIoCompletionPort(
-                                handle, ioPort, completionKey, 0
-                            ) == ioPort
-                        else {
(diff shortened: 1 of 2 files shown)


jakepetroules (approved):

Sources/Subprocess/IO/AsyncIO+Windows.swift
```diff
@@ -177,74 +177,205 @@ final class AsyncIO: @unchecked Sendable {
...
+            return nil
+        }
+
+        // Data is present, so this read completes promptly. Force the
+        // registration through despite the cancellation and await it normally.
+        let (stream, outcome) = self.registerHandle(
+            handle,
+            processIdentifier: processIdentifier,
+            bypassCancellation: true
+        )
```
iCharlesHu: Instead of "bypassCancellation", which IMO is a bit hacky, can we simply issue one non-blocking read call to read the remaining buffer, since `PeekNamedPipe` says there are bytes ready to be read? In the unix code path, we simply call `read()` to drain the remaining bytes. Can we do that here?

broken-circle: Done. Removed `bypassCancellation`; `registerHandle()` surfaces its outcome and `read()` drains on the cancelled path instead of re-registering through the monitor.

The handle is `FILE_FLAG_OVERLAPPED`, so the drain is a `PeekNamedPipe`-gated overlapped `ReadFile` on a private event, not a bare synchronous read. The handle may already be on the IOCP from an earlier read (and Windows can't unbind it), so the low bit of `OVERLAPPED.hEvent` is set to keep this completion off the port and away from the monitor (documented opt-out on [`GetQueuedCompletionStatus`](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getqueuedcompletionstatus)).

broken-circle:
Looks like CI failed due to a flaky Linux test. I'm looking into fixing that test in a separate PR.

iCharlesHu (approved):
Thanks!!

PR #298, opened 2026-06-04
Escape .bat/.cmd arguments against cmd.exe injection on Windows

iCharlesHu:
On Windows, `CreateProcessW` runs a batch file by implicitly spawning `cmd.exe /c <command line>`. `cmd.exe` parses that line differently from `CommandLineToArgvW`, so the existing argument quoting was not enough: a metacharacter such as `&`, `|`, or `>` in an argument was treated by `cmd.exe` as a command separator/redirect, allowing command injection whenever a `.bat`/`.cmd` is run with untrusted arguments. This is the `BatBadBut` vulnerability class (CVE-2024-24576, CVE-2024-43402) that affected most language runtimes.

When the target is a batch file, Subprocess now invokes cmd.exe explicitly (resolved from the system directory) with hardening flags `/d /e:ON /v:OFF /c`, and escapes each argument both for `CommandLineToArgvW` (step 1) and for cmd.exe by ^-prefixing every metacharacter (step 2, per Colascione). Detection normalizes trailing dots/spaces that Windows strips when resolving paths (the CVE-2024-43402 bypass). Arguments containing `NUL/CR/LF`, and script paths containing quotation or ending in `\`, are rejected rather than escaped unsafely.

Non-batch targets are unaffected: `CreateProcessW` only ever appends .exe, so a bare name never silently resolves to a .bat. Adds Windows unit tests for the escaping/detection plus an end-to-end test that common injection payloads do not execute.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Windows.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Windows.swift
@@ -1001,10 +1001,11 @@ extension Configuration {
         ) = try self.generateWindowsCommandAndArguments(
             withPossibleExecutablePath: executablePath
         )
-        // Omit applicationName (and therefore rely on commandAndArgs
-        // for executable path) if we don't need to override arg0
+        // `generateWindowsCommandAndArguments` already decided whether to
+        // omit `applicationName` (and rely on `commandAndArgs` for the
+        // executable path), so pass its result through unchanged.
         return (
-            applicationName: self.arguments.executablePathOverride == nil ? nil : applicationName,
+            applicationName: applicationName,
             commandAndArgs: commandAndArgs,
             environment: environmentString,
             intendedWorkingDir: self.workingDirectory?.string
@@ -1240,7 +1241,7 @@ extension Configuration {
         // For Unix-style argv[0] override (where argv[0] differs from actual exe):
         // Set lpApplicationName = "C:\\path\\to\\real.exe"
         // Set lpCommandLine = "fake_name.exe arg1 arg2"
-        var args = self.arguments.storage.map {
+        let args = self.arguments.storage.map {
             guard case .string(let stringValue) = $0 else {
                 // We should never get here since the API
                 // is guarded off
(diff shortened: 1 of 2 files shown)


Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -1249,76 +1250,210 @@ extension Configuration {
...
+
+        if targetsBatchFile {
+            // The argv[0] override is not applicable to a batch file: `cmd.exe`
+            // sets %0 to the script path, so it is intentionally not applied.
+            //
+            // Resolve `cmd.exe` from the system directory so a directory earlier
+            // on PATH can't substitute a different interpreter; fall back to
+            // letting `CreateProcessW` resolve it if the system directory can't
+            // be determined.
+            let commandPrompt = Self.resolveCommandPromptPath() ?? "cmd.exe"
```
jakepetroules: should this just throw if it returns nil instead of falling back to "cmd.exe"?

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -1249,76 +1250,210 @@ extension Configuration {
...
+        // closing quote).
+        guard !scriptPath.contains("\""), scriptPath.last != "\\" else {
+            throw SubprocessError.spawnFailed(
+                withUnderlyingError: nil,
+                reason:
+                    "Cannot safely run batch file \"\(scriptPath)\": its path must not contain '\"' or end with '\\'."
+            )
+        }
+        var commandLine = "cmd.exe /d /e:ON /v:OFF /c \""
+        commandLine += "\"\(scriptPath)\""
```
jakepetroules, line 1345: shouldn't this do some more sophisticated escaping than just double quotes? what if the path contains a double quote?

iCharlesHu: The guard statement above already checks for quotation marks. But looking more, it also shouldn't contain `` etc., so I'll add that check too.

Sources/Subprocess/Platforms/Subprocess+Windows.swift
iCharlesHu: Good point. Will do.

jakepetroules (approved):

Issue #301, opened 2026-06-06
`Test Linux Kernel / al2-5.10` nondeterministically times out


broken-circle:
The `Test Linux Kernel / al2-5.10` job hit the 6-hour limit on the [CI run](https://github.com/swiftlang/swift-subprocess/actions/runs/27044483180/job/79827502221?pr=300#logs) for #300. `al2-4.18` passed on the same run in ~25 minutes and the rest of CI was green. Also seen in #338.

This job was added in #280, where `al2-5.10` most recently [passed](https://github.com/swiftlang/swift-subprocess/actions/runs/27035826008/job/79799446758?pr=280) in ~25 minutes. #300 only changes the shell script of a single test, so this looks like an intermittent failure (likely a race) rather than something #300 introduced.

I've opened #302 adding `timeout-minutes` to `test_linux_kernel` so a recurrence fails fast instead of burning the full 6 hours.

PR #302, opened 2026-06-06
Add a timeout to the `test_linux_kernel` job

broken-circle:
`test_linux_kernel` has no `timeout-minutes`, so a hang runs until GitHub's 6-hour default. On the [CI run](https://github.com/swiftlang/swift-subprocess/actions/runs/27044483180/job/79827502221?pr=300#logs) for #300 the `al2-5.10` cell hung and consumed the full 6 hours (see issue #301).

This PR adds a 60-minute limit (more than 2x the typical ~25 minute run), so it won't trip on a slow run but caps a hang at an hour.


iCharlesHu (approved):
I agree one hour timeout should be good.

PR #306, opened 2026-06-09
Cancel a stalled standard-input writer when the child process exits

iCharlesHu:
When `run()` feeds standard input from a source that can stall and the child process exits before consuming that input, `run()` could hang indefinitely. The input-writing task was parked awaiting its source rather than blocked inside a pipe write, so the existing `cancelAsyncIO` path (which only unblocks in-flight pipe I/O) could not release it, and the I/O task group never finished draining. Spawn a `waitForProcessTermination` waiter alongside the input-writing task that cancels the I/O group once the child exits.

Also add edge-case coverage for process I/O and AsyncIO, ported from `swift-async-process` and `swift-nio` and replace `FileDescriptor.close` with `_safelyClosed`.

### DIFF
--- a/Tests/SubprocessTests/IntegrationTests.swift
+++ b/Tests/SubprocessTests/IntegrationTests.swift
@@ -3527,6 +3527,170 @@ extension SubprocessIntegrationTests {
     }
 }
 
+// MARK: - Streaming, Identity, and Input File Descriptor Tests
+extension SubprocessIntegrationTests {
+    @Test func testStreamingInputAndOutput() async throws {
+        // Stream input and output through one live process concurrently: a
+        // writer task feeds lines into standard input while a reader task reads
+        // the echoes back, in separate tasks (as in `testInteractiveShell`) so
+        // neither blocks the other.
+        #if os(Windows)
+        // A PowerShell loop that echoes each stdin line back and flushes so the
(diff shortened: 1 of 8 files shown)


Tests/SubprocessTests/IntegrationTests.swift
```diff
@@ -3527,6 +3527,170 @@ extension SubprocessIntegrationTests {
...
+        let result = try await _run(
+            setup,
+            input: .fileDescriptor(inputFD, closeAfterSpawningProcess: false),
+            output: .string(limit: 256),
+            error: .discarded
+        )
+        #expect(result.terminationStatus.isSuccess)
+        #expect(result.standardOutput?.trimmingNewLineAndQuotes() == content)
+        // The parent retained ownership, so closing here must succeed. If
+        // Subprocess had wrongly closed it, this would throw `.badFileDescriptor`.
```
jakepetroules: well, it _might_ throw badFileDescriptor. Remember, there's no guarantee of anything if the owner of the file descriptor did not close it. It could have since been reopened and that fd# pointed to something else.

jakepetroules: Also, suggest using closeAfter

jakepetroules (approved):

Issue #286, opened 2026-06-01
SubprocessOutputSequence.strings(): Different code in README, better documentation, and naming


stefanspringer1:
1) Code in README differs:

README: `execution.standardOutput.lines()` (does not compile)
from the tests: `execution.standardOutput.strings()`

2) Better documentation

The documentation comment of `strings()` does not say if the separator will be contained in the strings. According to the corresponding test (`stdout += line`) it should to be the case. Please add this info to the documentation comment. (And if the separator is not contained in the strings, the test should maybe do something like `stdout += line; stdout += "\n"`.)

3) Naming

I like `lines()` more than `strings()` (compare `line in execution.standardOutput.strings()` to `line in execution.standardOutput.lines()`).

iCharlesHu:
Fixed in https://github.com/swiftlang/swift-subprocess/pull/310

PR #310, opened 2026-06-11
Docs: fix README to use strings(), document separator behavior

nenadvulic:
Refs #286.

**README:** several examples used `execution.standardOutput.lines()`, but the API is `strings(separatedBy:bufferingPolicy:)` / `strings(separatedBy:bufferingPolicy:as:)` — there is no `lines()`, so those snippets don't compile. This aligns them with the shipping API, including the parameterized form where the encoding is passed via `as:` (not `encoding:`):

```swift
for try await line in execution.standardOutput.strings(
    bufferingPolicy: .maxLineLength(1024),
    as: UTF16.self
) { ... }
```

**Doc comment:** `strings()` didn't state whether the separator is kept. Added a note that the separator characters are not included in the returned strings, matching the existing `StringSequence` type documentation.

Docs-only change. I've intentionally left the `strings()` → `lines()` rename proposed in #286 to the maintainers — happy to follow up with that if you'd like to go that way (in which case the README would move the other direction).

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ let result = try await run(
     error: .string(limit: 4096)
 ) { execution in
     var lineCount = 0
-    for try await _ in execution.standardOutput.lines() {
+    for try await _ in execution.standardOutput.strings() {
         lineCount += 1
     }
     return lineCount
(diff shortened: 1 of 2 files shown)


iCharlesHu (approved):
Thanks!

PR #312, opened 2026-06-12
Retry transient `EAGAIN` when spawning a subprocess

### LINKED ISSUE #156: stressTestWithLittleOutput fails nondeterministically with EAGAIN
✘ Test stressTestWithLittleOutput() recorded an issue at IntegrationTests.swift:1475:6: Caught error: Failed to spawn the new process with underlying error: UnderlyingError(rawValue: 11)
✘ Test stressTestWithLittleOutput() failed after 0.031 seconds with 1 issue.

This might be from execve returning EAGAIN due to exceeding system resource limits, but `ulimit` claims the max user process count in the GitHub Actions CI is unlimited, so...

broken-circle:
Closes #156.

The spawn path surfaces a transient `EAGAIN` from `clone3()`/`fork()` (Linux, Android, FreeBSD, OpenBSD) or `posix_spawn()` (Darwin) as a terminal `SubprocessError.spawnFailed`. `EAGAIN` from the fork side is a momentary per-uid or system task-count limit, not a permanent condition, so a single refusal is reported as failure.

This PR adds a bounded retry around the spawn: up to `maxSpawnAttempts` (currently `6`, which is the initial attempt plus 5 retries) with jittered exponential backoff (up to ~31 ms), after which the original error surfaces unchanged. Genuine exhaustion still surfaces in tens of milliseconds rather than hanging. Only a fork-side `EAGAIN` is retried, where the kernel created nothing so re-attempting is clean: on `fork()`/`clone3()` identified by `processDescriptor == .invalidDescriptor`, and on Darwin by the absence of pre-fork setup. An exec-side failure is left to the existing handling. Windows is excluded; its spawn failures are reported as Win32 errors, not `EAGAIN`.

The backoff runs in the async context between attempts, not in the shim, which holds a process-wide fork lock with signals blocked and runs on one shared worker thread; sleeping there would stall every other spawn.

Related: While in the spawn path I noticed two pre-existing cleanup gaps on the exec/prefork side, unrelated to this change. On `fork()`/`clone3()`, an exec failure with an errno outside `{ENOENT, EACCES, ENOTDIR}` throws without closing the `clone3()`-allocated `pidfd`; and on Darwin, the pre-fork path doesn't reap its child on exec failure. This PR's retry is fork-side only, so it neither touches nor amplifies either. I'll open a follow-up PR to fix these.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -732,6 +732,68 @@ extension Configuration {
             return self._errorReadEnd.take()
         }
     }
+
+    #if !os(Windows)
+    /// The maximum number of `fork`/`exec` attempts, including the first one,
+    /// before a transient, retryable spawn failure is surfaced to the caller.
+    private static let maxSpawnAttempts = 6
+    /// The base spawn-retry backoff in nanoseconds. The delay doubles each
+    /// attempt up to ``spawnRetryBackoffCapNanoseconds``, with jitter applied.
+    private static let spawnRetryBackoffBaseNanoseconds: UInt64 = 1_000_000
+    /// The maximum spawn-retry backoff in nanoseconds.
+    private static let spawnRetryBackoffCapNanoseconds: UInt64 = 100_000_000
+
+    /// A jittered, capped exponential backoff (in nanoseconds) for the given
+    /// spawn attempt.
+    internal static func spawnRetryBackoffNanoseconds(forAttempt attempt: Int) -> UInt64 {
+        let shift = UInt64(min(attempt - 1, 16))
+        let ceiling = min(
+            spawnRetryBackoffCapNanoseconds,
+            spawnRetryBackoffBaseNanoseconds << shift
+        )
+        // Equal jitter: half fixed, half random, to de-synchronize concurrent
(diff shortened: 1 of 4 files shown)


broken-circle:
The FreeBSD CI is failing inside the swift-system dependency, unrelated to this change:

```
[38/59] Compiling SystemPackage RawBuffer.swift
.../Constants.swift:635:3: error: cannot find 'O_CLOFORK' in scope
633 | internal var _O_CLOFORK: CInt {
634 |   #if !os(WASI) && !os(Linux) && !os(Android) && !canImport(Darwin)
635 |   O_CLOFORK
    |   `- error: cannot find 'O_CLOFORK' in scope
636 |   #else
637 |   0
```

The CI for this PR resolved to swift-system 1.7.0, while the more recent [CI run](https://github.com/swiftlang/swift-subprocess/actions/runs/27373301520/job/80890927050?pr=311) for #311 resolved to 1.6.5. Might need to pin this dependency if there was a regression.

See apple/swift-system#316. Fixed by apple/swift-system#319. CI should be all green on the next run.

iCharlesHu (approved):
Thanks!

PR #314, opened 2026-06-12
Register each file descriptor once with epoll / kqueue

iCharlesHu:
Persistently register each descriptor with epoll/kqueue and reuse it across reads and writes, detaching only on EOF, error, cancellation, or close:

- Switch epoll/kqueue to edge-triggered (EPOLLET / EV_CLEAR).
- Make read()/write() attempt the non-blocking syscall first and await a readiness signal only on EAGAIN.
- Detach in _safelyClose on Unix, mirroring Windows.

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Linux.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Linux.swift
@@ -31,6 +31,10 @@ import Musl
 import _SubprocessCShims
 import Synchronization
 
+// `EPOLLET` (edge-triggered) is the bit `1 << 31` in the Linux epoll ABI, but
+// the C symbol imports inconsistently across libraries.Define the bit directly
+// so it's uniform everywhere.
+private let _EPOLLET: UInt32 = 1 << 31
 private let _epollEventSize = 256
 private let _registration: Mutex<Registration> = Mutex(Registration())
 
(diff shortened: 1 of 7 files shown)


Sources/Subprocess/IO/AsyncIO+Linux.swift
```diff
@@ -352,18 +359,13 @@ final class AsyncIO: Sendable {
                 for registration in previousRegistrations {
                     toBeCancelled.append(registration.continuation)
 
-                    let rc = epoll_ctl(
+                    // Best-effort detach; ignore `ENOENT` (already gone).
```
jakepetroules: Why'd you remove the error handling? It does indeed seem like ENOENT would be the only type of error you'd want to ignore.

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -363,24 +390,33 @@ internal struct Registration {
...
+        }
+        if let existing = self.fileDescriptorMap[fileDescriptor] {
+            // The descriptor is already attached to the multiplexer. Finish the
+            // previous continuation so any stale awaiter unblocks, then replace
+            // it and reuse the existing attachment.
+            existing.continuation.finish()
+            self.fileDescriptorMap[fileDescriptor] = (continuation, processIdentifier)
+            return .updated
         }
         self.fileDescriptorMap[fileDescriptor] = (continuation, processIdentifier)
```
jakepetroules: nit: you could express this more semantically like:

```suggestion
        if let existing = self.fileDescriptorMap.updateValue((continuation, processIdentifier), forKey: fileDescriptor) {
            existing.continuation.finish()
            return .updated
        }
```

(only need to specify the key once for the 2 writes and 1 read)

Sources/Subprocess/IO/AsyncIO+Unix.swift
```diff
@@ -363,24 +390,33 @@ internal struct Registration {
...
         var fileDescriptorSet = self.processIdentifierMap[processIdentifier] ?? Set()
         fileDescriptorSet.insert(fileDescriptor)
         self.processIdentifierMap[processIdentifier] = fileDescriptorSet
```
jakepetroules: nit: can also be simplified:

```suggestion
        self.processIdentifierMap[processIdentifier, default: Set()].insert(fileDescriptor)
```

Sources/Subprocess/IO/AsyncIO+Linux.swift
iCharlesHu: Ahh good catch. Thanks!

jakepetroules (approved):

PR #317, opened 2026-06-23
Fix Windows `Environment.Key` `Comparable` and `Equatable` conformance

### LINKED ISSUE #316: `Environment.Key` `Comparable` is inconsistent with `Equatable` on Windows
**Description:**

`Environment.Key`'s `Equatable` conformance uses case-insensitive comparison on Windows, but its `Comparable` conformance always uses raw code-point order. This violates the `Comparable` contract, which requires that exactly one of `a == b`, `a < b`, or `b < a` is true for any two values. For `"PATH"` and `"path"` on Windows, both `a == b` and `a < b` hold simultaneously.

**Source:**

- `Equatable.==` https://github.com/swiftlang/swift-subprocess/blob/049ddf99c37c50411ce2997dc92a36eb74460b98/Sources/Subprocess/Configuration.swift#L611

- `Comparable.<` https://github.com/swiftlang/swift-subprocess/blob/049ddf99c37c50411ce2997dc92a36eb74460b98/Sources/Subprocess/Configuration.swift#L594

broken-circle:
Closes #316.

The `Equatable` conformance of `Environment.Key` on Windows used case-insensitive comparison (via `lowercased()`), while the `Comparable` conformance used case-sensitive comparison. This violated the `Comparable` contract, in which exactly one of `==`, `<`, or `>` must be true for any two values.

This PR adds `Environment.Key._comparisonValue` to centralize comparison logic, and uses it for equality, hashing, and ordering.

Note: I preserved the existing equality semantics using `String.lowercased()`. Windows treats environment variable names as [case-insensitive ordinal identifiers](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#ordinal-string-operations), whereas `lowercased()` is a linguistic (and locale-independent) fold. The two can differ only on non-ASCII names, which generally exclude environment variable names; so on real inputs, the two are equivalent and this PR's behavior is already correct. An ordinal fold would be a self-contained change to `_comparisonValue` that provides a fixed, Unicode-version-independent comparison key; I'll follow up with that unless we'd rather keep the current behavior.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -588,10 +588,20 @@ extension Environment.Key {
 
 extension Environment.Key: CodingKeyRepresentable {}
 
+extension Environment.Key {
+    /// The value used for equality, hashing, and ordering.
+    private var _comparisonValue: String {
+        #if os(Windows)
+        return self.rawValue.lowercased()
+        #else
+        return self.rawValue
+        #endif
+    }
+}
+
 extension Environment.Key: Comparable {
     public static func < (lhs: Self, rhs: Self) -> Bool {
-        // Even on windows use a stable sort order.
-        lhs.rawValue < rhs.rawValue
+        lhs._comparisonValue < rhs._comparisonValue
     }
 }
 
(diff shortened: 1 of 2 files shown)


broken-circle:
Linux tests are failing due to the known compiler issue which c5e19e0 in #314 temporarily fixes with a workaround. Once that's merged, these tests should pass.

jakepetroules (approved):

iCharlesHu:
Please rebase to pick up https://github.com/swiftlang/swift-subprocess/commit/ab6ac74d1722232122245b71494d80732c1b93bc

broken-circle:
@iCharlesHu Sure thing, done! I'll rebase the other recent PRs too.

iCharlesHu (approved):
Thanks!!

PR #319, opened 2026-06-23
Prevent process-descriptor and zombie leaks on spawn exec failure

broken-circle:
Follow-up to #312.

### Leaked process descriptor on exec-failure spawn errors (Linux/Android/FreeBSD)

A `clone3()`/`fork()` that succeeds and then fails at `execve()` still produces a process descriptor (`clone3(CLONE_PIDFD)`, its `pidfd_open()` fallback, or `pdfork()`). `_subprocess_fork_exec()` reaps the child but returns that descriptor open, and the `spawnFailed` branch in `spawn()` closed only the stdio pipes, leaking it. (The `ENOENT`/`EACCES`/`ENOTDIR` branch already closes it before moving to the next candidate path.)

This PR closes it on that branch, after `safelyCloseMultiple()` and guarded on a valid descriptor, best-effort (`try?`, mirroring `ProcessIdentifier.close()`) so the spawn error remains the thrown error. The close is Swift-side rather than in the shim because the `EAGAIN` retry from #312 distinguishes a fork-side failure (no descriptor) from an exec-side one (descriptor present) by that value; clearing it in the shim would defeat the discriminator.

### Unreaped prefork child on exec failure (Darwin)

On Darwin, `_subprocess_spawn_prefork()` forks and execs via `posix_spawn(POSIX_SPAWN_SETEXEC)`. When setup or exec fails the child `_exit()`s, but the parent returned the error without reaping it, leaving a zombie. Unlike Linux, there is no `SIGCHLD` reaper, and the Swift spawn-failure path does not reap either. The plain path delegates to `posix_spawn()`, which reaps its own child, so only the manual-fork path is affected.

This PR reaps the child on the failure return, matching `_subprocess_fork_exec()`, and hoists `_subprocess_reap_pid()` above both spawn blocks under `TARGET_OS_MAC || TARGET_OS_UNIX` so the Darwin block can use it.

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Unix.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Unix.swift
@@ -532,6 +532,21 @@ extension Configuration {
                         errorRead: errorReadFileDescriptor,
                         errorWrite: errorWriteFileDescriptor
                     )
+                    // A fork/clone3 that succeeded but failed to execve still
+                    // produced a process descriptor. The fork-vs-exec retry
+                    // discriminator has already run on this outcome, so closing
+                    // it here cannot make the exec-side failure look fork-side.
+                    // Unlike the branch above, close best-effort (mirroring
+                    // ProcessIdentifier.close()) so the spawn error remains as
+                    // the thrown error. Descriptor-less fallbacks returned
+                    // .invalidDescriptor and need no close. This comes after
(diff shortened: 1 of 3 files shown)


jakepetroules (approved):

iCharlesHu (approved):

iCharlesHu:
Thanks for the fix!

PR #321, opened 2026-06-26
Remove `_IODescriptorBox` workaround

broken-circle:
The SIL verifier crash in swiftlang/swift#89932 has been fixed by swiftlang/swift#89958, so the workaround added in #314 can be removed. Verified locally by building the Subprocess target with an assert-enabled toolchain built from current `main` (swiftlang/swift/main@ec5fe00cba4), which includes the fix.

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -312,10 +312,7 @@ public func run<
                 var diskIO = outputIOBox.take()
                 try diskIO?.safelyClose()
             } else {
-                // Hand the descriptor to the child task through a reference box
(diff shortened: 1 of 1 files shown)


broken-circle:
This is probably safe to merge as-is, since this was a 6.5-dev-only regression and the fix it relies on (swiftlang/swift#89958) is already on `main`. For extra safety we could wait until swiftlang/swift#90016 lands and a nightly snapshot includes it, then re-run CI. That PR re-enables the `debug_value` type-chain verifier that's currently disabled on `main`.

jakepetroules (approved):

iCharlesHu (approved):
Thanks!

PR #326, opened 2026-06-28
Adopt SE-0456 `Span` and `Data` accessors

broken-circle:
[SE-0456](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0456-stdlib-span-properties.md) added `Span`-providing properties to standard library types; Foundation also added `Data.bytes` and `Data.span` accordingly. These additions subsume the need for the `Data.bytes` and `Span._bytes` shims.

The only caller of `Data.bytes` is `AsyncIO.write(_:to:for:)`, which now binds to Foundation's accessor. The removed accessor also escaped the pointer vended by `withUnsafeBytes(_:)`; Foundation's borrowing accessor avoids this. `DataProtocol.bytes` had no callers and is removed as dead code.

The only caller of `Span._bytes` is `run(_:input:output:error:)`, which now uses `input.bytes`. `Buffer.bytes` duplicated the hand-rolled `RawSpan` construction and now delegates to `Array._bytes`, leaving a single implementation.

`Array._bytes` remains because `Array.span` does not back-deploy below macOS 26, but Subprocess deploys to macOS 13. This can move to `self.span.bytes` once the deployment floor reaches macOS 26.

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -188,7 +188,7 @@ public func run<
         output: output,
         error: error
     ) { execution in
-        _ = try await execution.standardInputWriter.write(input._bytes)
+        _ = try await execution.standardInputWriter.write(input.bytes)
         try await execution.standardInputWriter.finish()
     }
 }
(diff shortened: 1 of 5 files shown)


iCharlesHu (approved):
Thanks!

PR #328, opened 2026-06-30
Spelling, wording, and grammar updates

heckj:
This updates the doc comments to fix a few spelling, grammar, and punctuation issues - but more generally breaks up and edits some of the content so that abstracts are simple and direct, and don't include monospaced text (aka "code voice") or symbol references. In a few cases, I just dropped the symbol references because they'll be obvious in the signature in the DocC output. In other cases, I broke up the sentence and pushed references to a discussion section where code voice and symbol links are more effective. I also tried to make sure that all the abstracts were a single sentence, followed by blank link so that any following content would appear in the discussion/overview section for that symbol, and not "explode" the abstract into a multi-line thing.

(We generally avoid putting code voice and symbols in abstracts because he abstracts appear in multiple locations within DocC content - in navigation areas and such, where the extra linking either just gets dropped or looks very awkward)

I went ahead and extended this to remove the biological or family metaphors from the documentation content (such as "spawn", "kill", "children", "parent") to use more neutral/agnostic terms. I've also rebased this over main to take advantage of the DocC catalog that was added in #329

I'll follow up with another PR to add more curation (organization, including the types) - and try to expand out the usage scenarios based on what I've seen talked about in threads on the forums and discussions around this overall API as it's developed. Nothing too extensive, just trying to get a baseline of "how to use it" and considerations in place.

/cc @invalidname

### DIFF
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -162,8 +162,7 @@ public func run<
 
 // MARK: - Configuration Based
 
-/// Runs a configuration asynchronously and returns an ``ExecutionResult`` that
-/// contains the output of the child process.
+/// Runs a subprocess with the given configuration and returns the collected output and exit status.
 ///
 /// - Parameters:
 ///   - configuration: The configuration to run.
@@ -193,8 +192,7 @@ public func run<
     }
 }
 
-/// Runs a ``Configuration`` asynchronously and returns
-/// an ``ExecutionResult`` that contains the output of the child process.
+/// Runs a subprocess with the given configuration and returns the collected output and exit status.
 ///
 /// - Parameters:
 ///   - configuration: The configuration to run.
@@ -217,8 +215,7 @@ public func run<
     }
 }
 
-/// Runs a ``Configuration`` asynchronously and lets a closure manage the running
-/// subprocess.
+/// Runs a subprocess with the given configuration and lets a closure interact with it while it runs.
 ///
 /// Use this overload when you need to interact with the subprocess while it runs,
 /// such as streaming its standard output, writing to its standard input, or sending
--- a/Sources/Subprocess/API.swift
+++ b/Sources/Subprocess/API.swift
@@ -18,7 +18,7 @@ public import SystemPackage
 // MARK: - Collected Result
 
 /// Runs an executable asynchronously and returns the collected output
-/// of the child process.
+/// of the subprocess.
 ///
 /// - Parameters:
 ///   - executable: The executable to run.
@@ -60,7 +60,7 @@ public func run<
 }
 
 /// Runs an executable asynchronously and returns the collected output
-/// of the child process.
+/// of the subprocess.
 ///
 /// - Parameters:
 ///   - executable: The executable to run.
@@ -125,7 +125,7 @@ public func run<
 ///     an ``Execution`` value that's valid only for the duration of the call.
 ///     Don't let the execution value escape the closure.
 /// - Returns: An ``ExecutionResult`` that contains the closure's return value and
-///   the termination status of the child process.
+///   the termination status of the subprocess.
 public func run<
     Result: ~Copyable,
     Input: InputProtocol,
@@ -233,7 +233,7 @@ public func run<
 ///     an ``Execution`` value that's valid only for the duration of the call.
 ///     Don't let the execution value escape the closure.
 /// - Returns: An ``ExecutionResult`` that contains the closure's return value and
-///   the termination status of the child process.
+///   the termination status of the subprocess.
 public func run<
     Result: ~Copyable,
     Input: InputProtocol,
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -41,20 +41,17 @@ public struct PlatformOptions: Sendable {
     public var qualityOfService: QualityOfService = .default
     /// The user ID for the subprocess.
     public var userID: uid_t? = nil
-    /// The real and effective group ID and the saved
-    /// set-group-ID of the subprocess, equivalent to calling
-    /// `setgid()` on the child process.
+    /// The real, effective, and saved set-group-ID for the subprocess.
     ///
-    /// The group ID controls permissions, particularly
-    /// for file access.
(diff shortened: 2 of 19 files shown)


Sources/Subprocess/API.swift
```diff
@@ -425,15 +422,15 @@ private struct _RunOutcome<
     let error: Error.OutputType
 }
 
-/// A reference box that carries a (noncopyable, non-`Sendable`)
+/// A reference box that carries a (non-Copyable, non-Sendable)
```
heckj: @invalidname I wasn't entirely sure what we're using when we talk about non copyable (~Copyable) and non sendable (~Sendable) types - and how we reference it in prose/narrative text. I've seen `noncopyable` in the community discussion generally. This was inconsistent one way or the other - so it seems like it needs a smidge of tweaking, but I'm not clear on which way to go here. Suggestions?

broken-circle: [SE-0390](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0390-noncopyable-structs-and-enums.md) (Noncopyable structs and enums) and [SE-0427](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0427-noncopyable-generics.md) (Noncopyable Generics) use  "noncopyable", while [SE-0518](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0518-tilde-sendable.md) (`~Sendable` for explicitly marking non-`Sendable` types) uses "non-`Sendable`". These might be as close to canonical as we'll find.

invalidname: Conversationally, like in the Swift Language Workgroup meetings I take notes for, we just say "noncopyable" and "nonsendable". Since this is going to be rendered alongside its actual signature, would you consider writing both of them in unstyled prose?

I'd be interested in @amartini51's thoughts about this too.

```suggestion
/// A reference box that carries a noncopyable, nonsendable
```

Also, as a `private` type, would rendered docs typically do anything with this doc comment anyways?

Sources/Subprocess/API.swift
```diff
@@ -425,15 +422,15 @@ private struct _RunOutcome<
...
-/// language-level way to move a noncopyable into an escaping closure is to
+/// language-level way to move a non-Copyable into an escaping closure is to
 /// capture a `var IODescriptor?` and `take()` it inside, but that captured
 /// `var` lowers to a `${ var Optional<IODescriptor> }` box whose debug info
 /// crashes the SIL verifier on some nightly toolchains: https://github.com/swiftlang/swift/issues/89932
-/// Holding the descriptor behind a class reference keeps that box off the closure's frame. the
+/// Holding the descriptor behind a class reference keeps that box off the closure's frame. The
 /// descriptor lives in heap storage and the task drains it via `take()`.
 ///
 /// TODO: Remove when https://github.com/swiftlang/swift/issues/89932 is fixed.
```
invalidname: Do we want the `TODO` as part of the documentation? Because that's what the `///` is going to do.

We can't demote the `TODO` to double-slash in place, because that would break off the triple-slash doc comment from the `_IODescriptorBox` symbol. I'd suggest converting this to double-slash and moving it either above the entire doc comment, or inside the implementation. Neither is ideal, I'll admit.

Sources/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -41,20 +41,17 @@ public struct PlatformOptions: Sendable {
...
     public var groupID: gid_t? = nil
     /// The list of supplementary group IDs for the subprocess.
     public var supplementaryGroups: [gid_t]? = nil
-    /// The process group for the subprocess, equivalent to
-    /// calling `setpgid()` on the child process.
+    /// The process group for the subprocess.
     ///
-    /// The process group ID groups related processes for
-    /// controlling signals.
+    /// Equivalent to calling `setpgid()` on the child process.
```
invalidname: We could make this a complete sentence.
```suggestion
    /// This is equivalent to calling `setpgid()` on the child process.
```

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -592,7 +593,7 @@ internal func reapProcess(
 // MARK: - Subprocess Control
 
 extension Execution {
-    /// Terminate the current subprocess with the given exit code
+    /// Terminate the current subprocess with the given exit code.
```
invalidname:
```suggestion
    /// Terminates the current subprocess with the given exit code.
```

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -621,7 +622,7 @@ extension Execution {
         }
     }
 
-    /// Suspend the current subprocess
+    /// Suspend the current subprocess.
```
invalidname:
```suggestion
    /// Suspends the current subprocess.
```

Sources/Subprocess/Platforms/Subprocess+Windows.swift
```diff
@@ -645,7 +646,7 @@ extension Execution {
         }
     }
 
-    /// Resume the current subprocess after suspension
+    /// Resume the current subprocess after suspension.
```
invalidname:
```suggestion
    /// Resumes the current subprocess after suspension.
```

invalidname (changes requested):
Nice fixes, thanks. There's a `TODO` in a doc comment that gave me pause; maybe we could take the opportunity to not do that.

Sources/Subprocess/API.swift
broken-circle: Just for context, `_IODescriptorBox` is pending removal in #321, so it's self-resolving once that lands.

Sources/Subprocess/API.swift
heckj: as a `private` symbol, it'll be hidden - I'm primarily asking for consistency

Sources/Subprocess/API.swift
heckj: resolved with the rebase and cleanup

iCharlesHu (approved):
Thank you so much @heckj !

iCharlesHu:
@invalidname I believe this one is ready to be merged.

invalidname (approved):
LGTM. Thanks for the fixes.

Issue #324, opened 2026-06-27
Documentation improvements for 1.0


broken-circle:
Subprocess is moving toward a stable 1.0, and user-facing documentation is an important part of that. I recently opened two PRs (#322, #323) related to broken/missing symbol links, and had some general documentation-related thoughts that I think are worth discussing before I open PRs for them, since they're not necessarily correctness issues.

## Protocol requirements

Some APIs may be under-documented. For example:

```swift
/// An output type that collects the subprocess's output as a `[UInt8]` array.
public struct BytesOutput: OutputProtocol, ErrorOutputProtocol {
    /// Creates an array from a `RawSpan`.
    public func output(from span: RawSpan) throws -> [UInt8] {
        span.withUnsafeBytes { Array($0) }
    }
}
```

`output(from:)` is a requirement of `OutputProtocol`, so it's understandable that we may not want to repeat the documentation here. However, a user landing on this page of the documentation doesn't particularly care about that; they just see a page with no explanation of the parameter, return type, or error-throwing behavior. The same applies to other extension-point requirements, such as `InputProtocol.write(with:)`.

One option is to remove the documentation from the witness (`BytesOutput.output(from:)`) and let it inherit the protocol's documentation via DocC. I confirmed locally that removing the comment causes the `BytesOutput.output(from:)` page to render the protocol's documentation. This gives a single source of truth at the cost of identical prose on every witness page, and would mean expanding the documentation on the requirement itself; in this case, `output(from:)` and `maxSize`, whose 128 KB default is currently undocumented.

The alternative is bespoke documentation on each witness, which keeps type-specific prose (for example, "Creates an array from a `RawSpan`") at the cost of duplication and drift.

The two approaches aren't mutually exclusive. For example, several `InputProtocol` witnesses already repeat the same one-line comment that inheritance would deduplicate, whereas `CustomWriteInput.write(with:)` is a no-op that deserves bespoke documentation.

I lean toward inheriting by default and reserving bespoke documentation for witnesses whose behavior differs.

## Landing page

Subprocess currently has no DocC catalog, so the generated documentation opens directly on the Topics section with no overview. A curated landing page would give users an entry point that frames what the library is and how to approach it.

[...]

PR #329, opened 2026-06-30
Add a DocC catalog with task-based topics

broken-circle:
Partially addresses #324.

Subprocess doesn't have a documentation catalog yet, so the generated documentation opens on the automatic topics section, grouping public symbols by language category, with no overview to frame the library.

This PR adds a catalog with a curated landing page containing an overview and a topics section with public top-level APIs organized by task. A few choices I made when organizing it:

- Curate exactly the same top-level symbols that automatic curation already surfaces, except now grouped by task.
    - This leaves nested public types (for example, `SubprocessOutputSequence.Buffer`) as nested. They're accessible through their parent API pages, the same as automatic curation. (Keeps this PR focused on pure regrouping; if we want to surface these, I can open a follow-up PR.)
- Order the sections to follow the lifecycle of running a subprocess.
- Within each section, lead with the protocol or aggregate type, then list concrete types from simplest to most specialized, with `SubprocessFoundation`-gated types toward the end.

### DIFF
--- a/Sources/Subprocess/Documentation.docc/Subprocess.md
+++ b/Sources/Subprocess/Documentation.docc/Subprocess.md
@@ -44,7 +44,7 @@ sequence of buffers you can iterate directly or read line by line. The
 ``Execution``, ``SubprocessOutputSequence``, and ``StandardInputWriter`` values 
 are valid only for the duration of the body closure; don't let them escape it.
 
-If the task running a subprocess is cancelled, Subprocess can run a configurable 
+If the task running a subprocess is canceled, Subprocess can run a configurable 
 teardown sequence (for example, a graceful shutdown followed by a forced kill) 
 before the call returns. You describe that sequence with ``TeardownStep`` values, 
 and you can trigger one yourself from the body closure.
--- a/Sources/Subprocess/Documentation.docc/Subprocess.md
+++ b/Sources/Subprocess/Documentation.docc/Subprocess.md
@@ -27,7 +27,7 @@ take a trailing closure: the closure receives an ``Execution`` value you use to
 write to standard input, stream standard output and standard error, and signal 
 or tear down the process.
 
-There are three independent parameters for the input, output, and errors, so a 
+There are three independent parameters for input, output, and errors, so a 
 single call can mix collecting and streaming. Input, output, and errors conform 
 to ``InputProtocol``, ``OutputProtocol``, or ``ErrorOutputProtocol``, 
 respectively.
--- a/Sources/Subprocess/Documentation.docc/Subprocess.md
+++ b/Sources/Subprocess/Documentation.docc/Subprocess.md
@@ -1,0 +1,128 @@
+# ``Subprocess``
+
+Subprocess is a cross-platform Swift package for spawning child processes, 
+built from the ground up with Swift concurrency.
+
+## Overview
+
+Subprocess centers on a family of `run` functions. Each spawns an executable, 
+waits for it to terminate, and returns an ``ExecutionResult`` carrying the 
+process identifier, the ``TerminationStatus``, and whatever output you asked it 
+to collect. The simplest form runs a command and collects its standard output:
+
+```swift
+import Subprocess
+
+let result = try await run(.name("ls"), output: .string(limit: 4096))
+
+print(result.processIdentifier) // e.g. 1234
+print(result.terminationStatus) // e.g. exited(0)
+print(result.standardOutput)    // e.g. Optional("LICENSE\nPackage.swift\n...")
+```
+
+You can run an executable directly, supplying its ``Arguments``, ``Environment``, 
+and working directory inline, or build a reusable ``Configuration`` and run that. 
+When you need to interact with the process while it runs, use the overloads that 
+take a trailing closure: the closure receives an ``Execution`` value you use to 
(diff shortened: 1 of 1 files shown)


broken-circle:
cc @heckj

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -0,0 +1,126 @@
...
+If the task running a subprocess is cancelled, Subprocess can run a configurable 
+teardown sequence (for example, a graceful shutdown followed by a forced kill) 
+before the call returns. You describe that sequence with ``TeardownStep`` values, 
+and you can trigger one yourself from the body closure.
+
+Platform-specific settings live on ``PlatformOptions``: user and group 
+identifiers and session behavior on Unix, quality of service on Darwin, and 
+console and window behavior on Windows. The `SubprocessFoundation` trait, 
+enabled by default, adds the `Data`-based input and output types; disable it to 
+drop the dependency on Foundation.
```
invalidname: Does this refer to Apple Foundation or the OSS swift-foundation?

invalidname (approved):
Nice overview. Thanks for taking care of this. Have you done a local build of the documentation on this branch (with command line `docc` or Product -> Build Documentation in Xcode) to make sure the links all work? (You can reach out to me if you need help doing so.)

broken-circle:
@invalidname Thanks for the review! Yeah the symbol links all resolve. I ran `swift package generate-documentation --target Subprocess --warnings-as-errors` on this branch without errors, and also Product -> Build Documentation in Xcode as a double-check.

Sources/Subprocess/Documentation.docc/Subprocess.md
broken-circle: Both, depending on platform. The `Data`-based types compile under `#if SubprocessFoundation`, and where they do they `import Foundation`: the system Foundation on Darwin and `FoundationEssentials` (swift-foundation) elsewhere. Disabling the trait drops whichever applies. I reworded the sentence to say that rather than leave "Foundation" ambiguous.

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -0,0 +1,128 @@
...
+and working directory inline, or build a reusable ``Configuration`` and run that. 
+When you need to interact with the process while it runs, use the overloads that 
+take a trailing closure: the closure receives an ``Execution`` value you use to 
+write to standard input, stream standard output and standard error, and signal 
+or tear down the process.
+
+There are three independent parameters for the input, output, and errors, so a 
+single call can mix collecting and streaming. Input, output, and errors conform 
+to ``InputProtocol``, ``OutputProtocol``, or ``ErrorOutputProtocol``, 
+respectively.
```
heckj:
```suggestion
There are three independent parameters for input, output, and errors, so a 
single call can mix collecting and streaming. Input, output, and errors conform 
to ``InputProtocol``, ``OutputProtocol``, or ``ErrorOutputProtocol``, 
respectively.
```

The article "the" on only on the three there just felt really wrong. Seemed easier perhaps to drop it for what you're stating here, although we could also add another "the" before "output" and "errors" for consistency.

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -0,0 +1,128 @@
...
+file descriptor, or stream it. Passing ``CombinedErrorOutput`` for the error 
+parameter merges standard error into standard output, the equivalent of the 
+shell's `2>&1`.
+
+Streamed output arrives as a ``SubprocessOutputSequence``, an asynchronous 
+sequence of buffers you can iterate directly or read line by line. The 
+``Execution``, ``SubprocessOutputSequence``, and ``StandardInputWriter`` values 
+are valid only for the duration of the body closure; don't let them escape it.
+
+If the task running a subprocess is cancelled, Subprocess can run a configurable
```
heckj:
```suggestion
If the task running a subprocess is canceled, Subprocess can run a configurable 
```

I make this mistake all the time - cancelled is British English, and the docs all lean into American English - so changing the spelling to align

heckj (commented):
Looks amazing @broken-circle - love it! I have two small (very small!) nits in wording as I read through it

heckj:
I also did a local run with preview-documentation, sharing the output
Screenshot 2026-06-30 at 8 33 33 AM

Sources/Subprocess/Documentation.docc/Subprocess.md
broken-circle: Yeah, generally I prefer the American English spelling too. I think Subprocess leans into the British English spelling because the standard library chose to spell the error type as [`CancellationError`](https://developer.apple.com/documentation/swift/cancellationerror). This type is used in the library, but isn't explicitly mentioned in any user-facing documentation.

If we do decide to go with the American English spelling, should we change it everywhere else as part of #328? There are some other instances floating around.

heckj: For consistency, yeah probably - most of the places the other uses sit are internal or private types, so it's not as externally visible. The one in Configuration.swift (line 1244) is the only public one that stood out to me other than this article.

PR #332, opened 2026-06-30
Validate environment variables before spawning

iCharlesHu:
Reject environment keys and values that would be malformed or silently corrupted when handed to the child process, surfacing a clear SubprocessError before the child is created rather than spawning it with a broken environment.

POSIX (Subprocess+Unix.swift):
- Reject keys containing '=' or NUL, and keys beginning with a digit.
- Reject values containing NUL.

Windows (Subprocess+Windows.swift):
- Reject keys containing '=' or NUL and values containing NUL; an embedded NUL would truncate the null-separated environment block passed to CreateProcessW.
- Reject serialized "key=value" entries exceeding 32,767 UTF-16 code units, counting the '=' and null terminator, to match the documented Windows limit (measured in UTF-16 code units, not grapheme clusters).

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -205,7 +205,7 @@ extension Configuration {
         var outputPipeBox: CreatedPipe? = consume outputPipe
         var errorPipeBox: CreatedPipe? = consume errorPipe
 
-        return try await self.preSpawn { args throws -> SpawnResult in
+        func spawnFunc(_ args: PreSpawnArgs) async throws -> SpawnResult {
             let (env, uidPtr, gidPtr, supplementaryGroups) = args
             var _inputPipe = inputPipeBox.take()!
             var _outputPipe = outputPipeBox.take()!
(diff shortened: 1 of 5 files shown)


compnerd:
I think that this doesn't work for Windows; `=` is permissible in Windows environment blocks. `=C:=C:\` is used as a hidden variable for the current directory on a drive.

iCharlesHu:
> I think that this doesn't work for Windows; `=` is permissible in Windows environment blocks. `=C:=C:\` is used as a hidden variable for the current directory on a drive.

I believe those are reserved / used by the system and only set by the system, and users are not supposed to set them (please correct me if I'm wrong). Here we are validating that the user-supplied dictionary cannot contain '=', which does not affect the ones set by the system. 

Microsoft documentation https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables clearly states:

> The name of an environment variable cannot include an equal sign (=).

Therefore I think it's right for Subprocess to enforce this rule.

@compnerd do you expect users to be able to set environment variables like

```
try await run(..., environment: .inherit.updating(["=C:" : "=C:\"])
```

?

compnerd:
> hose are reserved / used by the system and only set by the system

This is correct.

jakepetroules:
But we might still need to pass such env vars _through_ the API, so maybe we need to relax the = validation?

iCharlesHu:
> But we might still need to pass such env vars _through_ the API, so maybe we need to relax the = validation?

@jakepetroules do you have any use case in mind? I decided to enforce this one because it seems both POSIX and Windows are pretty clear on not allowing `=`:

> These strings have the form name=value; names shall not contain the character `=`.  

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

> The name of an environment variable cannot include an equal sign (=).  

https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables

Just to be clear, we are only enforcing key cannot have `=` and you can have `=` in the value. I can't find a valid reason where someone might want to use `=` in the environment key on purpose. 

The other reason I think this is worth enforcing is because both `posix_spawn` and `exec` will happily take these "malformed" environment strings and just silently set incorrect environment key-value pairs, which could lead to unpredictable behavior. IMO it's much better to fail loudly in this case.

jakepetroules:
Let's go with this for now and if we find we need to allow it for some edge case, we can address it then.

jakepetroules (approved):

Issue #330, opened 2026-06-30
FreeBSD CI nondeterministically times out


broken-circle:
FreeBSD CI has recently started nondeterministically timing out after 6 hours during the test phase. Observed on failing CI runs for documentation-only PRs:

#328: [run 1](https://github.com/swiftlang/swift-subprocess/actions/runs/28410785582/job/84183107703?pr=328), [run 2](https://github.com/swiftlang/swift-subprocess/actions/runs/28456012733/job/84330876269?pr=328)
#329: [run 1](https://github.com/swiftlang/swift-subprocess/actions/runs/28456762132/job/84333576338?pr=329)

It seems to reproduce fairly frequently. Issue #287 previously reported a deterministic failure that was fixed by #290.

PR #333, opened 2026-07-01
Annotate `pthread_t` out-parameter nullability on musl

broken-circle:
On musl, `pthread_t` is a pointer `typedef`, so the `ptr` out-parameter of `_subprocess_pthread_create()` left its inner pointer unannotated and tripped `-Wnullability-completeness` on the static Linux SDK build:

```shell
In file included from /__w/swift-subprocess/swift-subprocess/Sources/_SubprocessCShims/process_shims.c:23:
/__w/swift-subprocess/swift-subprocess/Sources/_SubprocessCShims/include/process_shims.h:48:5: warning: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified) [-Wnullability-completeness]
   48 |     pthread_t * _Nonnull ptr,
      |     ^
/__w/swift-subprocess/swift-subprocess/Sources/_SubprocessCShims/include/process_shims.h:48:5: note: insert '_Nullable' if the pointer may be null
   48 |     pthread_t * _Nonnull ptr,
      |     ^
      |               _Nullable
/__w/swift-subprocess/swift-subprocess/Sources/_SubprocessCShims/include/process_shims.h:48:5: note: insert '_Nonnull' if the pointer should never be null
   48 |     pthread_t * _Nonnull ptr,
      |     ^
      |               _Nonnull
```

This PR extends the existing platform condition that already annotates this pointer on the other pointer-typed `pthread_t` platforms (Darwin, FreeBSD, OpenBSD) to also cover musl. The condition is factored into a `TARGET_LIBC_MUSL` macro so that the declaration and definition, which must carry it identically, share a single source of truth.

Note: I'm working on a separate PR to clear other warnings like `OpaquePointer` `Sendable` conformance.

### DIFF
--- a/Sources/_SubprocessCShims/include/process_shims.h
+++ b/Sources/_SubprocessCShims/include/process_shims.h
@@ -37,12 +37,22 @@
 #include <sys/wait.h>
 #endif // TARGET_OS_LINUX || TARGET_OS_FREEBSD
 
+// musl provides no identifying macro, so it is spelled as Linux that is
+// neither glibc nor Bionic. Correct only after a libc header (<pthread.h>
+// above) has put __GLIBC__ in scope, which is why this lives here and not in
+// target_conditionals.h.
+#if TARGET_OS_LINUX && !defined(__GLIBC__) && !defined(__ANDROID__)
+#define TARGET_LIBC_MUSL 1
+#else
+#define TARGET_LIBC_MUSL 0
+#endif // TARGET_LIBC_MUSL
+
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


broken-circle:
One Windows CI job is failing, but these changes don't affect Windows. From the logs, it looks like the issue is unrelated to Subprocess code. It should pass on a subsequent run.

FreeBSD CI is going to time out at the 6 hour mark, unrelated to this change (reported in #330).

iCharlesHu (approved):

PR #335, opened 2026-07-01
Migrate CI from Amazon Linux 2 to Amazon Linux 2023

broken-circle:
Amazon Linux 2 reached end of life on 2026-06-30, and Swift.org has [stopped publishing](https://forums.swift.org/t/dropping-support-for-amazon-linux-2/87920) the nightly toolchains for main and release/6.4.x. The static Linux SDK build on the amazonlinux2 host now fails.

This PR drops Amazon Linux 2 from CI and also removes the now-dead yum branch from the pre-build command, since Amazon Linux 2 was the only distribution in the matrix that used yum.

It is replaced with Amazon Linux 2023. There is no `swift:6.2-amazonlinux2023` image, so 6.2 is excluded on that OS for both the native test and the static-SDK build. AL2023 is otherwise covered natively on `nightly-main`; it is dnf-based, so it uses the existing dnf pre-build branch.

Note that AL2023 is enabled for `nightly-main`. This cell will fail whenever the latest static-SDK snapshot is ahead of the AL2023 nightly container; the install step won't find a matching nightly toolchain. When it happens, it's easy to spot in the log ("Toolchain not found", exit 44).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -12,7 +12,7 @@ jobs:
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
     with:
-      linux_os_versions: '["amazonlinux2", "bookworm", "noble", "jammy", "rhel-ubi9"]'
+      linux_os_versions: '["bookworm", "noble", "jammy", "rhel-ubi9"]'
       linux_swift_versions: '["6.2", "nightly-main"]'
       linux_pre_build_command: |
         if command -v apt-get >/dev/null 2>&1 ; then # bookworm, noble, jammy
@@ -29,11 +29,6 @@ jobs:
 
           # Test dependencies
           dnf install -y procps
-        elif command -v yum >/dev/null 2>&1 ; then # amazonlinux2
-          yum update -y
-
-          # Test dependencies
-          yum install -y procps
         fi
       linux_build_command: "swift test && swift test -c release && swift test --disable-default-traits"
       enable_freebsd_checks: true
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -12,7 +12,7 @@ jobs:
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
     with:
-      linux_os_versions: '["bookworm", "noble", "jammy", "rhel-ubi9"]'
+      linux_os_versions: '["amazonlinux2023", "bookworm", "noble", "jammy", "rhel-ubi9"]'
       linux_swift_versions: '["6.2", "nightly-main"]'
       linux_pre_build_command: |
         if command -v apt-get >/dev/null 2>&1 ; then # bookworm, noble, jammy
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
jakepetroules:
```suggestion
        elif command -v dnf >/dev/null 2>&1 ; then # amazonlinux2023, rhel-ubi9
```

.github/workflows/pull_request.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
     with:
-      linux_os_versions: '["amazonlinux2", "bookworm", "noble", "jammy", "rhel-ubi9"]'
+      linux_os_versions: '["bookworm", "noble", "jammy", "rhel-ubi9"]'
```
jakepetroules: Can you add Amazon Linux 2023 in its place?

```suggestion
      linux_os_versions: '["amazonlinux2023", "bookworm", "noble", "jammy", "rhel-ubi9"]'
```

jakepetroules (changes requested):

.github/workflows/pull_request.yml
broken-circle: Sure, just tried it, but looks like `swift:6.2-amazonlinux2023` doesn't exist. The [`6.2/amazonlinux/`](https://github.com/swiftlang/swift-docker/tree/main/6.2/amazonlinux) directory in swift-docker only has `2/`, no `2023/`, so both 6.2 amazonlinux2023 cells are failing immediately. I don't think there's a way to add this for nightly-main only through `linux_os_versions` (it's crossed with `linux_swift_versions`, so it'd always produce a 6.2 cell too), so it'd need its own job. Maybe a separate PR?

broken-circle: Got it working as a separate job.

jakepetroules: The correct way to do this would be to keep it as part of the same job like you had before, but add:

```
linux_exclude_swift_versions: '[{"swift_version": "6.2", "os_version": "amazonlinux2023"}]'
```

to exclude AL2023 from 6.2

broken-circle: Thanks! That fixes the native `linux-build` cell.

The [`linux-static-sdk-build` job](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/swift_package_test.yml) has its own matrix (`linux_static_sdk_versions` crossed with `linux_os_versions`) and its own `linux_static_sdk_exclude_swift_versions`, so with AL2023 in `linux_os_versions` it still produces two failing static-SDK cells: 6.2 (no matching image) and `nightly-main` (no matching nightly toolchain).

Since the static-SDK build cross-compiles to musl, the host OS adds little coverage there, so I've excluded AL2023 from static-SDK for both versions:

```yaml
linux_exclude_swift_versions: '[{"swift_version": "6.2", "os_version": "amazonlinux2023"}]'
linux_static_sdk_exclude_swift_versions: '[{"swift_version": "6.2", "os_version": "amazonlinux2023"}, {"swift_version": "nightly-main", "os_version": "amazonlinux2023"}]'
```

If you'd rather keep the `nightly-main` static-SDK cell for the host coverage, I can drop the second entry from `linux_static_sdk_exclude_swift_versions`, though it'll flake red whenever the snapshot leads the AL2023 nightly container.

iCharlesHu (approved):

broken-circle:
Also updated the README Cross-Platform Support table by changing "Amazon Linux 2" to "Amazon Linux 2023". Kept it as "Automated" since AL2023 is tested natively on every PR; the static-SDK build is excluded there (no 6.2 image; musl host adds little), but the table doesn't break down per-version or static-SDK coverage for any row anyway.

.github/workflows/pull_request.yml
jakepetroules: I'd leave this one enabled `{"swift_version": "nightly-main", "os_version": "amazonlinux2023"}` - it's enabled by default and there's no particular reason to _exclude_ it.

broken-circle: Sounds good, keeping it enabled. Just mentioning it since it'll surface later: this cell fails whenever the latest static-SDK snapshot is ahead of the AL2023 nightly container; the install step can't find a matching nightly toolchain. It should pass once the container rebuilds to the current snapshot, but it'll fail until then. It's easy to spot in the log though ("Toolchain not found", exit 44).

jakepetroules (approved):

PR #336, opened 2026-07-01
Fix use-after-free when Windows input writing is cancelled

broken-circle:
Follow-up to #325, which fixed a tail-loss bug on the Windows output-capture path where a completed overlapped `ReadFile()` was discarded when child termination cancelled the read.

On Windows, when a subprocess terminates while its standard input is still being written, the write's completion races the cancellation that termination triggers. When the cancellation won, `write(_:to:for:)` observed `nil` from the signal stream and returned the byte count confirmed so far without reconciling a `WriteFile()` that might still be pending. A pending `WriteFile()` is the kernel reading from the caller's borrowed `span` storage and poised to write its result into the stack `overlapped`; both were released once `write(_:to:for:)` returned and the borrow ended, so the kernel could read freed storage or write a freed `overlapped`. Unlike the analogous read-side bug, this had no functional symptom: the bytes were destined for a pipe whose reader (the child process) had already terminated, so no captured output was affected.

On non-Windows platforms the write path writes the file descriptor directly with a synchronous call, retains no reference to the buffer past return, and has no equivalent exposure.

This PR reconciles the overlapped operation before returning on cancellation. A `WriteFile()` that already finished needs nothing, and one still pending (reachable when a byte-mode write is parked on a full pipe buffer at the moment the child stops draining and terminates) is cancelled with `CancelIoEx()` and waited on with `GetOverlappedResult()` before the borrow is released, so the kernel is provably done with both the buffer and `overlapped`. The returned count is deliberately left unchanged; bytes the reconciliation observes as transferred are not folded in, since the pipe's only reader has terminated and nothing will consume them. The read-side counterpart is renamed to `reconcileReadOverlappedAfterCancellation()` for symmetry with the new `reconcileWriteOverlappedAfterCancellation()`.

Note: The same lifetime exposure exists on both read and write when the signal stream throws rather than finishing, where a pending operation's buffer is likewise released without reconciliation. It's symmetric across read and write, so I've left it out of scope here and will open a separate PR to fix it.

<details>
<summary>Testing and documentation</summary>

Unlike #325, this bug produces no observable behavioral difference: correct output is unaffected, and the returned count is intentionally identical with or without the fix. As far as I can tell, no functional test can guard it since there's nothing to assert. The defect is a borrow-contract violation that would manifest as memory corruption only under a checking allocator with heap reuse landing in the freed window, which isn't a clean assertion.

Below is documentation upon which this fix depends:

- [Ownership of the output buffer during a write operation](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile#remarks):

> [The output] buffer must remain valid for the duration of the write operation. The caller must not use this buffer until the write operation is completed. ... Accessing the output buffer while a write operation is using the buffer may lead to corruption of the data written from that buffer. Applications must not write to, reallocate, or free the output buffer that a write operation is using until the write operation completes.

> Because the write operation starts at the offset that is specified in the [`OVERLAPPED`](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped) structure, and `WriteFile` may return before the system-level write operation is complete (write pending), neither the offset nor any other part of the structure should be modified, freed, or reused by the application until the event is signaled (that is, the write completes).

- [Setting the file handle to nonsignaled/signaled](https://learn.microsoft.com/en-us/windows/win32/fileio/synchronous-and-asynchronous-i-o#synchronous-and-asynchronous-io-considerations):

> Each time an I/O operation is started, the operating system sets the file handle to the nonsignaled state. Each time an I/O operation is completed, the operating system sets the file handle to the signaled state. 

- [`GetOverlappedResult()` `NULL` `hEvent` fallback to file handle](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getoverlappedresult#remarks):

> If the hEvent member of the [`OVERLAPPED`](https://learn.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-overlapped) structure is `NULL`, the system uses the state of the `hFile` handle to signal when the operation has been completed. Use of file, named pipe, or communications-device handles for this purpose is discouraged. It is safer to use an event object because of the confusion that can occur when multiple simultaneous overlapped operations are performed on the same file, named pipe, or communications device. In this situation, there is no way to know which operation caused the object's state to be signaled.

- [Byte-mode blocking writes pend on a full buffer](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipe-type-read-and-wait-modes#wait-mode):

> With a blocking-wait handle, the write operation cannot succeed until sufficient space is created in the buffer by a thread reading from the other end of the pipe.

</details>

### DIFF
--- a/Sources/Subprocess/IO/AsyncIO+Windows.swift
+++ b/Sources/Subprocess/IO/AsyncIO+Windows.swift
@@ -298,7 +298,7 @@ final class AsyncIO: @unchecked Sendable {
                 // with its bytes sitting in `resultBuffer`. Those bytes have been
                 // pulled out of the pipe, so the post-cancellation drain cannot
                 // recover them: dropping them here loses exactly one buffer.
-                return self.reconcileOverlappedAfterCancellation(
+                return self.reconcileReadOverlappedAfterCancellation(
                     handle: handle,
                     overlapped: &overlapped
                 )
@@ -339,7 +339,7 @@ final class AsyncIO: @unchecked Sendable {
     /// this call. On the pending path this function blocks until the cancelled
     /// read has fully unwound, guaranteeing the kernel is done with both before
     /// they are freed.
-    private func reconcileOverlappedAfterCancellation(
+    private func reconcileReadOverlappedAfterCancellation(
         handle: HANDLE,
         overlapped: inout _OVERLAPPED
     ) -> OverlappedReadOutcome {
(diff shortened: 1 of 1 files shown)


broken-circle:
The Linux CI failure is due to Amazon Linux 2 end-of-life and will be resolved once #335 lands. The Android failure is unrelated to Subprocess (emulator installation flake).

compnerd (approved):

PR #338, opened 2026-07-02
Narrow `StringOutput.OutputType` to non-optional

broken-circle:
`StringOutput.output(from:)` never returns an optional string, so the associated `OutputType` does not need to be optional. This PR narrows its type to `String`. This is a breaking API change, but one I think is worth landing before 1.0.

The current behavior is that the user always gets a non-optional `String`. Invalid byte sequences are replaced with the Unicode replacement character (U+FFFD), which is the behavior of `String(decoding:as:)` that `output(from:)` already uses via `String(decodingBytes:as:)`. This also makes `StringOutput` consistent with `DataOutput`, whose `OutputType` is non-optional `Data`.

The optional type was introduced in [iCharlesHu/Subprocess/main@`7f9cea8`](https://github.com/iCharlesHu/Subprocess/commit/7f9cea8f5e632f6af33d3fb8006386c582777e35#diff-104673b9ef276affe58f7b7f1b81cf357f8d67a32e1608cc2f969927c170608a) and brought over to this repository in the initial commit. I couldn't find a reason the type was optional. The type permits `nil`, but no code path produces it; if returning `nil` on invalid input is behavior we want, let me know and I'll look into it as a follow-up PR. [`String(validating:as:)`](https://developer.apple.com/documentation/swift/string/init(validating:as:)-84qr9) would be the appropriate API, but it's gated on macOS 15+ (Subprocess supports macOS 13+), so the implementation will require availability-gated fallbacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,8 @@ let result = try await run(
     return lineCount
 }
 
-print(result.closureResult)         // The line count returned from the closure.
-print(result.standardError ?? "")   // The captured standard error.
+print(result.closureResult) // The line count returned from the closure.
+print(result.standardError) // The captured standard error.
 ```
 
 Stream both standard output and standard error, writing to standard input from the same closure:
--- a/Sources/Subprocess/IO/Output.swift
+++ b/Sources/Subprocess/IO/Output.swift
@@ -109,10 +109,10 @@ public struct FileDescriptorOutput: OutputProtocol, ErrorOutputProtocol {
 
 /// An output type that collects the subprocess's output as decoded text using the given encoding.
 public struct StringOutput<Encoding: Unicode.Encoding>: OutputProtocol, ErrorOutputProtocol {
-    public typealias OutputType = String?
+    public typealias OutputType = String
     public let maxSize: Int
 
-    public func output(from span: RawSpan) throws -> String? {
+    public func output(from span: RawSpan) throws -> String {
         span.withUnsafeBytes { ptr in
             let array = Array(ptr)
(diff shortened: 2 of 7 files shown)


broken-circle:
@iCharlesHu, somewhat related question: the 1.0 beta release includes the deprecated `.standardOutput`/`.standardError` aliases on `FileDescriptorOutput` (the `where Self == FileDescriptorOutput` extension in `Output.swift`), each marked `// TODO: remove for 1.0`. Now that the beta has shipped with them present, is removal before 1.0 still the plan, or do they need to stay through 1.0 for source compatibility? I can open a follow-up PR to remove them if you want.

jakepetroules:
Personally I think `String(decoding:as:)` + non-Optional is the right default, if people really want to validate that the text doesn't contain invalid bytes they could always use bytes and do the parsing themselves.

broken-circle:
Exactly what I was thinking!

iCharlesHu:
I originally designed this API to return `String?` because of the possible invalid characters. I forgot `String(decoding:as:)` variant simply adds a replacement character instead of returning `nil`. I agree `String(decoding:as:) + non-Optional` seems like the right choice.

> is removal before 1.0 still the plan,

Yes. We will remove them in the final 1.0 tag. We can also include this change as well.

iCharlesHu:
Proposal update: https://github.com/swiftlang/swift-foundation/pull/2089

iCharlesHu (approved):

PR #339, opened 2026-07-02
deeper curation, revising some wording to less formal structures

heckj:
I ran through and assembled a deeper curation , organizing the symbols within all the various types. There's some notable symbols in here for Windows that - without the symbolgraphs from a Windows build - result in DocC warnings (such as building on Darwin or Linux directly) that I'm intentionally leaving in place, in the hopes that we'll have something to help wrangle the multi-platform capture of symbol graphs here in the near-ish future. It falls back reasonably - missing symbols simply don't appear.

### DIFF
--- a/Sources/Subprocess/Documentation.docc/reference/Arguments.md
+++ b/Sources/Subprocess/Documentation.docc/reference/Arguments.md
@@ -1,0 +1,14 @@
+# ``Arguments``
+
+## Topics
+
+### Creating arguments
+
+- ``init(arrayLiteral:)``
+- ``init(_:)-([String])``
+- ``init(_:)-([[UInt8]])``
+
+### Overriding the executable path
+
+- ``init(executablePathOverride:remainingValues:)-(String?,_)``
+- ``init(executablePathOverride:remainingValues:)-([UInt8]?,_)``
--- a/Sources/Subprocess/Documentation.docc/reference/Environment.Key.md
+++ b/Sources/Subprocess/Documentation.docc/reference/Environment.Key.md
@@ -1,0 +1,12 @@
+# ``Environment/Key``
+
+## Topics
+
+### Creating a key
+
+- ``init(stringLiteral:)``
+- ``init(rawValue:)``
+
+### Reading the key
+
+- ``rawValue``
--- a/Sources/Subprocess/Documentation.docc/reference/Environment.Key.md
+++ b/Sources/Subprocess/Documentation.docc/reference/Environment.Key.md
@@ -5,8 +5,8 @@
 ### Creating a key
 
 - ``init(stringLiteral:)``
+
+### Working with raw values
+
 - ``init(rawValue:)``
-
-### Reading the key
-
 - ``rawValue``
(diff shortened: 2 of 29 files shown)


Sources/Subprocess/Documentation.docc/reference/Signal.md
```diff
@@ -0,0 +1,31 @@
...
+- ``terminalClosed``
+- ``windowSizeChange``
+
+### Sending alarm and user-defined signals
+
+- ``alarm``
+- ``userDefinedOne``
+- ``userDefinedTwo``
+
+### Creating a signal
```
invalidname: Since you've also got the `rawValue` property in here, I don't think we should call this topic "Creating".
```suggestion
### Working with raw values
```
And you're right to put it at the bottom because it's just not very interesting to most developers. I think in the Apple docs, this stuff gets collected into a default `RawRepresentable implementations` topic and you don't have to worry about it.

Sources/Subprocess/Documentation.docc/reference/SubprocessError.md
```diff
@@ -0,0 +1,16 @@
...
+- ``code``
+- ``underlyingError``
+
+### Accessing error codes
+
+- ``Code``
+
+### Inspecting the underlying error on Windows
+
+- ``WindowsError``
```
invalidname, line 16: Interesting. Is `WindowsError` a subtype of `SubprocessError`?

Sources/Subprocess/Documentation.docc/reference/Arguments.md
```diff
@@ -0,0 +1,14 @@
+# ``Arguments``
+
+## Topics
+
+### Creating arguments
+
+- ``init(arrayLiteral:)``
+- ``init(_:)-([String])``
+- ``init(_:)-([[UInt8]])``
```
invalidname, line 9: Nice job with all the disambiguation hashes, by the way. Did you have a tool to help you (like `docc process-catalog emit-generated-curation`) or did you do it by hand?

Sources/Subprocess/Documentation.docc/reference/Environment.Key.md
```diff
@@ -0,0 +1,12 @@
+# ``Environment/Key``
+
+## Topics
+
+### Creating a key
+
+- ``init(stringLiteral:)``
+- ``init(rawValue:)``
```
invalidname: I mentioned this somewhere else, but I would generally collect the `init(rawValue:)` initializer and the `rawValue` property into a topic called `## Working with raw values`, and get it down to the bottom of the curation because nobody finds it interesting. It's just an obvious conformance to `RawRepresentable`.

invalidname (approved):
Looks good, thanks. I have a couple minor comments about burying the `RawRepresentable` conformance spam.

Sources/Subprocess/Documentation.docc/reference/Arguments.md
heckj: used DocC - with the 6.4 (from the Xcode beta) to get the feedback for ambiguous bits, which in 6.4 gives the "nice" versions ;-) A little help from an LLM to do the scut work of validating and applying them.

PR #345, opened 2026-07-04
Narrow `testSubprocessPlatformOptionsSupplementaryGroups()` on Darwin

broken-circle:
`testSubprocessPlatformOptionsSupplementaryGroups()` fails on Darwin:

```shell
% sudo xcrun swift test --scratch-path .build-root --filter testSubprocessPlatformOptionsSupplementaryGroups
...
Test testSubprocessPlatformOptionsSupplementaryGroups() recorded an issue at UnixTests.swift:115:9:
Expectation failed: (actualGroups → []) == (expectedGroups → [1079, 1526, 1744, 1210, 1055, 1878, 1538, 1130, 1836])
```

`id -G` on Darwin doesn't report the groups, but they are applied: the prefork child aborts the spawn if `setgroups(2)` fails, so a successful spawn confirms the call returned `0` and the credential was set; only Darwin's `id -G` read-back omits it. This PR keeps the membership assertion where `id -G` is faithful and asserts only spawn success on Darwin.

### DIFF
--- a/Tests/SubprocessTests/UnixTests.swift
+++ b/Tests/SubprocessTests/UnixTests.swift
@@ -133,6 +133,9 @@ extension SubprocessUnixTests {
             error: .string(limit: .max),
         )
         #expect(idResult.terminationStatus.isSuccess, Comment(rawValue: idResult.standardError))
+        // On Darwin, id -G run as root doesn't report the supplementary groups
+        // installed via setgroups() in the prefork child (`isSuccess` above).
+        #if !canImport(Darwin)
         let ids =
             try idResult
             .standardOutput.split(separator: " ")
@@ -141,6 +144,7 @@ extension SubprocessUnixTests {
         // supplementary groups, so filter to just the expected range
         let actualGroups = Set(ids.filter { expectedGroups.contains($0) })
         #expect(actualGroups == expectedGroups, Comment(rawValue: idResult.standardError))
+        #endif
     }
 
     @Test(


broken-circle:
The Unix `PlatformOptions` tests that require root run on Linux and FreeBSD, but are skipped on macOS. I verified this fix locally, but this change is currently a no-op on CI, since macOS skips the test and the other platforms take the unchanged assertion.

These tests were originally introduced in [iCharlesHu/Subprocess/main@`0600d01`](https://github.com/iCharlesHu/Subprocess/commit/0600d010694e9b1e646fe5d092af8890d62bcded#diff-0a0fdcd4204c34fedb0448ce47be61549509a3500236fe3830c9a22ea73f76fc) with the same skip-without-`sudo` condition. I think these were meant to run in CI, and I can see a few ways to make that happen:

1. A bespoke macOS `sudo` job in this repository's `pull_request.yml`, filtered to a new root suite (`SubprocessUnixRootTests`).
2. A macOS "run as root" option added to the shared `swift_package_test.yml` in `swiftlang/github-workflows`, keeping the runner-specific details centralized there rather than here.
3. Leaving them skipped on macOS (current behavior).

I was leaning toward option 1. I'd open a draft PR initially to validate that the macOS runners allow passwordless `sudo`. There's also runner cleanup to handle (root-owned build artifacts on a persistent runner).

Which approach would you most prefer? I don't think we should leave them skipped. I have the option 1 draft PR mostly ready, or I can get started with option 2, if it's likely such an addition would be accepted there.

jakepetroules:
1 seems like a good idea if it's possible

broken-circle:
I tried in #351, but passwordless `sudo` isn't available on the self-hosted macOS runners. I verified on my fork that it's available if we use `macos-26`. Are you okay with one hosted-macOS exception to gain macOS coverage on these tests?

jakepetroules:
Hmm, I was under the impression that _only_ self-hosted macOS runners were available for the swiftlang org, but @shahmishal could confirm the details there for sure.

broken-circle:
@shahmishal Could you please let us know the policy on non-self-hosted macOS runners in swiftlang CI?

Context: Subprocess has a handful of Unix tests that require root access to run. These tests have never run on macOS CI, leaving the door open for potential regressions. We cannot currently use passwordless `sudo` on the self-hosted runners, but we can use it on `macos-latest`, and would run only those few tests that require root on the non-self-hosted runner. See #351.

Issue #151, opened 2025-08-17
[Feature] How do you pipe?


DevilDimon:
I'm trying the library out and essentially my question is how to execute the equivalent of the following:

```
grep 'text' . | wc -l
```

using this library without manually collecting the output of each step.

The `run` function signature only appears to accept one subprogram name as an argument to launch a single process (which is understandable).

The current outputs/inputs are types like String or Data, so it's not immediately clear how to 'pipe' the output of one subprocess to another, without resorting to doing it manually via an intermediate String or something.
I suppose all the shells on all supported platforms support piping so there could be a way to simplify this common task? (or maybe there already is?)

cmcgee1024:
You can check out the implementation in #147 to see how you can set up a process pipeline manually. Basically, there is FileDescriptor.pipe() to set up read/write id's, and those can be provided to the input/output of `run()` using the `.fileDescriptor()` construction. Give each process its own Task, and you can block on the last one to get the final output in whichever form works best for you.

heckj:
Note: poke @heckj to help update the streaming article we added with improved API when it's available (after the 1.0 release)

PR #347, opened 2026-07-08
draft articles, providing details about using subprocess run and it's more common inputs and patterns

heckj:
I wanted to wrap up the great conversations I saw during the development work for this library to provide some of the concrete "how to use it" advice touching on collecting result content, searching for executables vs. providing a path, and the just edging on the stream processing for results from a long running process - or even the input streaming.

I put together these two articles, pulling heavily in influence from the conversations in Swift Forums.

Please look over the example code blocks with a high degree of skepticism, and verify that I'm not suggesting poor code patterns or encouraging use in a way that would be harmful.

### DIFF
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -148,7 +148,7 @@ Keeping them apart is the difference between handling a real problem and misread
 Throughout the Subprocess documentation:
 
 - A **Swift error** is a value thrown out of `run` that you handle with `do`/`catch`. Subprocess reserves this for problems it can't express as a result: a broken setup or environment, or a body closure of your own that throws.
-- A **command failure** is a subprocess that ran to completion but exited with a non-zero code. This is reported by the ``terminationStatus`` of the ``ExecutionResult``, not a Swift error.
+- A **command failure** is a subprocess that ran to completion but exited with a non-zero code. This is reported by the ``ExecutionResult/terminationStatus`` of the ``ExecutionResult``, not a Swift error.
 - **Standard error output** is bytes the command wrote to its standard error stream. Many programs write progress or diagnostics there while succeeding, so it's ordinary output you collect through the `error:` parameter — not a signal that anything went wrong.
 
 `run` doesn't throw a Swift error when a command exits with a non-zero code.
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -40,6 +40,8 @@ The preceding example uses ``run(_:arguments:environment:workingDirectory:platfo
 
 Take note of the following behaviors in this example:
 
+Take note of the following behaviors in this example:
+
 - The `output` parameter is required.
   The example uses `.string(limit:)` to collect standard output as a string (`String`) no longer than 4096 bytes.
 - The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -141,8 +141,6 @@ let result = try await run(
 print(result.standardOutput)   // the commit hash
 ```
 
-Because `.string(limit:)` produces a `String` rather than an optional, you use `result.standardOutput` directly, with no unwrapping.
-
 ### Detect failure
 
 Around a subprocess, the word “error” pulls in three directions.
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -9,9 +9,11 @@ Running a subprocess takes the input you provide, and you control how its output
 Depending on the command, a subprocess might produce output, error output, both, or neither.
 Subprocess lets you handle each of these cases independently.
 
-You can either collect or stream the output.
-Collecting waits for the subprocess to finish and hands you the result.
-Streaming lets you read output while the subprocess is still running.
+The function that launches a subprocess, `run`, comes in two forms.
+The collecting form waits for the subprocess to finish and provides the result.
+The streaming form lets you read output while the subprocess runs.
+You choose between them by whether you pass a trailing closure.
+This article covers the collecting form; for streaming, see <doc:StreamingAndInput>.
 
 Subprocess encloses the lifetime of the process you run within the `run` call.
 When `run` returns, the process is complete and cleaned up.
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -5,8 +5,9 @@ and process lifetime fit together.
 
 ## Overview
 
-Running a subprocess takes the input you provide and produces both output and error.
-The API of Subprocess lets you control each one separately.
+Running a subprocess takes the input you provide, and you control how its output and error are collected or streamed.
+Depending on the command, a subprocess might produce output, error output, both, or neither.
+Subprocess lets you handle each of these cases independently.
 
 You can either collect or stream the output.
 Collecting waits for the subprocess to finish and hands you the result.
@@ -28,17 +29,15 @@ let result = try await run(
     arguments: ["-la"],
     output: .string(limit: 4096)
 )
-if let stdout = result.standardOutput {
-    print(stdout)
-}
+print(result.standardOutput)
 ```
 
 The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
 
 - The `output` parameter is required.
-  The example uses `.string(limit:)` to collect standard output as an optional string (`String?`) no longer than 4096 bytes.
+  The example uses `.string(limit:)` to collect standard output as a string (`String`) no longer than 4096 bytes.
 - The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
-  If you don't specify a limit, the default limit is 128 KB.
+  There's no default limit — you always specify one, because a reasonable ceiling depends entirely on the command you run, anywhere from a few bytes to many megabytes.
   Treat the limit as a ceiling, not a truncation: output that fits today can grow past it tomorrow, and when it does, `run` throws ``SubprocessError`` rather than silently handing you a partial result you might mistake for the whole.
 - The `input` parameter defaults to `.none` and the `error` parameter defaults to `.discarded`.
 
@@ -58,7 +57,7 @@ Each element is one argument, passed to the command exactly as written:
 let result = try await run(
     .name("git"),
     arguments: ["commit", "-m", "a message with spaces"],
-    output: .string(limit: 4096)
+    output: .string(limit: 16 * 1024)
 )
 ```
 
@@ -81,7 +80,7 @@ let result = try await run(
 )
 ```
 
-To keep the two streams together instead — the equivalent of using `2>&1` in a shell — pass ``ErrorOutputProtocol/combinedWithOutput`` as the error.
+To keep the output and error together instead — the equivalent of using `2>&1` in a shell — pass ``ErrorOutputProtocol/combinedWithOutput`` as the error.
 This merges standard error into standard output.
 Only the `error:` line changes:
 
@@ -106,7 +105,7 @@ For more detail on how to use streaming output, see <doc:StreamingAndInput>.
 A collecting `run` returns an ``ExecutionResult``, which includes:
 
 - `standardOutput` and `standardError`, typed by the choices you made — a
-  `String?` for `.string(limit:)`, a `[UInt8]` for `.bytes(limit:)`, and so on.
+  `String` for `.string(limit:)`, a `[UInt8]` for `.bytes(limit:)`, and so on.
 - ``TerminationStatus`` — how the process ended.
 - ``ProcessIdentifier`` — the process's identifier.
 
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -39,8 +39,7 @@ The preceding example uses ``run(_:arguments:environment:workingDirectory:platfo
   The example uses `.string(limit:)` to collect standard output as an optional string (`String?`) no longer than 4096 bytes.
 - The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
   If you don't specify a limit, the default limit is 128 KB.
-  If the subprocess produces more than the limit, `run` throws ``SubprocessError`` rather than truncating silently,
-  so you don't mistake a partial result for a complete one.
+  Treat the limit as a ceiling, not a truncation: output that fits today can grow past it tomorrow, and when it does, `run` throws ``SubprocessError`` rather than silently handing you a partial result you might mistake for the whole.
 - The `input` parameter defaults to `.none` and the `error` parameter defaults to `.discarded`.
 
 ### Name the command and its arguments
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -12,7 +12,7 @@ Subprocess lets you handle each of these cases independently.
 The function that launches a subprocess, `run`, comes in two forms.
 The collecting form waits for the subprocess to finish and provides the result.
 The streaming form lets you read output while the subprocess runs.
-You choose between them by whether you pass a trailing closure.
+You choose between them depending on whether you pass a trailing closure.
 This article covers the collecting form; for streaming, see <doc:StreamingAndInput>.
 
 Subprocess encloses the lifetime of the process you run within the `run` call.
@@ -104,16 +104,51 @@ For more detail on how to use streaming output, see <doc:StreamingAndInput>.
 
 ### Read what you get back
 
-A collecting `run` returns an ``ExecutionResult``, which includes:
+A collecting `run` returns an ``ExecutionResult`` that carries everything the finished process produced:
 
-- `standardOutput` and `standardError`, typed by the choices you made — a
-  `String` for `.string(limit:)`, a `[UInt8]` for `.bytes(limit:)`, and so on.
-- ``TerminationStatus`` — how the process ended.
-- ``ProcessIdentifier`` — the process's identifier.
+| Property | Type | What it holds |
+| --- | --- | --- |
+| `standardOutput` | matches your `output:` choice | the collected standard output |
+| `standardError` | matches your `error:` choice | the collected standard error |
+| `terminationStatus` | ``TerminationStatus`` | how the process ended |
+| `processIdentifier` | ``ProcessIdentifier`` | the identifier of the process that ran |
 
-A command that ran and failed is a normal result, not an error.
-`run` doesn't throw when a command exits with a non-zero code.
-Use the termination status to understand how the process completed:
+The types of `standardOutput` and `standardError` follow the collection method you chose for each stream:
+
+| Collection method | Result type |
+| --- | --- |
+| `.string(limit:)` | `String` |
+| `.bytes(limit:)` | `[UInt8]` |
+| ``OutputProtocol/discarded`` | `Void` (nothing collected) |
+
+Both `.string(limit:)` and `.bytes(limit:)` take a required `limit` — a byte count with no default, applied before decoding.
+Use `.bytes(limit:)` when the output isn't text, such as an image or an archive.
+
+In the common case, a command runs, succeeds, and hands you its output, which you read from the result:
+
+```swift
+let result = try await run(
+    .name("git"),
+    arguments: ["rev-parse", "HEAD"],
+    output: .string(limit: 4096)
+)
+print(result.standardOutput)   // the commit hash
+```
+
+Because `.string(limit:)` produces a `String` rather than an optional, you use `result.standardOutput` directly, with no unwrapping.
+
+### Detect failure
+
+Around a subprocess, the word “error” pulls in three directions.
+Keeping them apart is the difference between handling a real problem and misreading a healthy result as a broken one.
+Throughout the Subprocess documentation:
+
+- A **Swift error** is a value thrown out of `run` that you handle with `do`/`catch`. Subprocess reserves this for problems it can't express as a result: a broken setup or environment, or a body closure of your own that throws.
+- A **command failure** is a subprocess that ran to completion but exited with a non-zero code. This is a normal ``ExecutionResult``, not a Swift error.
+- **Standard error output** is bytes the command wrote to its standard error stream. Many programs write progress or diagnostics there while succeeding, so it's ordinary output you collect through the `error:` parameter — not a signal that anything went wrong.
+
+`run` doesn't throw a Swift error when a command exits with a non-zero code.
+A command that ran and failed is a normal result; inspect its ``TerminationStatus`` to see how it ended:
 
 ```swift
 switch result.terminationStatus {
@@ -126,13 +161,35 @@ case .signaled(let signal):
 }
 ```
 
-For a simple pass-or-fail check, ``TerminationStatus/isSuccess`` collapses that
-to a Boolean. The `signaled` case exists only on platforms that report
-signals; on Windows, every command ends as `exited`.
+For a simple pass-or-fail check, ``TerminationStatus/isSuccess`` collapses that to a Boolean.
+The `signaled` case exists only on platforms that report signals; on Windows, every command ends as `exited`.
 
-`run` does throw when provided a command that can't start,
-for example, when ``Executable/name(_:)`` finds nothing in `PATH`.
-As mentioned earlier, `run` also throws ``SubprocessError`` if the output exceeds a limit.
+`run` throws a Swift error in only two situations:
+
+- It throws a ``SubprocessError`` for a problem with the setup or environment — for example, when ``Executable/name(_:)`` finds nothing in `PATH`, or when the collected output exceeds the `limit` you specified.
+- It rethrows any error your own body closure throws, in the streaming form described in <doc:StreamingAndInput>.
+
+Because a command failure isn't a Swift error, you handle the two concerns in different places — `do`/`catch` for a process that couldn't run, and ``TerminationStatus/isSuccess`` for one that ran and failed:
+
+```swift
+do {
+    let result = try await run(
+        .name("git"),
+        arguments: ["status"],
+        output: .string(limit: 64 * 1024)
+    )
+    if result.terminationStatus.isSuccess {
+        print(result.standardOutput)
+    } else {
+        print("git reported failure")
+    }
+} catch {
+    // A Swift error: git wasn't found, or the output exceeded the limit.
+    print("couldn't run git: \(error)")
+}
+```
+
+> Note: A non-zero exit never surfaces as a Swift error, and neither does output on standard error. To run a `catch` block when a command “fails,” check ``TerminationStatus/isSuccess`` yourself and throw from your own code.
 
 ### Understand the process lifetime
 
--- a/Sources/Subprocess/Documentation.docc/GettingStarted.md
+++ b/Sources/Subprocess/Documentation.docc/GettingStarted.md
@@ -9,9 +9,10 @@ Running a subprocess takes the input you provide, and you control how its output
 Depending on the command, a subprocess might produce output, error output, both, or neither.
 Subprocess lets you handle each of these cases independently.
 
-The function that launches a subprocess, `run`, comes in two forms.
-The collecting form waits for the subprocess to finish and provides the result.
-The streaming form lets you read output while the subprocess runs.
+The function that launches a subprocess, `run`, comes in two forms:
+* The *collecting* form waits for the subprocess to finish and provides the result.
+* The *streaming* form lets you read output while the subprocess runs.
+
 You choose between them depending on whether you pass a trailing closure.
 This article covers the collecting form; for streaming, see <doc:StreamingAndInput>.
 
@@ -20,8 +21,9 @@ When `run` returns, the process is complete and cleaned up.
 
 ### Run a command and read its output
 
-Import the package, identify what to run and possibly its arguments,
-specify how to collect its output, and await the result:
+To run a subprocess, import the `Subprocess` framework and call a flavor of the `run` function, identifying what to run, any arguments to use, and how to handle its output.
+You `await` this call, which means that the subprocess execution is finished when the code resumes.
+The following snippet shows an example of executing the `ls -la` command as a subprocess:
 
 ```swift
 import Subprocess
@@ -36,6 +38,8 @@ print(result.standardOutput)
 
 The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
 
+Take note of the following behaviors in this example:
+
 - The `output` parameter is required.
   The example uses `.string(limit:)` to collect standard output as a string (`String`) no longer than 4096 bytes.
 - The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
@@ -45,7 +49,7 @@ The preceding example uses ``run(_:arguments:environment:workingDirectory:platfo
 
 ### Name the command and its arguments
 
-The first parameter is an ``Executable`` that you can specify in one of two ways:
+The first parameter to `run` is an ``Executable`` that you can specify in one of two ways:
 
 - ``Executable/name(_:)`` — `.name("ls")` looks the command up using the `PATH`
   environment variable, the way a shell finds it.
@@ -108,10 +112,10 @@ A collecting `run` returns an ``ExecutionResult`` that carries everything the fi
 
 | Property | Type | What it holds |
 | --- | --- | --- |
-| `standardOutput` | matches your `output:` choice | the collected standard output |
-| `standardError` | matches your `error:` choice | the collected standard error |
-| `terminationStatus` | ``TerminationStatus`` | how the process ended |
-| `processIdentifier` | ``ProcessIdentifier`` | the identifier of the process that ran |
+| `standardOutput` | Matches your `output:` choice | The collected standard output |
+| `standardError` | Matches your `error:` choice | The collected standard error |
+| `terminationStatus` | ``TerminationStatus`` | How the process ended |
+| `processIdentifier` | ``ProcessIdentifier`` | The identifier of the process that ran |
 
 The types of `standardOutput` and `standardError` follow the collection method you chose for each stream:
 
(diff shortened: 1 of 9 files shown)


Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
+# Getting started with Subprocess
+
+Run a command, collect its output, and see how input, output,
+and process lifetime fit together.
+
+## Overview
+
+Running a subprocess takes the input you provide and produces both output and error.
```
dempseyatgithub: I think the error part of ‘produces both output and error’ reads awkwardly.

Especially for a reader new to subprocesses or unfamiliar with UNIX stdin stdout, stderr, the phrase 'produces error' could be confusing.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+    .name("ls"),
+    arguments: ["-la"],
+    output: .string(limit: 4096)
+)
+if let stdout = result.standardOutput {
+    print(stdout)
+}
+```
+
+The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
```
dempseyatgithub, line 39: I think this is supposed to end after `error:)`

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+
+- The `output` parameter is required.
+  The example uses `.string(limit:)` to collect standard output as an optional string (`String?`) no longer than 4096 bytes.
+- The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
+  If you don't specify a limit, the default limit is 128 KB.
```
dempseyatgithub: Just confirming that the default limit is equivalent to passing in 131072 (128 * 1024) as the limit.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+  location and skips the search.
+
+Arguments are a separate ``Arguments`` value that you can write as an array literal.
+Each element is one argument, passed to the command exactly as written:
+
+```swift
+let result = try await run(
+    .name("git"),
+    arguments: ["commit", "-m", "a message with spaces"],
+    output: .string(limit: 4096)
```
dempseyatgithub: Is there a reason both initial examples use a limit of 4096?  Using it in the first example motivates discussion of the limit argument, the default limit, byte count vs character count.

Is it not recommended to just use the default value?  This might be a good place to show sample code that does use the default.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+```swift
+let result = try await run(
+    .name("swift"),
+    arguments: ["build"],
+    output: .string(limit: 2 * 1024 * 1024),
+    error: .string(limit: 512 * 1024)
+)
+```
+
+To keep the two streams together instead — the equivalent of using `2>&1` in a shell — pass ``ErrorOutputProtocol/combinedWithOutput`` as the error.
```
dempseyatgithub: Use of 'two streams' may be confusing since this is using the 'collect' version of `run`.

Maybe just explicitly say 'to keep the output and error together instead...'

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+### Read what you get back
+
+A collecting `run` returns an ``ExecutionResult``, which includes:
+
+- `standardOutput` and `standardError`, typed by the choices you made — a
+  `String?` for `.string(limit:)`, a `[UInt8]` for `.bytes(limit:)`, and so on.
+- ``TerminationStatus`` — how the process ended.
+- ``ProcessIdentifier`` — the process's identifier.
+
+A command that ran and failed is a normal result, not an error.
```
dempseyatgithub: I think the difference between a Swift error being thrown, a command failing, and a command writing to standard error could be confusing, especially with the overloading of the word 'error'.

It may make sense to add a brief paragraph that states the differences and possibly states some naming choices to be used throughout the Subprocess documentation.  (e.g. Always use 'Swift error' to refer to a thrown error from the API.)

dempseyatgithub: This section might make a good subheading of its own, something like 'Detecting failure' or 'Handling command failure'.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+The examples in this article wait for the process to complete and collect the output.
+This approach fits a short-lived subprocess whose output fits in memory, when you want the result after it exits.
+
+The alternative, when a subprocess is long-running or its output is large or open-ended, and you want to act on output before it exits, is to use ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:body:)``.
+This `run` takes a trailing closure in which you iterate over the output as it arrives.
+For more detail on how to use streaming output, see <doc:StreamingAndInput>.
+
+### Read what you get back
```
dempseyatgithub, line 109: About 3/4 of this section is about failed commands. Maybe add a code example of reading `standardOutput`.

Note also that this is the first mention of `.bytes(limit:)`.

Possibly instead of putting it in the bullet point add a paragraph that says something like. 'You will receive output of the type specified in the run command.  Earlier, you saw `.string(limit:)` specified. Here are some other options.' - followed by a table like:

`.string(limit:)`   `String?`
`.bytes(limit:)`    `[UInt8]`

This would make the 'Read what you get back' section more focused on the success case and allow the discussion of command failures to be its own section.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,253 @@
+# Streaming output and providing input
+
+Stream a subprocess's output as it arrives and feed it input, while following
```
dempseyatgithub: Possibly 'stream the output of a subprocess as it arrives...'

(Personal preference - I'm not a fan of possessives.)

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,253 @@
...
+the one rule that keeps pipes from deadlocking.
+
+## Overview
+
+<doc:GettingStarted> shows how `run` waits for the subprocess to exit, then hands you all of its output at once.
```
dempseyatgithub: Since this article is all about using `run` to stream output, I think it's confusing to say `run` waits for the subprocess to exit.

I think in both articles, it would be useful to have a very clear mention that you can use `run` in two different ways, summarize the ways, and that how `run` will operate depends on the arguments you provide.  (The Getting Started article includes this, but I think it would be useful in both places.)

Then this sentence would follow and say something like "<doc:GettingStarted> shows how to use `run` to wait for the subprocess to exit and collect the output all at once."  Using the more precise 'collect'.

Sources/Subprocess/Documentation.docc/GettingStarted.md
iCharlesHu: I agree with @dempseyatgithub. Additionally, depending on the child process, it might not produce any output or error. IMO this paragraph could be rephrased something like

```
Running a subprocess with the input you provide and controlling how the output and error are collected or streamed.
```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+
+```swift
+import Subprocess
+
+let result = try await run(
+    .name("ls"),
+    arguments: ["-la"],
+    output: .string(limit: 4096)
+)
+if let stdout = result.standardOutput {
```
iCharlesHu: Sorry this is a brand new change: https://github.com/swiftlang/swift-subprocess/pull/338

`StringOutput.OutputType` is no longer optional, so here you can just `print(result.standardOutput)`.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+    print(stdout)
+}
+```
+
+The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
+
+- The `output` parameter is required.
+  The example uses `.string(limit:)` to collect standard output as an optional string (`String?`) no longer than 4096 bytes.
```
iCharlesHu: This section needs to be updated:  #338

broken-circle:
Heads-up: #349 removes the deprecated `standardOutput`/`standardError` aliases on `FileDescriptorOutput`, including their DocC curation. This PR touches those same two lines in `OutputProtocol.md`, so depending on merge order there may be a collision. Mentioning it in case it's easiest to just drop those two entries here; I haven't looked at the rest of this PR's changes yet, so those properties may also be referenced elsewhere.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,253 @@
...
+When you need to step beyond that, you can process a stream of results instead.
```
dempseyatgithub: I think 'step beyond' gives the connotation that the collecting version of run is the 'baby' version, but it's a very valid use case that shouldn't be discounted.

Was thinking maybe this section could be something like this and include a list of use cases as bullet list:

"Sometimes, you may need to process a stream of subprocess results instead.

For example:
- Reading each line of a long-running subprocess as it is available
- Handling output that is too large to hold in memory
- Providing input to the subprocess as it runs"

Sources/Subprocess/Documentation.docc/GettingStarted.md
iCharlesHu: > Is it not recommended to just use the default value?

We purposefully did not set a default value for this limit. Depending on the child process being run, the expected output could be from bytes to megabytes. There is not a single default value that would suffice for most use cases. Therefore, `output:` does not have a default value to "force" the user to pick one.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,143 @@
...
+The alternative, when a subprocess is long-running or its output is large or open-ended, and you want to act on output before it exits, is to use ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:body:)``.
+This `run` takes a trailing closure in which you iterate over the output as it arrives.
+For more detail on how to use streaming output, see <doc:StreamingAndInput>.
+
+### Read what you get back
+
+A collecting `run` returns an ``ExecutionResult``, which includes:
+
+- `standardOutput` and `standardError`, typed by the choices you made — a
+  `String?` for `.string(limit:)`, a `[UInt8]` for `.bytes(limit:)`, and so on.
```
iCharlesHu: Ditto for https://github.com/swiftlang/swift-subprocess/pull/338 updates

Sources/Subprocess/Documentation.docc/GettingStarted.md
iCharlesHu: > It may make sense to add a brief paragraph that states the differences and possibly states some naming choices to be used throughout the Subprocess documentation.

I agree. Having a dedicated section discussing error handling would be very useful. Just to clarify:
- Subprocess only throws `Swift.Error` when it fails to spawn the child process. Child process exiting with non-zero code is not considered a failure.
- Subprocess only throws two kinds of errors:
  - `SubprocessError` which indicates something's wrong with the environment or setup
  - Any custom error thrown from the streaming body closure.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Understand the closure scope
+
+The trailing closure on `run` is how the library guarantees that the subprocess lives only
+as long as the `run` call.
+The closure runs concurrently with the live subprocess: while your code reads output inside the body, the process is still running.
+The function `run` returns only after the body returns and the process exits.
+Because of that, the `execution` value, its streams, and the input writer are valid only inside the body.
+
+## Drain every stream, concurrently
```
iCharlesHu: Thank you!! This section is very useful. I was struggling to find the right place to emphasize the concurrency requirement.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,253 @@
...
+The following example provides a static string as input:
+
+```swift
+let result = try await run(
+    .name("wc"),
+    arguments: ["-l"],
+    input: .string("one\ntwo\n"),
+    output: .string(limit: 4096)
+)
+print(result.standardOutput ?? "")   // the line count
```
iCharlesHu: Ditto: no longer optional.

Sources/Subprocess/Documentation.docc/GettingStarted.md
dempseyatgithub: This is misleading. You must specify a limit, as @iCharlesHu states in a subsequent comment. There is no 'non-limit' .string output option.

Perhaps a small amount of guidance here as to the size of the limit would be useful?

Sources/Subprocess/Documentation.docc/GettingStarted.md
dempseyatgithub: And just to clarify, one of the things that could cause a `Subprocess Error` is a limit that is too low for the output/error returned by the subprocess.

Sources/Subprocess/Documentation.docc/GettingStarted.md
heckj: Nope, need the disambiguating extension on that one. The only one that doesn’t need it (or the run) is the one with `body:` closure

heckj:
thanks @broken-circle - no worries, it's already merged - so I'll rebase and I need to update based on it (and other's) changes anyway based on the feedback. Bits I missed when assembling ;-)

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
heckj: Gave this a pass and update, take a look and see how it flows now @dempseyatgithub

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -18,49 +18,35 @@ let result = try await run(.name("ls"), output: .string(limit: 4096))
...
+  default, adds `Data`-based input and output. It imports Foundation — the 
+  system Foundation on Darwin, and swift-foundation's `FoundationEssentials` 
+  elsewhere. Disable the trait to build without that dependency.
 
 ## Topics
 
 ### Running a subprocess
 
+- <doc:GettingStarted>
+- <doc:StreamingAndInput>
```
invalidname: I think it would be cleaner to put the articles in their own section, rather than make them children of the "Running" topic. For the Apple frameworks, we often use an "Essentials" topic for things you need to read first, and that's a perfect place for a "getting started" kind of article.

```suggestion
### Essentials

- <doc:GettingStarted>
- <doc:StreamingAndInput>

### Running a subprocess
```

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -18,49 +18,35 @@ let result = try await run(.name("ls"), output: .string(limit: 4096))
...
-Platform-specific settings live on ``PlatformOptions``: user and group 
-identifiers and session behavior on Unix, quality of service on Darwin, and 
-console and window behavior on Windows. The `SubprocessFoundation` trait is 
-enabled by default and adds the `Data`-based input and output types, which 
-import Foundation (the system Foundation on Darwin and swift-foundation's 
-`FoundationEssentials` elsewhere). Disable the trait to build without that 
-dependency.
+and working directory inline, or build a reusable ``Configuration`` to run repeatedly.
+
+To get more detail on how to `run` to invoke a command, wait for it to finish, then provide everything it produced, read <doc:GettingStarted>.
```
invalidname:
```suggestion
To get more detail on how to use `run` to invoke a command, wait for it to finish, then provide everything it produced, read <doc:GettingStarted>.
```

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -18,49 +18,35 @@ let result = try await run(.name("ls"), output: .string(limit: 4096))
...
+If you're invoking a long running command, a command produces more output than fits into memory, or you want to process what that command produces while it's still running, read <doc:StreamingAndInput>.
```
invalidname:
```suggestion
If you're invoking a long-running command, a command that produces more output than fits into memory, or you want to process what that command produces while it's still running, read <doc:StreamingAndInput>.
```

Sources/Subprocess/Documentation.docc/Subprocess.md
```diff
@@ -18,49 +18,35 @@ let result = try await run(.name("ls"), output: .string(limit: 4096))
...
+
+Subprocess also handles the concerns that surround a running process:
```
invalidname, line 30: Consider reformatting this as a [term list](https://www.swift.org/documentation/docc/formatting-your-documentation-content#Add-Bulleted-Numbered-and-Term-Lists) and see what you think of how it renders. Bolding a word or short phrase and following it with one or more sentences is a smell that you're writing a term list, whether you realize it or not.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Overview
+
+Running a subprocess takes the input you provide, and you control how its output and error are collected or streamed.
+Depending on the command, a subprocess might produce output, error output, both, or neither.
+Subprocess lets you handle each of these cases independently.
+
+The function that launches a subprocess, `run`, comes in two forms.
+The collecting form waits for the subprocess to finish and provides the result.
+The streaming form lets you read output while the subprocess runs.
```
invalidname: A list would give you some visual variety to break up the paragraphs. And what you're doing here is basically setting up a two-item list anyways.
```suggestion
The function that launches a subprocess, `run`, comes in two forms:
* The *collecting* form waits for the subprocess to finish and provides the result.
* The *streaming* form lets you read output while the subprocess runs.

```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+You choose between them depending on whether you pass a trailing closure.
+This article covers the collecting form; for streaming, see <doc:StreamingAndInput>.
+
+Subprocess encloses the lifetime of the process you run within the `run` call.
+When `run` returns, the process is complete and cleaned up.
+
+### Run a command and read its output
+
+Import the package, identify what to run and possibly its arguments,
+specify how to collect its output, and await the result:
```
invalidname: The second-person idiom here is confusing because you're doing two things at once:
* Explaining what the developer in any typical invocation of `run`.
* Setting up the code snippet.

Maybe slow down and do those separately.
```suggestion
To run a subprocess, import the `Subprocess` framework and call a flavor of the `run` function, identifying what to run, any arguments to use, and how to handle its output.
You `await` this call, which means that the subprocess execution is finished when the code resumes.
The following snippet shows an example of executing the `ls -la` command as a subprocess:
```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+    .name("ls"),
+    arguments: ["-la"],
+    output: .string(limit: 4096)
+)
+print(result.standardOutput)
+```
+
+The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
+
+- The `output` parameter is required.
```
invalidname, line 43: I think you could use a sentence to set up this list.
```suggestion
Take note of the following behaviors in this example:

- The `output` parameter is required.
```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+  The example uses `.string(limit:)` to collect standard output as a string (`String`) no longer than 4096 bytes.
+- The `limit` parameter is a byte count, not a character count, and Subprocess applies it before decoding.
+  There's no default limit — you always specify one, because a reasonable ceiling depends entirely on the command you run, anywhere from a few bytes to many megabytes.
+  Treat the limit as a ceiling, not a truncation: output that fits today can grow past it tomorrow, and when it does, `run` throws ``SubprocessError`` rather than silently handing you a partial result you might mistake for the whole.
+- The `input` parameter defaults to `.none` and the `error` parameter defaults to `.discarded`.
+
+### Name the command and its arguments
+
+The first parameter is an ``Executable`` that you can specify in one of two ways:
```
invalidname:
```suggestion
The first parameter to `run` is an ``Executable`` that you can specify in one of two ways:
```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+### Read what you get back
+
+A collecting `run` returns an ``ExecutionResult`` that carries everything the finished process produced:
+
+| Property | Type | What it holds |
+| --- | --- | --- |
+| `standardOutput` | matches your `output:` choice | the collected standard output |
+| `standardError` | matches your `error:` choice | the collected standard error |
+| `terminationStatus` | ``TerminationStatus`` | how the process ended |
+| `processIdentifier` | ``ProcessIdentifier`` | the identifier of the process that ran |
```
invalidname: Just capitalizing the cells.
```suggestion
| Property | Type | What it holds |
| --- | --- | --- |
| `standardOutput` | Matches your `output:` choice | The collected standard output |
| `standardError` | Matches your `error:` choice | The collected standard error |
| `terminationStatus` | ``TerminationStatus`` | How the process ended |
| `processIdentifier` | ``ProcessIdentifier`` | The identifier of the process that ran |
```

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,202 @@
...
+    }
+} catch {
+    // A Swift error: git wasn't found, or the output exceeded the limit.
+    print("couldn't run git: \(error)")
+}
+```
+
+> Note: A non-zero exit never surfaces as a Swift error, and neither does output on standard error. 
+> To run a `catch` block when a command “fails,” check ``TerminationStatus/isSuccess`` yourself and throw from your own code.
+
```
invalidname, line 196: This was a nice section that clears up some potential misunderstandings.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+You choose between them depending on whether you pass that closure.
+This article covers the streaming form; for collecting, see <doc:GettingStarted>.
+
+Reach for streaming when, for example:
+
+- Reading each line of a long-running subprocess as it becomes available.
+- Handling output that's too large to hold in memory.
+- Providing input to the subprocess while it runs.
+
+The streaming API gives you live streams, which come with one rule:
```
invalidname: This looks like you're trying to set up the aside box, but I don't think it's going to render the way you want. It would make sense if the box didn't say "Warning:". 

Maybe consider following the colon on this line with a concise statement of the rule, then make the point more completely and strongly in the box.

```suggestion
The streaming API gives you live streams, which come with one rule: **don't block pipes.**
```

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+The trailing closure on `run` is how the library guarantees that the subprocess lives only
+as long as the `run` call.
+The closure runs concurrently with the live subprocess: while your code reads output inside the body, the process is still running.
+The function `run` returns only after the body returns and the process exits.
+Because of that, the `execution` value, its streams, and the input writer are valid only inside the body.
+
+## Drain every stream, concurrently
+
+Standard output and standard error are separate pipes.
+A process writes to both, and each pipe holds only so much before a write to it blocks — on Linux, for example, about 64 KB.
```
invalidname:
```suggestion
A process writes to both, and each pipe holds only so much data before a write to it blocks — on Linux, for example, about 64 KB.
```

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+- Handling output that's too large to hold in memory.
+- Providing input to the subprocess while it runs.
+
+The streaming API gives you live streams, which come with one rule:
+
+> Warning: A subprocess can't finish while a pipe it depends on is stuck.
+>
+> Drain every output stream you open — concurrently, when you open more than one — and close standard input once you're done writing to it.
+
+## Stream output as it arrives
```
invalidname: All the topics other than Overview should be H3s (`###`), as they were in the previous article.

invalidname (changes requested):
This looks really useful. I've left a few suggestions to consider.

Sources/Subprocess/Documentation.docc/GettingStarted.md
dempseyatgithub: I realize now this is DocC-specific markup, so it won't appear exactly like this once processed by DocC. Feel free to mark this as resolved.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,206 @@
...
+```swift
+let result = try await run(
+    .name("git"),
+    arguments: ["rev-parse", "HEAD"],
+    output: .string(limit: 4096)
+)
+print(result.standardOutput)   // the commit hash
+```
+
+Because `.string(limit:)` produces a `String` rather than an optional, you use `result.standardOutput` directly, with no unwrapping.
```
dempseyatgithub: This sentence may not be necessary since the result type of `.string(limit:)` is already listed above as `String` and the code sample does not do any unwrapping. I think most Swift developers would already know that using something of type `String` does not require unwrapping.

(Is it called out here potentially for those who may have used an earlier version of the API?)

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Detect failure
+
+Around a subprocess, the word “error” pulls in three directions.
+Keeping them apart is the difference between handling a real problem and misreading a healthy result as a broken one.
+Throughout the Subprocess documentation:
+
+- A **Swift error** is a value thrown out of `run` that you handle with `do`/`catch`. Subprocess reserves this for problems it can't express as a result: a broken setup or environment, or a body closure of your own that throws.
+- A **command failure** is a subprocess that ran to completion but exited with a non-zero code. This is a normal ``ExecutionResult``, not a Swift error.
```
dempseyatgithub: Possibly be a bit more explicit:
"This is reported by the ``terminationStatus`` of the  ``ExecutionResult``, not a Swift error."

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+            for try await line in execution.standardOutput.strings() {
+                print(line)
+            }
+        }
+        try await group.waitForAll()
+    }
+}
+```
+
+The same structure can chain two subprocesses, sending the stream of output from one subprocess to the input of another.
```
dempseyatgithub, line 191: I think creating the equivalent to a shell pipe is important enough that this should be its own section.
Possibly 'Chaining Subprocesses' or even 'Chaining Subprocesses To Pipe Commands' or 'Pipe Commands By Chaining Subprocesses'.

When I first scanned the article, I was hoping to find code on how to pipe using Subprocess and missed this section completely.

I also think using the term 'pipe' or 'piping' either in a heading title or body text would make this section easier to find in searches.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Use files and inherited standard I/O
+
+Not every stream has to flow through your process.
+You can point the input or output of a subprocess at a file descriptor directly with
+``FileDescriptorInput/fileDescriptor(_:closeAfterSpawningProcess:)`` and
+``FileDescriptorOutput/fileDescriptor(_:closeAfterSpawningProcess:)``,
+which is efficient for large amounts of data.
+When you use a file descriptor, the bytes never pass through your code.
+
```
dempseyatgithub, line 243: Possibly add a code example showing writing the output of a command that generates a lot of text directly to a file via file descriptor?

heckj: I wanted to keep this article focused on the basics and most common usage. I agree that an example showing how to use the file descriptors effectively - where they shine - would be a great add, but I don't want to expand the scope of this article to include that. I think that would be better as it's own, focused thing - I'm seeing that as "advanced use" of this library.

Sources/Subprocess/Documentation.docc/GettingStarted.md
heckj: leftover edit that I should have removed ;-)

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
heckj: I think this would be an excellent additional article to add, referenced from here. I don't think it's a great addition for this article though, since it's trying to be more introductory and common use case, and doing so would expand the scope and narrative of the this article, which is really meant to provide the basics and rules behind handling streaming. The piping and chaining concepts are something I'd describe as more advanced topics. Not invalid, just not something I want to try and cover in this article.

invalidname (approved):
Looks great. Thanks for all your work on this.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,206 @@
...
+
+Running a subprocess takes the input you provide, and you control how its output and error are collected or streamed.
+Depending on the command, a subprocess might produce output, error output, both, or neither.
+Subprocess lets you handle each of these cases independently.
+
+The function that launches a subprocess, `run`, comes in two forms:
+* The *collecting* form waits for the subprocess to finish and provides the result.
+* The *streaming* form lets you read output while the subprocess runs.
+
+You choose between them depending on whether you pass a trailing closure.
```
iCharlesHu: This sentence isn't entirely accurate. You choose whether you stream or collect by passing the appropriate options to `output` and `error` parameters. Technically, you could choose to collect (via .string, .data, etc.) even when passing a trailing closure. In other words, you are not forced to stream if you pass in a trailing closure.

Sources/Subprocess/Documentation.docc/GettingStarted.md
```diff
@@ -0,0 +1,206 @@
...
+    .name("ls"),
+    arguments: ["-la"],
+    output: .string(limit: 4096)
+)
+print(result.standardOutput)
+```
+
+The preceding example uses ``run(_:arguments:environment:workingDirectory:platformOptions:input:output:error:)-(_,_,_,_,_,Input,_,_)``:
+
+Take note of the following behaviors in this example:
```
iCharlesHu, line 41: Extra line here

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+Stream the output of a subprocess as it arrives and feed it input, while following
+the one rule that keeps pipes from deadlocking.
+
+## Overview
+
+The function that launches a subprocess, `run`, comes in two forms.
+The collecting form waits for the subprocess to exit and hands you all its output at once.
+The streaming form takes a trailing closure so you can process output while the subprocess runs.
+You choose between them depending on whether you pass that closure.
+This article covers the streaming form; for collecting, see <doc:GettingStarted>.
```
iCharlesHu: Similar to the edit in `GettingStarted.md`: this description is not accurate. You choose whether you stream by passing in `.sequence` as the output or error type. You can pass in a closure and not stream, but if you want to stream you must pass in a closure.

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+_ = try await run(
+    .name("sort"),
+    input: .inputWriter,
+    output: .sequence,
+    error: .discarded
+) { sort in
+    let writer = sort.standardInputWriter
+    try await withThrowingTaskGroup(of: Void.self) { group in
+        group.addTask {
+            _ = try await run(
```
iCharlesHu, line 207: I'm not 100% sure if we should highlight this example. Subprocess 1.0 deliberately lacks process piping because it's scheduled for 1.1: https://github.com/swiftlang/swift-subprocess/issues/151

While this example works, it's inefficient (parent output shouldn't need to be "read out" to a sequence and then write into child process via `writer.write`; they should just have the pipes connected directly), and it's not the shape we want to encourage developers to copy.

On the other hand, this example is how you can emulate process piping with the API today, so I do see it has value. @heckj, what do you think? Should we leave this out if we know we will be introducing an official API later?

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
```diff
@@ -0,0 +1,259 @@
...
+
+Not every stream has to flow through your process.
+You can point the input or output of a subprocess at a file descriptor directly with
+``FileDescriptorInput/fileDescriptor(_:closeAfterSpawningProcess:)`` and
+``FileDescriptorOutput/fileDescriptor(_:closeAfterSpawningProcess:)``,
+which is efficient for large amounts of data.
+When you use a file descriptor, the bytes never pass through your code.
+
+You can also let the subprocess share your process's own terminal. For input,
+that's `.standardInput`; for output and error, it's `.currentStandardOutput` and
```
iCharlesHu: Ahh thanks for catching this! Input should also be `.currentStandardInput`. I'll update the code.

iCharlesHu (approved):
Thanks so much!!

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
heckj: Will update to match!

Sources/Subprocess/Documentation.docc/GettingStarted.md
heckj: Awesome, thank you! I'm going to use almost your sentence description directly!

Sources/Subprocess/Documentation.docc/GettingStarted.md
heckj: thank you!

Sources/Subprocess/Documentation.docc/StreamingAndInput.md
heckj: I was on the fence about this, given how awkward the API is, but it was a frequent ask in the conversations I followed, not even also tracking #151. I'm inclined to leave it in, primarily because of the asks about it specifically, and it's what we have with the current release of this API. I'd also suggest that as we pin down better API for this down the road, we make a note to explicitly come back and change up this article, or break it out into it's own article (if it's sufficiently deep to warrant it) when updated API is available.

More than happy to pop back when we get there to revise and update, or draft a new article along these lines!

PR #354, opened 2026-07-14
Remove explicit swift-docc-plugin dependency

ktoso:
Nowadays (as weird as it is) projects are not expected to have the dependency declared explicitly. It must be added ad-hoc or by the swift package index. Otherwise all kinds of tools break, including swiftlang/docs.

It would be ideal that all projects can just add the plugin, but that's not the state of the world nowadays.

Meanwhile I did workarounds in adopter projects to unbreak swiftlang/docs building with those added, more details here https://github.com/swiftlang/swift-java/pull/845

cc @heckj

Here's the related swiftpm issue https://github.com/swiftlang/swift-package-manager/issues/10294

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,14 +9,6 @@ var dep: [Package.Dependency] = [
         from: "1.5.0"
     )
 ]
-#if !os(Windows)
-dep.append(
-    .package(
-        url: "https://github.com/apple/swift-docc-plugin",
-        from: "1.4.5"
-    ),
-)
-#endif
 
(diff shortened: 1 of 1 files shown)


iCharlesHu (approved):

iCharlesHu:
Hmm maybe this is why I can't get the docs to show up on swiftpackageindex.com... Thanks!!

ktoso:
Yeah that could be it tbh, we can ask Sven!

Issue #357, opened 2026-07-29
`Executable.name(_:)` searches the current working directory before `PATH` — decide and document the contract


jakepetroules:
## Summary

`Executable.name(_:)` is documented as a `PATH` lookup, but the resolver searches the **current working directory** as well — and searches it *before* `PATH`. The details of *which* current directory, and at what point it is consulted, differ between the eager `resolveExecutablePath(in:)` API and the spawn-time search, and between Unix and Windows. Notably the eager API and the spawn path disagree with each other on both platforms, in opposite directions (measurements below).

This is not a crash or a leak; it is an unspecified contract. Whichever behavior we want, developers currently cannot predict it from the documentation, and the two platforms cannot both be satisfied by the same mental model. This issue lays out what the code does today, the precedents, and the options, so we can pick one deliberately: **document the current behavior, or change it (ideally uniformly).**

Filed as a follow-up to the `access(dir, X_OK)` directory false-positive fix (#358), which is a separate, narrower bug: a directory named like the executable was accepted as a runnable candidate. That fix does not address, and deliberately does not change, the question below.

## The documented contract

`Sources/Subprocess/Configuration.swift` (`Executable.name(_:)`):

```swift
/// Locates the executable by name.
///
/// Subprocess searches for the executable by name in the directories listed
/// in the `PATH` environment variable, in order, and runs the first match it
/// finds. To run an executable at a known location instead, use ``path(_:)``.
public static func name(_ executableName: String) -> Self
```

`Documentation.docc/GettingStarted.md` says the same thing twice: ".name("ls") looks the command up using the `PATH`" and "it throws a `SubprocessError` … when `Executable/name(_:)` finds nothing in `PATH`".

No mention of the current working directory anywhere.

### Unix

Measured on macOS 26.0 / arm64, Swift 6.4-dev, at `eee1243`. Setup: a `tool` script exists **only** in `/tmp/cwdprobe/scratch`, which is **not** on `PATH` (`PATH=/usr/bin:/bin`, passed explicitly as `Environment.custom`).

```swift
let env = Environment.custom(["PATH": "/usr/bin:/bin"])

// 1. eager resolve
try await Executable.name("tool").resolveExecutablePath(in: env)

// 2. spawn with an explicit workingDirectory
try await Subprocess.run(
    .name("tool"),
    environment: env,
    workingDirectory: FilePath("/tmp/cwdprobe/scratch"),
    output: .string(limit: 64)
)
```

[...]

PR #359, opened 2026-07-31
Resolve Executable.name(_:) against PATH only on every platform

iCharlesHu:
`Executable.name(_:)` is documented as a `PATH` lookup, but it also searched a current directory ahead of `PATH`. This behavior is opposite of what every POSIX runtime agreed on, and it allows potential arbitrary execution. Update `Executable.name(_:)` such that an executable name now resolves the same way on all platforms: by walking the `PATH` the subprocess will receive, falling back to the current process's value and then to the directories the platform reports as standard (`confstr(_CS_PATH)` on Unix, the `CreateProcessW` search list without the current directory on Windows). Empty and relative entries are skipped, so every resolved path is absolute and `resolveExecutablePath(in:)` agrees with the spawn path, and a name containing a path separator is now rejected with `spawnFailed` rather than resolved against a current directory.

### DIFF
--- a/Sources/Subprocess/Configuration.swift
+++ b/Sources/Subprocess/Configuration.swift
@@ -332,6 +332,18 @@ public struct Executable: Sendable, Hashable {
     /// Subprocess searches for the executable by name in the directories listed
     /// in the `PATH` environment variable, in order, and runs the first match it
     /// finds. To run an executable at a known location instead, use ``path(_:)``.
+    ///
+    /// The search uses the `PATH` of the environment you pass to the subprocess,
+    /// falling back to the `PATH` of the current process when that environment
+    /// doesn't set one, and to the directories the platform considers standard
+    /// when neither sets one. Only absolute directories are searched: Subprocess
+    /// skips empty and relative `PATH` entries, and never searches the current
+    /// working directory of either process. The working directory you pass to
+    /// `run` therefore has no effect on which executable runs.
+    ///
+    /// The executable name must be a name, not a path: passing a name that
+    /// contains a path separator throws a ``SubprocessError`` with the code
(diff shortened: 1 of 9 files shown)


iCharlesHu:
Resolves: https://github.com/swiftlang/swift-subprocess/issues/357

Sources/Subprocess/Configuration.swift
```diff
@@ -350,6 +373,61 @@ public struct Executable: Sendable, Hashable {
...
+extension Executable {
+    #if os(Windows)
+    /// The character that separates one directory from the next in a `PATH`
+    /// value.
+    internal static let pathValueSeparator: Character = ";"
+    /// The characters that name a location, and so may not appear in a name
+    /// passed to ``name(_:)``.
+    ///
+    /// Windows accepts either slash as a separator, and `:` separates a drive
+    /// from a drive-relative path, so `C:tool` names a location in the same way
```
jakepetroules, line 387: C:tool means relative to the current directory of drive C, but the current drive might not be drive C. If you are on drive E, then C:tool and .\tool mean different things.

iCharlesHu: Right, but `C:tool` would still be a "path" even if you are on `E`, right? Basically, the purpose of `pathSeparators` is that we want to reject any "executable name" that might look like a path, especially relative ones. Here, we'd reject `.name("C:tool")` regardless of whether you are on `C` or not.

jakepetroules: Ah, makes sense - I hadn't been thinking about it specifically in that context.

jakepetroules (approved):

PR #360, opened 2026-08-06
Use neutral phrasing in README

broken-circle:
#328 made wording changes to use more neutral phrasing:

> I went ahead and extended this to remove the biological or family metaphors from the documentation content (such as "spawn", "kill", "children", "parent") to use more neutral/agnostic terms.

I noticed some of these phrases were still present in the README.

cc @heckj

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Subprocess
 
-Subprocess is a cross-platform Swift package for spawning child processes, built from the ground up with Swift concurrency.
+Subprocess is a cross-platform Swift package for launching subprocesses, built from the ground up with Swift concurrency.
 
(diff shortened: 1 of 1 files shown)


broken-circle:
This and prior PRs address user-facing documentation, but many uses of the non-neutral phrases still exist in internal documentation and code comments. If we're open to removing these, I'll open a separate PR to do that. I think it's a good idea because it keeps the entire codebase, public or otherwise, aligned on the same phrasing.

One exception is when referring specifically to APIs like `posix_spawn`; any documentation or comments referring to such an API would be left as-is.

jakepetroules (approved):

iCharlesHu (approved):

iCharlesHu:
@broken-circle could you rebase? I'm not sure why the jobs are not running...

broken-circle:
@iCharlesHu Sure thing. Something seemed to be wrong with GitHub Actions late last week, since it didn't run here or on a few other PRs I updated, but looks like it's working now!

heckj:
(thank you @broken-circle  - I didn't see the ping to review until after it was merged, but it looks great!)

PR #365, opened 2026-08-13
Prefer posix_spawn over fork/exec on every Unix platform

### LINKED ISSUE #172: Improve file descriptor disinheritance performance
Wrote this up a while back and pasting it here so we can consider applying some of the APIs described here to maximize file descriptor disinheritance performance across various platforms.

--

In POSIX, spawning a new process inherits the parent's file descriptors by default. This is widely considered a security issue, and many libraries (including swift-subprocess) attempt to effectively flip this default.

Unfortunately, preventing subprocesses from inheriting all open file descriptors requires using non-portable API and comes with a significant performance penalty depending on the platform and libc versions. In most cases, swift-subprocess will attempt to use `posix_spawn` in cases where the platform offers a related `posix_spawn` API for efficiently closing all open file descriptors. If the platform offers no such API, it will fall back to `fork`+`exec` and use a related API for efficiently closing all open file descriptors, and finally fall back to manually computing the largest open file descriptor and manually issuing a potentially very large number of `close()` syscalls.

The following table indicates the performance and techniques used based on the spawn strategy used. The `posix_spawn` column indicates the specific _O(1)_ `posix_spawn` API used to close all file descriptors. The `pre-fork` column indicates the file descriptor closing strategy used when `posix_spawn` can't be used. `N/A` in both columns indicates that _O(n)_ manual closing will be used.

| Platform | Version                      | posix_spawn                                              | pre-fork                |
| -------- | ---------------------------- | -------------------------------------------------------- | ----------------------- |
| macOS    | 10.7+                        | posix_spawnattr_setflags + POSIX_SPAWN_CLOEXEC_DEFAULT   | -                       |
| Linux    | kernel 5.9+ with Glibc 2.34+ | posix_spawn_file_actions_addclosefrom_np                 | close_range             |
| ^        | kernel 5.9+ with older Glibc | -                                                        | close_range             |
| ^        | kernel 5.9+ with Musl        | -                                                        | close_range             |
| ^        | kernel < 5.9                 | -                                                        | -                       |
| Android  | kernel 5.11+                 | posix_spawnattr_setflags + POSIX_SPAWN_CLOEXEC_DEFAULT   | close_range             |
| ^        | kernel < 5.11                | posix_spawnattr_setflags + POSIX_SPAWN_CLOEXEC_DEFAULT * | -                       |
| FreeBSD  | 13.1+                        | posix_spawn_file_actions_addclosefrom_np                 | close_range / closefrom |
| OpenBSD  | 3.5+                         | -                                                        | closefrom               |

\* POSIX_SPAWN_CLOEXEC_DEFAULT requires Android 13 (API level 33) and provides no good programmatic way to know if it's supported other than checking the version at runtime
\* Android with a kernel older than Linux 5.11 falls back to an _O(n)_ solution when using posix_spawn because close_range with CLOSE_RANGE_CLOEXEC is not available.

jakepetroules:
> [!IMPORTANT]
> The second commit, **[DO NOT MERGE] Run the test suite on OpenBSD**, is temporary. It exists so this PR can be seen green on every platform the package supports, and must be reverted before merging.

Subprocess used `posix_spawn` only on Darwin and `fork`/`exec` everywhere else, in two near-duplicate implementations that each owned the candidate-path loop, the transient-failure retry, the file-descriptor bookkeeping and the error mapping.

Both now live in one shared implementation in `Subprocess+Spawn.swift`, which prefers `posix_spawn` and falls back to `fork`/`exec` — or, on Darwin, a pre-`fork` plus `posix_spawn(POSIX_SPAWN_SETEXEC)`, so Darwin still never reaches a raw `execve` — only for configurations `posix_spawn` cannot express.

### When the fallback path is still required

Capabilities are probed once per process **at runtime**, not at compile time, since a binary built against one C library is expected to run against another.

1. `userID`, `groupID` or `supplementaryGroups` is set. No platform has a `posix_spawn` attribute for `setuid`, `setgid` or `setgroups`.
2. The C library offers no way to close every inherited descriptor. Closing them by enumerating this process's own table would leak any descriptor another thread opens between the enumeration and the spawn, so that is not an acceptable substitute.
3. `createSession` without `POSIX_SPAWN_SETSID` (FreeBSD, OpenBSD).
4. `createSession` together with `processGroupID`. glibc calls `setsid` then `setpgid`, Bionic the reverse, where `setsid` then fails `EPERM` because the child already leads a process group. The fallback path controls the order explicitly and reproduces the previous behavior.
5. `workingDirectory` without an `fchdir` file action, or without a way to open a descriptor for it (OpenBSD, Android below API 34). Both halves are tested because they are established differently — one by `dlsym` at runtime, one by which of `O_SEARCH`/`O_PATH` the headers define — and a working directory silently ignored would be the worst available outcome.

### Effect

Linux with glibc 2.34+, Android 13+ and FreeBSD 13.1+ stop forking for the common case: `posix_spawn` on glibc clones with `CLONE_VM` and `CLONE_VFORK`, so there is no page-table copy and no `fork()` in a multithreaded parent. musl, older glibc, older Android and OpenBSD keep `fork`/`exec`. On Darwin, `createSession` no longer forks, since `POSIX_SPAWN_SETSID` works there.

### Why the C shim grew

Swift no longer names `posix_spawn_file_actions_t` or `posix_spawnattr_t`. glibc and musl typedef structs, Darwin and the BSDs typedef pointers, and Bionic annotates its mutators `_Nonnull`, so no single Swift spelling compiles everywhere; the shim wraps every entry point behind an opaque handle. The public `PlatformSpawnAttributes` and `PlatformSpawnFileActions` name the form each platform's C library accepts, and an unavailable function calls the library's mutators through them so a mistake there is a build failure rather than a public API nobody can use. Darwin declares neither: its `preSpawnProcessConfigurator` predates them and keeps its `inout` signature, so code that configures the spawn on Darwin *and* elsewhere still has to branch on `#if canImport(Darwin)`.

Those typealiases have to be public, and on every platform whose C library declares the two types as structs, Swift resolves them through the shim rather than through the libc overlay — even where that overlay is itself imported publicly. The shim is therefore imported publicly off Darwin, which is also why OpenBSD already carried such an import; the `// FIXME: Why is this necessary only on OpenBSD?` it was added under is now answered.

### Working directory

A working directory is now applied through an `fchdir` file action on a descriptor opened in the parent, which replaces the TOCTOU-prone guess about whether a spawn failure meant a missing executable or a missing directory. The open flags come from the shim: `chdir` needs only execute permission, so `O_RDONLY` would reject a directory that is a perfectly legal working directory.

Only a real `O_SEARCH` (Darwin, FreeBSD) actually checks that permission, though. The kernel zeroes `acc_mode` for `O_PATH`, so on glibc, musl and Bionic a directory with mode `000` opens successfully and the child's `fchdir` then fails `EACCES` — one of the errnos meaning "try the next candidate path", so it would surface as `executableNotFound`, the very misdiagnosis this replaces. Searchability is therefore checked explicitly, against the returned descriptor rather than the path, so the object checked is the object `fchdir` will use and no TOCTOU window is added. Whether that check can ask about effective IDs is another C library question answered in C: `AT_EACCESS` where it works, `0` on Bionic, which rejects every nonzero `faccessat` flag and has no set-uid programs for the flag to matter to. Where no search-only open exists the directory is validated without being opened.

### Ordering that is load-bearing

File actions are appended in the order the child must run them, which matters in three places. The `fchdir` action goes first, because the descriptor it names can itself be 0, 1 or 2 — it is opened after the pipes, so it takes the lowest free number, and a caller whose standard descriptors are closed leaves low numbers free — and a `dup2` onto that number would replace the directory before the child reached it. Parent-end closes precede the `dup2`s, or a close of a parent end that happens to occupy 0, 1 or 2 would close the child's standard stream instead. The close-all action goes last, after everything that still needs a descriptor.

The child's signal dispositions are reset through `POSIX_SPAWN_SETSIGDEF` with every signal *except* `SIGKILL` and `SIGSTOP`. Excluding those two is required rather than tidiness: Bionic resets each member of that set unconditionally, and Linux rejects `sigaction()` on them with `EINVAL` even for `SIG_DFL`, whereupon Bionic `_exit(127)`s a child that `posix_spawn` still reports as successfully launched. Neither signal can be caught or ignored, so neither can have a disposition needing a reset on any platform.

### Apple platforms other than macOS

`posix_spawn_file_actions_addfchdir_np` is referenced under `TARGET_OS_OSX` rather than `TARGET_OS_MAC`, which is *every* Apple platform: the SDK marks it `__API_UNAVAILABLE(ios, tvos, watchos, visionos)`, and that is a hard error rather than a deprecation warning. Mac Catalyst is inside that exclusion, since it reports as iOS even though it runs on macOS. Leaving the capability unset there routes a working directory through rule 5 to the fallback path.

### Other observable changes

A `setpgid` or `setsid` failure is reported rather than silently ignored **where the C library applies it**, as part of `posix_spawn`; the fork/exec path still discards both results, which is now documented on `processGroupID`. Working-directory errors carry the errno that actually occurred, and on FreeBSD a process descriptor is available only on the `fork`/`exec` path until `posix_spawnattr_setprocdescp_np` can be required. Both FreeBSD 15.1 follow-ups are recorded as FIXMEs.

### Testing

A test-only task-local forces the fallback path, so every spawn-sensitive test runs under both paths on every platform, instead of macOS never exercising `fork` and modern glibc never exercising `fork`/`exec`. Which path a spawn is expected to take is derived from the same rule the implementation consults rather than from that override alone, because where `posix_spawn` cannot express the request the unforced case legitimately takes the fallback too.

CI is what validates this change: its entire behavioral effect lands on platforms that cannot be exercised on a development Mac. The matrix covers glibc across five distros on 6.2 and nightly, the al2 4.18 and 5.10 kernels (glibc 2.26, so the fallback path), Android on NDK r27d and r29, FreeBSD 14.3, musl static-SDK builds, Windows, macOS, the CMake build that does not glob sources, and — via the temporary job above — OpenBSD.

### Found but not fixed here

A review of this change turned up three **pre-existing** defects, all of which behave identically before and after it. They are out of scope — this PR is already large, and the remedies touch the `fork`/`exec` path — but they should not be lost:

- **A child's pipe end landing on descriptor 0, 1 or 2 is clobbered.** Both paths bind the child's ends in the fixed order `dup2(inputRead, 0)`, `dup2(outputWrite, 1)`, `dup2(errorWrite, 2)`. If `outputWrite` is itself descriptor 0, the first `dup2` overwrites it and the second then copies `inputRead` onto 1, so the child's stdout becomes the read end of its own stdin pipe; the child dies on the redirection rather than producing wrong output. This needs no caller-supplied descriptors, only a caller whose standard descriptors are closed — a daemon, or a process re-exec'ing itself. The fix is to relocate any child end numbered below 3 with `F_DUPFD_CLOEXEC` in the parent, which would cover both paths at once. `testStandardStreamsSurviveLowParentDescriptors` does not catch it: it uses pipes for all three streams, where the *parent* end always takes the lower number.

- **`supplementaryGroups: []` silently retains the parent's groups.** Both the rule that routes privileged configurations to the `fork`/`exec` path and the shim itself test `count > 0`, so an empty array — the natural way to ask that inherited supplementary groups be *dropped* — results in no `setgroups` call at all. `setgroups(0, NULL)` is the meaningful call and is never made.

- **A child can inherit the entire descriptor table if `/dev/fd` cannot be opened.** Where `close_range(…, CLOSE_RANGE_CLOEXEC)` is unavailable (kernels before 5.9, which includes the `al2-4.18` CI leg), the forked child falls back to enumerating `/dev/fd`, and if that `open` fails it simply returns, leaking every descriptor into the child with no error reported. Reachable in a `FROM scratch` container or a chroot with no `/proc` — a normal deployment for a statically linked musl binary, which is also a configuration where rule 2 forces this path for every spawn. glibc's own `posix_spawn` fails the spawn in the equivalent situation rather than proceeding. This change *reduces* exposure, since glibc ≥ 2.34, Android ≥ 13 and FreeBSD ≥ 13.1 all move off that path, but it is still the designated safe fallback.

Fixes #172


Sources/Subprocess/Platforms/Subprocess+Spawn.swift
```diff
@@ -0,0 +1,1040 @@
...
+            (stdio.errorWrite, CInt(2)),
+        ] {
+            guard let descriptor else { continue }
+            let result = _subprocess_spawn_file_actions_adddup2(fileActions, descriptor, target)
+            guard result == 0 else { throw spawnFailure(result) }
+        }
+
+        // 4. Close everything else, last: this must run after the dup2 and
+        //    fchdir actions, or it would close the descriptors they need.
+        if capabilities.closeAllViaSpawn == .closefromAction {
```
iCharlesHu, line 433: Should this be `assert closeAllViaSpawn` is in fact `.closefromAction`? If underlying posix_spawn does not support close from action we can't use posix_spawn at all right?

Sources/Subprocess/Platforms/Subprocess+Spawn.swift
```diff
@@ -0,0 +1,1040 @@
...
+    ///   `.invalidDescriptor`.
+    private func resolveWorkingDirectory() -> (PlatformFileDescriptor, CInt) {
+        guard let workingDirectory = self.workingDirectory?.string else {
+            return (.invalidDescriptor, 0)
+        }
+
+        // Both branches capture `errno` inside the closure, before
+        // `withPlatformString` tears its buffer down, so nothing between the
+        // failed call and the read can clobber it.
+        if _subprocess_can_open_directory_for_search() != 0 {
```
iCharlesHu, line 668: Could you elaborate on why you chose to explicitly open the working directory here and then pass the fd to `addfchdir` instead of using `addchdir`, which takes a path directly? It doesn't look like we are using that fd in other ways, right? `addfchdir` seems so much more complex and you had to add many shims for it.

Sources/Subprocess/Platforms/Subprocess+Spawn.swift
```diff
@@ -0,0 +1,1040 @@
...
+    }
+}
+
+/// A count of the spawns each path performed. See
+/// ``SpawnCapabilities/spawnPathTallyForTesting``.
+///
+/// `@unchecked Sendable` because it is a task-local reference mutated only from
+/// the single task that installed it, so its mutable state cannot be reached
+/// concurrently.
+internal final class SpawnPathTally: @unchecked Sendable {
```
iCharlesHu, line 147: While I understand the intent behind the `SpawnPathTally`, I don't think it's appropriate or necessary to include code in a library *solely for testing purposes*.  It looks like the only purpose for `SpawnPathTally` is to allow unit tests to inspect the internal state of Subprocess under specific conditions. IMO those classes of unit tests are too granular. We shouldn't need tests to verify whether a fallback path is taken, as long as Subprocess works end to end on the target platform. It's good to have these while you are developing to ensure each method implementation is correct, but we should remove them before merging.

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -863,6 +602,116 @@ extension ProcessIdentifier: CustomStringConvertible, CustomDebugStringConvertib
...
+///
+/// Never called; it exists so that getting the condition above wrong is a build
+/// failure on the affected platform instead of a public API nobody can use.
+///
+/// It deliberately probes mutating entry points rather than the `init`s. Bionic
+/// annotates only its `init`s `_Nullable` — they are out-parameters — and
+/// everything else `_Nonnull`, so the two import differently there and `init` is
+/// the unrepresentative one. These are the calls a configurator actually makes.
+@available(*, unavailable)
+private func _assertPlatformSpawnTypesMatchLibc(
```
iCharlesHu, line 707: I don't really follow the point of this function 🤔.  The comment didn't really help either… why is it necessary?

Sources/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -891,6 +749,32 @@ public struct PlatformOptions: Sendable {
...
+    /// `posix_spawn()`.
+    ///
+    /// This is best-effort: it may not be called when a configuration cannot be
+    /// expressed with `posix_spawn` and must be launched with `fork` and `exec`
+    /// instead. That happens when ``userID``, ``groupID`` or
+    /// ``supplementaryGroups`` is set, when ``createSession`` is combined with
+    /// ``processGroupID``, or when this platform's C library cannot close
+    /// inherited file descriptors, change directory, or create a session
+    /// through `posix_spawn`.
+    public var preSpawnProcessConfigurator:
```
iCharlesHu, line 771: Doesn't this technically change the API on Linux, since we decided to remove the fork/exec-based `preSpawnProcessConfigurator`? I also have concerns about 

> This is best-effort: it may not be called when a configuration cannot be expressed with `posix_spawn`

This seems like a foot gun. It's not clear from the call site whether `preSpawnProcessConfigurator` is guaranteed to be called, and there is no confirmation in the spawn result to confirm that it is indeed called. Wouldn't it lead to silent failures?

Do we need to add this API with this PR? IMO we should add it separately, especially since technically it requires API review.

PR #368, opened 2026-08-24
Fix misreported errno when posix_spawnattr configuration fails on Darwin

jakepetroules:
## Summary
- The `spawnAttributeError` branch of `Configuration.spawn` on Darwin reported `Errno(rawValue: result)`, but `result` is left over from the preceding stdin/stdout/stderr `posix_spawn_file_actions_adddup2`/`addclose` setup calls above it, not the actual `posix_spawnattr_setflags`/`posix_spawnattr_setpgroup`/`posix_spawnattr_set_qos_class_np` failure.
- Report `spawnAttributeError` instead, so the underlying error actually reflects why `posix_spawnattr` configuration failed.

## Test plan
- [x] `swift build` succeeds
- [x] `swift test` passes (170 tests)

### DIFF
--- a/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
+++ b/Sources/Subprocess/Platforms/Subprocess+Darwin.swift
@@ -410,7 +410,7 @@ extension Configuration {
 
                     let error: SubprocessError
                     if spawnAttributeError != 0 {
-                        error = SubprocessError.spawnFailed(withUnderlyingError: Errno(rawValue: result))
+                        error = SubprocessError.spawnFailed(withUnderlyingError: Errno(rawValue: spawnAttributeError))
                     } else {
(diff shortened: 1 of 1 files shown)


iCharlesHu (approved):

iCharlesHu:
@jakepetroules do you know what's going on with CI?

jakepetroules:
Looks like Amazon Linux 2 disappeared from images.linuxcontainers.org, and the amazonlinux2023 and ubi9 builds for the latest main snapshot, failed. I reached out to @cmcgee1024 about the AL2 ones.

Shall we merge without those, considering the rest is passing?

cmcgee1024:
I recommend going ahead and merging this if it's important. Linuxcontainers.org dropped support for AL2 entirely. It might be possible to test older Linux kernels using another Swift supported distro, but that'll require some time and analysis.

PR #369, opened 2026-08-25
Close .sequence output/error pipes even when the consumer stops early

jakepetroules:
## Summary
- `SubprocessOutputSequence.Iterator` only closed its underlying pipe descriptor when a read returned EOF. If a caller stopped consuming `execution.standardOutput`/`standardError` before reaching end-of-stream -- for example, breaking out of a `for await` loop early -- that read end was never closed. This leaks one fd per invocation, unbounded, with no relationship to how many subprocesses are running concurrently: a long-running process that streams output and doesn't always drain to completion can leak thousands of fds over its lifetime this way.
- Confirmed with a targeted repro before writing the fix: a control group that fully drains `.sequence` output leaked 0 fds over 30 iterations, while a group that broke out after the first chunk leaked exactly 1 fd every single iteration (30/30).
- The iterator itself never closes its descriptor. `run()` is the sole owner of closing each `.sequence` output/error stream, doing so exactly once right after the body closure returns or throws -- whether or not the stream was ever iterated or fully drained. This is done with explicit calls on both the success and thrown-error paths rather than `defer`, so a future asynchronous close implementation (e.g. backed by io_uring) has an async context to run in.

## Test plan
- [x] Added `testBreakingOutOfSequenceOutputEarlyDoesNotLeakResources`, which breaks out of a `.sequence` output stream early across 200 iterations and asserts the process's open fd/handle count doesn't grow unbounded.
- [x] Verified the new test fails without the fix (traps on the leak-detection precondition) and passes with it.
- [x] `swift build` succeeds
- [x] `swift test` passes (171 tests)

### DIFF
--- a/Sources/Subprocess/SubprocessOutputSequence.swift
+++ b/Sources/Subprocess/SubprocessOutputSequence.swift
@@ -63,11 +63,13 @@ public struct SubprocessOutputSequence: AsyncSequence, @unchecked Sendable {
         private let processIdentifier: ProcessIdentifier
         private let preferredBufferSize: Int
         private var buffer: [Buffer]
+        private let closeGate: AtomicCounter
 
-        internal init(diskIO: DiskIO, processIdentifier: ProcessIdentifier) {
+        internal init(diskIO: DiskIO, processIdentifier: ProcessIdentifier, closeGate: AtomicCounter) {
             self.diskIO = diskIO
             self.processIdentifier = processIdentifier
             self.buffer = []
+            self.closeGate = closeGate
             // Only need to query it once at beginning of stream
             self.preferredBufferSize = AsyncIO.queryPipeBufferSize(for: diskIO)
         }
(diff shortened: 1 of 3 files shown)


Sources/Subprocess/SubprocessOutputSequence.swift
```diff
@@ -120,7 +128,27 @@ public struct SubprocessOutputSequence: AsyncSequence, @unchecked Sendable {
         guard self.state.initializedCount() == 1 else {
             fatalError("SubprocessOutputSequence is single-pass. It can only be iterated once.")
         }
-        return Iterator(diskIO: self.diskIO, processIdentifier: self.processIdentifier)
+        return Iterator(diskIO: self.diskIO, processIdentifier: self.processIdentifier, closeGate: self.closeGate)
+    }
+
+    /// Closes the underlying descriptor if the iterator hasn't already
```
FranzBusch: FWIW, I don't think the iterator should ever close the descriptor at end of stream. I kinda expect that those FDs will only be closed at end of `run` scope.

iCharlesHu: > I don't think the iterator should ever close the descriptor at end of stream. 

@FranzBusch could you elaborate on what’s the concern for closing the file descriptor early? I suppose closing it only at the end of the `run` scope allows us to remove the close gate 🤔

FranzBusch: My main concern is that with asynchronous close this won't be implementable at all. If we ever were to back this with I/O mechanisms that only offer asynchronous close i.e. io_uring then we can not do it from here since the context is not async. So instead of differing when it happens I would recommend only doing this at the end of `run`.

jakepetroules: Good point; done.

iCharlesHu (approved):
Thanks!

